PR #10, opened 2024-08-19
Remove dashmap

yamadapc:
The bundler uses the “DashMap” crate to handle multi-threaded hashmaps.

The dependency include significant amounts of unsafe code, which is unaudited. We’ve a large amount of exceptions that point at different dashmap sections. Segmentation faults have been previously reported on the dashmap repository, since its code is inherently memory unsafe and prone to such memory bugs.

This commit removes this dependency replacing it with `RwLock` structures. I've already tested that this does not introduce a large regression.

Previous problems:

* `https://github.com/xacrimon/dashmap/pull/290`
* `https://github.com/xacrimon/dashmap/issues/167`
* `https://github.com/xacrimon/dashmap/issues/139`
* `https://github.com/xacrimon/dashmap/issues/127`
* `https://github.com/xacrimon/dashmap/issues/97`
* `https://github.com/xacrimon/dashmap/issues/107`


.gitignore
```diff
@@ -2,6 +2,7 @@
 .idea/
 .nyc_output/
 .atlaspack-cache/
+.parcel-cache/
```
mattcompiles, line 5: Accident?

yamadapc: We should have both listed

MonicaOlejniczak: We shouldn't need it in this repo, internally sure

yamadapc: I added because switching branches leads to cache files being there. 

Is there a downside from ignoring it?

MonicaOlejniczak: Why does it exist, the rename should have removed these references?

yamadapc: It exists because I was coming from the original repository, with an added upstream. It will also exist if we bitsect or try to bring in changes from the parcel repository.

Do you see a downside from ignoring? I can just remove this line.

CONTRIBUTING.md
```diff
@@ -6,9 +6,9 @@ Thank you for considering a contribution to Atlaspack! Unfortunately pull reques
 
 For pull requests, please:
 
-* Add tests for new features and bug fixes
-* Follow the existing style
-* Separate unrelated changes into multiple pull requests
+- Add tests for new features and bug fixes
```
yamadapc, line 9: This will be changed by the pre-commit hook for anyone doing a commit

.gitignore
MonicaOlejniczak: I'd rather remove links to parcel in this repo. You could add it to a global ignore on your machine, but I won't block you

MonicaOlejniczak (approved):

yamadapc:
Resolving 500k imports in a real-world application as this stands

**Unsafe parcel resolver**
* total_runs=10
* average_duration=2.2535219919

**Safe parcel resolver**
* total_runs=10
* average_duration=1.9010456582000004

**Atlaspack resolver without dashmap**
* total_runs=10
* average_duration=3.2973253873000004

yamadapc:
This is slower than it was, but still plenty fast. I want to accept a regression. We can make this faster if there is bulk loading of file-system metadata.

.gitignore
yamadapc: I'm resolving this unless there's a reason not to add

PR #15, opened 2024-08-20
Ensure watcher snapshot is written to correct FS

mattcompiles:
## Motivation

Fix the atlaspack-query test.

## Changes

Turns out that we were previously writing the watcher snapshot to the `inputFS` when it should be the `outputFS`. Fixing this plus outputting to `NodeFS` for the atlaspack query tests should resolve the issue.

## Checklist


yamadapc (approved):

MonicaOlejniczak (approved):

mattcompiles:
Looks like the unit tests are wrong now. Just fixing that before merging.

MonicaOlejniczak (approved):

PR #18, opened 2024-08-20
Start to vendor lmdb-js-lite

yamadapc:
This starts to add a Rust based wrapper for LMDB, as published in the `lmdb-js-lite` npm package. That package will be deprecated and vendored into the parcel repository. 

This pull-request will be changed to that the LMDB bindings are built into the main bundler rust native add-on, rather than a separate one.


crates/lmdb-js-lite/src/writer.rs
```diff
@@ -0,0 +1,527 @@
+use std::path::Path;
```
yamadapc, line 1: And here

crates/lmdb-js-lite/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+      let error = (|| -> Result<(), DatabaseWriterError> {
+        read.put(&mut write_txn, &format!("{i}"), &buffer)?;
+        write_txn.commit()?;
+        Ok(())
+      })();
+      if let Err(DatabaseWriterError::HeedError(heed::Error::Mdb(heed::MdbError::MapFull))) = error
+      {
+        current_size = current_size * 2;
+        tracing::info!("Resizing database {current_size}");
+        unsafe { read.environment().resize(current_size).unwrap() }
```
yamadapc, line 449: I did not implement resizing because resizing requires us to put a lock around all the transactions so that we only resize when there are no open transactions. Even in that case it's hard to validate & it'll be a bit of UB to resize.

crates/lmdb-js-lite/.npmignore
```diff
@@ -0,0 +1,13 @@
+target
```
MonicaOlejniczak, line 1: Would it be easier just to specify files instead of an ignore list?

crates/lmdb-js-lite/__benchmarks__/README.md
```diff
@@ -0,0 +1,3 @@
+```
```
MonicaOlejniczak, line 1: This isn't really a readme

yarn.lock
```diff
@@ -18,11 +18,24 @@
     "@babel/highlight" "^7.23.4"
     chalk "^2.4.2"
 
+"@babel/code-frame@^7.12.13", "@babel/code-frame@^7.24.7":
+  version "7.24.7"
+  resolved "https://packages.atlassian.com/api/npm/npm-remote/@babel/code-frame/-/code-frame-7.24.7.tgz#882fd9e09e8ee324e496bd040401c6f046ef4465"
```
MonicaOlejniczak, line 23: Should be public, not atlassian

crates/lmdb-js-lite/__benchmarks__/prepare-read-benchmark-unsafe.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const unsafeDB = openLMDBUnsafe({
+    path: "./databases/unsafe/read",
+    encoding: "binary",
+    compression: true,
+    eventTurnBatching: true,
+  });
+
+  console.log("Generating entries for testing");
+  const entries = [...Array(NUM_ENTRIES)].map(() => {
```
MonicaOlejniczak, line 35: /nit you'd allocate a bit less with `Array.from({ length: NUM_ENTRIES }, () => generateEntry())`

crates/lmdb-js-lite/__test__/index.test.ts
```diff
@@ -0,0 +1,170 @@
...
+      asyncWrites,
+      mapSize: MAP_SIZE,
+    });
+
+    {
+      const value = Math.random().toString();
+      await db.put("key", v8.serialize(value));
+      const result = await db.get("key");
+      const resultValue = result && v8.deserialize(result);
+      expect(value).toEqual(resultValue);
```
MonicaOlejniczak, line 55: /nit arguments should be flipped

crates/lmdb-js-lite/__test__/index.test.ts
```diff
@@ -0,0 +1,170 @@
...
+          path: "./databases/unsafe",
+          compression,
+        });
+
+        await unsafeDB.transaction(async () => {
+          for (let i = 0; i < numEntriesToTest; i += 1) {
+            await unsafeDB?.put(`${i}`, v8.serialize(i));
+          }
+        });
+      }, 40000);
```
MonicaOlejniczak, line 138: Does the global jest timeout not apply here?

crates/lmdb-js-lite/tsconfig.json
```diff
@@ -0,0 +1,121 @@
+{
+  "exclude": [
+    "__test__",
+    "__benchmarks__",
+    "dist"
+  ],
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
```
MonicaOlejniczak, line 8: /nit I personally prefer only having the options you use listed, otherwise it's hard to read

crates/lmdb-js-lite/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "lmdb-js-lite",
+  "version": "0.1.5",
+  "main": "typescript/compat.ts",
```
MonicaOlejniczak, line 4: No js?

crates/lmdb-js-lite/package.json
```diff
@@ -0,0 +1,60 @@
...
+        "x86_64-unknown-linux-gnu",
+        "aarch64-apple-darwin",
+        "universal-apple-darwin"
+      ]
+    }
+  },
+  "license": "APACHE-2",
+  "devDependencies": {
+    "@napi-rs/cli": "^2.18.4",
+    "@parcel/cache": "*",
```
MonicaOlejniczak, line 26: Were you going to change these to atlaspack?

crates/lmdb-js-lite/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "bench-node": "^0.0.1-beta.0",
+    "jest": "^29.7.0",
+    "lmdb": "^3.0.13",
+    "prettier": "^3.3.3",
+    "ts-jest": "^29.2.4",
+    "ts-node": "^10.9.2",
+    "typescript": "^5.5.4"
+  },
+  "engines": {
+    "node": ">= 10"
```
MonicaOlejniczak, line 41: That's a pretty old version, is that really the case?

crates/lmdb-js-lite/__benchmarks__/prepare-read-benchmark-unsafe.ts
yamadapc: I hadn't seen that trick before ; 
This isn't supposed to be fast. Creating the buffers is quite slow, but this part is very fast so it does not matter.

Micro-benchmark is a lot faster chrome (`[...Array(...)]` is faster), and on firefox `Array.from(...)` is faster. Both run in way less than 1ms. If you run this section you'll see that it takes several seconds to generate and insert the entries.

https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiI3dzdsaUQtaGppMk5LVW1SZUw5T2giLCJjb2RlIjoicmV0dXJuIFsuLi5BcnJheSgxMDAwKV0ubWFwKCgpID0-IE1hdGgucmFuZG9tKCkpIiwibmFtZSI6InNwcmVhZCIsImRlcGVuZGVuY2llcyI6W119LHsiaWQiOiJ5Qk81OUMzQUgtckcxQ3ZqWFRUWkgiLCJjb2RlIjoicmV0dXJuIEFycmF5LmZyb20oeyBsZW5ndGg6IDEwMDAgfSwgKCkgPT4gTWF0aC5yYW5kb20oKSkiLCJuYW1lIjoiZnJvbSBvYmplY3QiLCJkZXBlbmRlbmNpZXMiOltdfV0sImNvbmZpZyI6eyJuYW1lIjoiQmFzaWMgZXhhbXBsZSIsInBhcmFsbGVsIjp0cnVlLCJnbG9iYWxUZXN0Q29uZmlnIjp7ImRlcGVuZGVuY2llcyI6W119LCJkYXRhQ29kZSI6IiJ9fQ

yamadapc: Screenshot 2024-08-21 at 4 31 36 PM

crates/lmdb-js-lite/.npmignore
yamadapc: This was just generated by napi bootstrapping ; I don't think it makes a difference

crates/lmdb-js-lite/__benchmarks__/README.md
yamadapc: I've added more text to it.

yarn.lock
yamadapc: Fixed, thanks

crates/lmdb-js-lite/__test__/index.test.ts
yamadapc: Fixed

crates/lmdb-js-lite/__test__/index.test.ts
yamadapc: You're right, I've removed i

crates/lmdb-js-lite/package.json
yamadapc: Fixed it ; this will change as internally consuming this package will be a bit different

crates/lmdb-js-lite/package.json
yamadapc: Changed

mattcompiles (approved):

PR #23, opened 2024-08-21
Fix plugin forwards and backwards compatibility

yamadapc:
Use the parcel plugin config symbol. This lets parcel plugins work with this tool and will allow all plugins from this tool to work with parcel.

In the future the atlaspack plugins can skip this symbol and just pass around objects that implement an interface. Adding the symbol into the plugin would be a compatibility layer.

## Motivation

Fix parcel plugin compatibility

* It'd be possible to handle both atlaspack and parcel symbols here, but that will mean plugins declared with `@atlaspack/plugin` will not work with parcel

### DIFF
--- a/packages/core/plugin/src/PluginAPI.js
+++ b/packages/core/plugin/src/PluginAPI.js
@@ -13,6 +13,8 @@ import type {
   Validator as ValidatorOpts,
 } from '@atlaspack/types';
 
+// This uses the `parcel-plugin-config` symbol so it's backwards compatible with
+// parcel plugins.
 const CONFIG = Symbol.for('parcel-plugin-config');
 
 export class Transformer {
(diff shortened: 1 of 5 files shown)


packages/core/plugin/src/PluginAPI.js
```diff
@@ -13,7 +13,7 @@ import type {
   Validator as ValidatorOpts,
 } from '@atlaspack/types';
 
-const CONFIG = Symbol.for('atlaspack-plugin-config');
+const CONFIG = Symbol.for('parcel-plugin-config');
```
mattcompiles, line 16: Maybe leave a comment saying this is for backwards compat with parcel plugins?

mattcompiles (approved):

PR #29, opened 2024-08-23
Add atlaspack migration package

MonicaOlejniczak:
## Motivation

It is currently tedious to migrate from Parcel to Atlaspack, as there are several parts of a codebase that need to change. This change introduces a CLI to automate some of the migration process.

## Changes

Add a `@atlaspack/parcel-to-atlaspack` package to expose a global `parcel-to-atlaspack` command that makes these changes:

**.parcelrc**
  * Rename to `.atlaspackrc`
  * Update any `@parcel/` plugin references to `@atlaspack/`

**package.json**
  * Update `engines.parcel` to `engines.atlaspack`
  * Update `@parcel/` plugin configurations to `@atlaspack/`
  * Update `parcel` and `@parcel/` dependencies to `@atlaspack`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ yarn-debug.log*
 yarn-error.log*
 *.log
 *.min.js
+*.tsbuildinfo
--- a/packages/migrations/parcel-to-atlaspack/README.md
+++ b/packages/migrations/parcel-to-atlaspack/README.md
@@ -1,0 +1,34 @@
+# @atlaspack/parcel-to-atlaspack
+
+This package exposes a CLI that can be used to migrate a Parcel application to Atlaspack
+
+## Installation
+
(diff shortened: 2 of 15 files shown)


yamadapc (approved):
Looks good but I think

* with parcelrc it'd be better to parse than search-and-replace
* when changing package.json we likely should add resolutions so parcel packages resolve to these

MonicaOlejniczak:
I personally don't think there's much benefit to parsing the parcelrc file, since the replacements are very straightforward. If we need to add onto it, then that can be revisited. As for resolutions, imo that will depend on the repository. A lot of smaller repos could just replace them, but monorepos may want to preserve them. I'll just leave it as is for now, and it can be manually modified since it's usually at the package root anyway.

PR #30, opened 2024-08-25
Fix enum mapping for bundleBehaviour and sourceType

mattcompiles:
## Motivation

Fix native asset graph compatibility with legacy js asset graph.

## Changes

- Added mapping for bundle behaviour in `getAssetGraph`
- Added mapping for source type in `getAssetGraph`

## Checklist


yamadapc (approved):

MonicaOlejniczak (approved):

mattcompiles:
Looks like I broke a test with this change. Looking into it before merging.

PR #31, opened 2024-08-26
Fix more native test scenarios

mattcompiles:
## Motivation

Fix more native test scenarios. 144/278 now passing. 

## Changes

- Merging duplicate dependencies instead of adding them to the graph
  - I think ideally we'll remove duplicate deps at some point but I want to avoid changing the transformer as it's still used outside of v3
- Add correct default for `Asset.is_source`
- Add correct default for `Asset.is_bundle_splittable`
- Add correct empty file resolution to the default resolver plugin
- Fix built-in resolution to return empty file rather than excluded 

## Checklist

### DIFF
--- a/crates/atlaspack/src/requests/asset_graph_request.rs
+++ b/crates/atlaspack/src/requests/asset_graph_request.rs
@@ -247,10 +247,25 @@ impl AssetGraphBuilder {
       .insert(request_id, asset_node_index);
 
     // Connect dependencies of the Asset
-    for dependency in &dependencies {
-      let _ = self
-        .graph
-        .add_dependency(asset_node_index, dependency.clone());
+    let mut unique_deps: HashMap<u64, Dependency> = HashMap::new();
+
+    for mut dependency in dependencies {
+      unique_deps
+        .entry(dependency.id())
+        .and_modify(|d| {
+          // This code is an incomplete version of mergeDependencies in packages/core/core/src/Dependency.js
+          // Duplicate dependencies can occur when node globals are polyfilled
+          // e.g. 'process'. I think ideally we wouldn't end up with two
+          // dependencies post-transform but that needs further investigation to
+          // resolve and understand...
+          d.meta.append(&mut dependency.meta);
+          d.symbols.extend(dependency.symbols.clone());
+        })
+        .or_insert(dependency.clone());
+    }
+
+    for (_id, dependency) in unique_deps.into_iter() {
+      let _ = self.graph.add_dependency(asset_node_index, dependency);
     }
 
     self.graph.propagate_requested_symbols(
--- a/crates/atlaspack_core/src/types/asset.rs
+++ b/crates/atlaspack_core/src/types/asset.rs
@@ -195,8 +195,8 @@ impl Asset {
       code: Arc::new(code),
       side_effects,
       file_type,
-      is_source,
       is_bundle_splittable: true,
+      is_source,
       ..Asset::default()
     })
   }
--- a/crates/atlaspack_core/src/types/asset.rs
+++ b/crates/atlaspack_core/src/types/asset.rs
@@ -186,6 +186,8 @@ impl Asset {
       Code::from(code_from_disk)
     };
 
+    let is_source = !file_path.ancestors().any(|p| p.ends_with("/node_modules"));
+
     Ok(Self {
       id: create_asset_id(&env, &file_path, &pipeline, &query, &None),
       file_path,
@@ -193,6 +195,8 @@ impl Asset {
       code: Arc::new(code),
       side_effects,
       file_type,
+      is_source,
+      is_bundle_splittable: true,
       ..Asset::default()
     })
   }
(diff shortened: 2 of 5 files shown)


crates/atlaspack/src/requests/asset_graph_request.rs
```diff
@@ -247,10 +247,25 @@ impl AssetGraphBuilder {
...
-      let _ = self
-        .graph
-        .add_dependency(asset_node_index, dependency.clone());
+    let mut unique_deps: HashMap<u64, Dependency> = HashMap::new();
+
+    for mut dependency in dependencies {
+      unique_deps
+        .entry(dependency.id())
+        .and_modify(|d| {
+          // This code is an incomplete version of mergeDependencies in packages/core/core/src/Dependency.js
```
MonicaOlejniczak, line 256: I believe it primarily occurs when you define multiple imports in the same file, for example:

```javascript
import { a } from 'foo';
import { b } from 'foo';

import('foo');
require('foo');
```

Where `a` and `b` style imports may be a stylistic choice, but isn't really used internally. The inline require optimisation may also duplicate imports possibly.

crates/atlaspack/src/requests/asset_graph_request.rs
```diff
@@ -247,10 +247,25 @@ impl AssetGraphBuilder {
       .insert(request_id, asset_node_index);
 
     // Connect dependencies of the Asset
-    for dependency in &dependencies {
-      let _ = self
-        .graph
-        .add_dependency(asset_node_index, dependency.clone());
+    let mut unique_deps: HashMap<u64, Dependency> = HashMap::new();
+
+    for mut dependency in dependencies {
```
MonicaOlejniczak, line 252: Previously this code was in the transformer, what do you think of moving it there?

crates/atlaspack_core/src/types/asset.rs
```diff
@@ -186,13 +186,17 @@ impl Asset {
       Code::from(code_from_disk)
     };
 
+    let is_source = !file_path.ancestors().any(|p| p.ends_with("/node_modules"));
```
MonicaOlejniczak, line 189: You probably want a `starts_with` since internally for local consumption we have many prefixes starting with `node_modules` that are symlinks.

crates/atlaspack_core/src/types/asset.rs
```diff
@@ -186,13 +186,17 @@ impl Asset {
...
+
     Ok(Self {
       id: create_asset_id(&env, &file_path, &pipeline, &query, &None),
       file_path,
       env,
       code: Arc::new(code),
       side_effects,
       file_type,
+      is_source,
+      is_bundle_splittable: true,
```
MonicaOlejniczak, line 199: /nit `is_bundle_splittable` before `is_source`

crates/atlaspack_plugin_resolver/src/atlaspack_resolver.rs
```diff
@@ -221,8 +221,11 @@ impl AtlaspackResolver {
       _ => {
         return Ok(Resolved {
           invalidations: Vec::new(),
-          resolution: Resolution::Excluded,
-        })
+          resolution: Resolution::Resolved(ResolvedResolution {
+            file_path: self.options.core_path.join("_empty.js"),
```
MonicaOlejniczak, line 225: I think you can remove committing `_empty.js` in core, if you leverage the optional `code` field, so:

```rust
resolution: Resolution::Resolved(ResolvedResolution {
  code: Some(String::default()),
  file_path: self.options.core_path.join("_empty.js"),
  ..ResolvedResolution::default()
}),
```

crates/atlaspack/src/requests/asset_graph_request.rs
mattcompiles: I actually ran into it from the usage of global polyfills, specifically `process`. The above makes more sense but I didn't realise they caused duplicates.

crates/atlaspack/src/requests/asset_graph_request.rs
mattcompiles: Well I'd argue that it belonged in core as it was part of `UncommittedAsset.addDependency`. We could move it to the transformer but it'd be mean other transformers wouldn't have this sorted.

crates/atlaspack_core/src/types/asset.rs
mattcompiles: Do you have an example? This should only match on exact directory names as `ancestors` walks up the tree which I thought was equivalent to what we're doing in v2?

crates/atlaspack_plugin_resolver/src/atlaspack_resolver.rs
mattcompiles: Interesting. I'll try that.

crates/atlaspack_core/src/types/asset.rs
MonicaOlejniczak: Yeah that's equivalent to v2, I messaged you the exact names

crates/atlaspack/src/requests/asset_graph_request.rs
MonicaOlejniczak: Makes sense, happy to leave it this way then

crates/atlaspack_plugin_resolver/src/atlaspack_resolver.rs
mattcompiles: Done

MonicaOlejniczak (approved):

PR #32, opened 2024-08-27
Retain .parcelrc suffix in migration cli

MonicaOlejniczak:
## Motivation

Currently the migration cli will rename all files that start with `.parcelrc` to `.atlaspackrc` which will cause many files to be overwritten. For example, two files named `.parcelrc` and `.parcelrc-customsuffix` will both be renamed as `.atlaspackrc`.

## Changes

Retain the suffix when renaming `.parcelrc` files


yamadapc:
I think I changed things so the config must be `parcelrc`

MonicaOlejniczak:
@yamadapc That's fine, you can opt out of the change in the script with `--skip-parcelrc`. The current script represents an end-state migration.

yamadapc (approved):

PR #34, opened 2024-08-27
[V3] Registering transformer plugins

alshdavid:
This PR adds the ability for the Nodejs context to initialize Transformer plugins. They currently don't do anything other than resolve, initialize & store the target transformer plugin 

- removed `AtlaspackV3.build()` because it's currently unused
- expanded `NodejsWorker` to include `transformerRegister()` method
  - Used to resolve and register a transformer plugin within the nodejs context
- injecting `RpcWorker` into `RpcTransformerPlugin`
- calling `transformerRegister` when `RpcTransformerPlugin` is initialized

### DIFF
--- a/crates/atlaspack_plugin_rpc/src/nodejs/rpc_conn_nodejs.rs
+++ b/crates/atlaspack_plugin_rpc/src/nodejs/rpc_conn_nodejs.rs
@@ -38,7 +38,7 @@ impl RpcWorker for NodejsWorker {
     Ok(())
   }
 
-  fn transformer_register(&self, resolve_from: &Path, specifier: &str) -> anyhow::Result<()> {
+  fn register_transformer(&self, resolve_from: &Path, specifier: &str) -> anyhow::Result<()> {
     self
       .transformer_register_fn
       .call_with_return(
--- a/crates/atlaspack_plugin_rpc/src/nodejs/rpc_conn_nodejs.rs
+++ b/crates/atlaspack_plugin_rpc/src/nodejs/rpc_conn_nodejs.rs
@@ -1,3 +1,6 @@
+use std::path::Path;
+
+use anyhow::Context;
 use atlaspack_napi_helpers::anyhow_from_napi;
 use atlaspack_napi_helpers::js_callable::JsCallable;
 use napi::{JsObject, JsUnknown};
@@ -8,12 +11,17 @@ use crate::RpcWorker;
 /// single Nodejs worker thread
 pub struct NodejsWorker {
   ping_fn: JsCallable,
+  transformer_register_fn: JsCallable,
 }
 
 impl NodejsWorker {
   pub fn new(delegate: JsObject) -> napi::Result<Self> {
     Ok(Self {
       ping_fn: JsCallable::new_from_object_prop_bound("ping", &delegate)?,
+      transformer_register_fn: JsCallable::new_from_object_prop_bound(
+        "transformerRegister",
+        &delegate,
+      )?,
     })
   }
 }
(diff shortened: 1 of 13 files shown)


packages/core/core/src/atlaspack-v3/worker/worker.js
```diff
@@ -1,14 +1,27 @@
...
     // console.log('Hi');
   }
+
+  async transformerRegister(resolve_from: string, specifier: string) {
+    let customRequire = module.createRequire(resolve_from);
+    let resolvedPath = customRequire.resolve(specifier);
+    // $FlowFixMe
+    let transformer = await import(resolvedPath);
+    this.#transformers.set(`${resolve_from}:${specifier}`, transformer);
+  }
```
alshdavid, line 24: Notes:

I'm using ESM `import` to support both `type: module` and `type: commonjs`.

I am doing resolution within Nodejs rather than using the internal Atlaspack resolver so it can use builtin Node features (for example; `--condition`, commonly used to tell nodejs to resolve TypeScript sources locally and dist JavaScript files in production)

crates/atlaspack_plugin_rpc/src/plugin/transformer.rs
```diff
@@ -9,26 +9,40 @@ use atlaspack_core::plugin::TransformResult;
...
+      _rpc_worker: rpc_worker,
+      name: plugin.package_name.clone(),
     })
   }
 }
 
 impl TransformerPlugin for RpcTransformerPlugin {
-  fn transform(&mut self, _asset: Asset) -> Result<TransformResult, Error> {
-    todo!()
+  fn transform(&mut self, asset: Asset) -> Result<TransformResult, Error> {
```
mattcompiles, line 41: So the idea is you'll call something like `rpc_worker.transform(&plugin.package_name)` to actually execute it? 

Also is there a reason we separate the register from the actual transform? If we don't load the transformer until it's actually used we can avoid the overhead of loading one that goes unused by the project?

alshdavid: > So the idea is you'll call something like rpc_worker.transform(&plugin.package_name) to actually execute it?

Yep, that's pretty much it. ofc you have to also pass in the values (like Asset, Config, etc) in that call too.

> is there a reason we separate the register from the actual transform?

Yeah I was thinking of deferring plugin loading until the first call to `transform` in a follow up PR

crates/atlaspack_plugin_rpc/src/nodejs/rpc_conns_nodejs.rs
```diff
@@ -8,21 +10,21 @@ use crate::RpcWorker;
...
-  conns: Vec<NodejsWorker>,
+  workers: Vec<NodejsWorker>,
 }
 
 impl NodejsWorkerFarm {
   pub fn new(conns: Vec<NodejsWorker>) -> Self {
     Self {
       current_index: Default::default(),
-      conns,
+      workers: conns,
```
MonicaOlejniczak, line 20:
```suggestion
      workers,
```

crates/atlaspack_plugin_rpc/src/nodejs/rpc_conn_nodejs.rs
```diff
@@ -29,4 +37,24 @@ impl RpcWorker for NodejsWorker {
       .map_err(anyhow_from_napi)?;
     Ok(())
   }
+
+  fn transformer_register(&self, resolve_from: &Path, specifier: &str) -> anyhow::Result<()> {
```
MonicaOlejniczak, line 41: Shouldn't it be `register_transformer`? I find the current naming hard to interpret

alshdavid: Can be, I have it the other way around so it can remain in alphabetical ordering when all the methods are added.
```
resolver_load_config
resolver_register
resolver_resolve
transformer_load_config
transformer_register
transformer_transform
```
Happy to change it to be the other way around though

```
load_config_resolver
load_config_transformer
resolve_resolver
register_resolver
register_transformer
transform_transformer
```

MonicaOlejniczak (approved):

PR #38, opened 2024-08-28
[V3] Fix scope hoisting value and enable scope hoisting tests

mattcompiles:
## Motivation

We currently don't model unset values in the `default_target_options` so any unset values show up as their default which is false for booleans. We can implement sensible defaulting in future but for now we may as well let the JS layer pass in the values. 

## Changes

- Honour passed in `default_target_optins.should_scope_hoist` value
- Enable a portion of the scope hoisting test suite

## Checklist

### DIFF
--- a/crates/atlaspack/src/requests/target_request.rs
+++ b/crates/atlaspack/src/requests/target_request.rs
@@ -380,9 +380,9 @@ impl TargetRequest {
             .output_format
             .unwrap_or_else(|| fallback_output_format(context)),
           should_optimize: self.default_target_options.should_optimize,
-          should_scope_hoist: self.default_target_options.should_scope_hoist
-            && self.mode == BuildMode::Production
-            && !self.default_target_options.is_library,
+          // TODO: Handle defaulting for unset values, currently this value can
+          // only be true or false so we need to honour it.
+          should_scope_hoist: self.default_target_options.should_scope_hoist,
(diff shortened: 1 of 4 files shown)


crates/atlaspack/src/requests/target_request.rs
```diff
@@ -380,9 +380,9 @@ impl TargetRequest {
             .output_format
             .unwrap_or_else(|| fallback_output_format(context)),
           should_optimize: self.default_target_options.should_optimize,
-          should_scope_hoist: self.default_target_options.should_scope_hoist
-            && self.mode == BuildMode::Production
-            && !self.default_target_options.is_library,
+          // TODO: Handle defaulting for unset values, currently this value can
+          // only be true or false so we need to honour it.
+          should_scope_hoist: self.default_target_options.should_scope_hoist,
```
MonicaOlejniczak, line 385: Shouldn't you just make the default target scope hoist value optional then?

mattcompiles: I can. I was initially concerned that it would get out of sync with JS land, but with fresh eyes maybe that's not an issue. 🤔

mattcompiles: Updated

MonicaOlejniczak (approved):

PR #50, opened 2024-08-30
[V3] Migrate `FileSystem` to use Rust's `std::fs` interface

alshdavid:
This PR expands the `FileSystem` trait and implementations to adhere to the "interface"/"shape" of the Rust standard library's fs api.

The existing `FileSystem` methods have been retained but marked as deprecated. They will be removed in following PRs as it'll be too big for one PR.

I started on this because I needed access to the `std::fs::read` function (and eventually `File::open`) - but figured, rather than adding my own versions to our existing `FileSystem` trait, it would be better to tidy up the file system crate before it deviates too far from the standard library.

So this PR syncs up our `FileSystem` crate with the Rust standard library. Only the methods we use have been implemented on the in memory fs (and the various other fs implementations we have... we have like 6 different filesystem implementations?), but we can expand them as needed in future.


crates/atlaspack/src/requests/entry_request.rs
```diff
@@ -36,7 +36,11 @@ impl Request for EntryRequest {
       entry_path = request_context.project_root.join(entry_path);
     };
 
-    if request_context.file_system().is_file(&entry_path) {
```
alshdavid, line 39: `is_file` would return `false` if the file did not exist. Some other circuits relied on this - is that required in this case?

crates/atlaspack_filesystem/src/lib.rs
```diff
@@ -1,66 +1,7 @@
...
-    Err(std::io::Error::new(
-      std::io::ErrorKind::Other,
-      "Not implemented",
-    ))
-  }
-
-  fn canonicalize(
-    &self,
-    path: &Path,
-    _cache: &FileSystemRealPathCache,
```
alshdavid, line 50: I moved the cache implementation within the struct to function as an inner detail. We weren't sharing the cache around, just newing it up and passing it back into the fs.

```rust
let fs = OsFileSystem {
  cache: FileSystemRealPathCache::default()
}

fs.read(Path::from("/foo/bar.txt"))
```

crates/atlaspack_config/src/atlaspack_rc_config_loader.rs
```diff
@@ -119,7 +119,7 @@ impl AtlaspackRcConfigLoader {
         .resolved
     };
 
-    self.fs.canonicalize_base(&path).map_err(|source| {
```
alshdavid, line 122: is `canonicalize_base` supposed to behave differently to `std::fs::canonicalize`? Should it be part of `FileSystemExt`?

crates/atlaspack_filesystem/Cargo.toml
```diff
@@ -5,9 +5,9 @@ edition = "2021"
 description = "FileSystem wrapper trait for use in Atlaspack codebase."
 
 [dependencies]
-mockall = "0.12.1"
```
alshdavid, line 8: Removed `automock` because it has issues with the trait and we were not using it, all tests were using the `InMemoryFileSystem`

crates/atlaspack_filesystem/src/file_system.rs
```diff
@@ -0,0 +1,210 @@
...
+use std::path::PathBuf;
+use std::sync::Arc;
+use std::time::SystemTime;
+
+pub type FileSystemRef = Arc<dyn FileSystem + Send + Sync>;
+
+pub trait FileSystem {
+  // -----------------------------------
+  // Custom methods on top of std::fs::*
+  // -----------------------------------
```
alshdavid, line 18: TODO: Move the additional methods to a `FileSystemExt` trait

crates/atlaspack_filesystem/src/in_memory/file_system.rs
```diff
@@ -0,0 +1,267 @@
...
+  }
+}
+
+impl FileSystem for InMemoryFileSystem {
+  fn cwd(&self) -> io::Result<PathBuf> {
+    Ok(self.current_working_directory.read().clone())
+  }
+
+  fn open(&self, path: &Path) -> io::Result<Box<dyn crate::File>> {
+    Ok(Box::new(Cursor::new(self.read(path)?)))
```
alshdavid, line 62: TODO; write support for `fs::open`. This isn't being used atm so I will finish it when it's needed

crates/atlaspack_filesystem/src/in_memory/file_system.rs
```diff
@@ -0,0 +1,267 @@
...
+    while let Some(path) = dir {
+      files.insert(path.to_path_buf(), InMemoryFileSystemEntry::Directory);
+      dir = path.parent();
+    }
+
+    Ok(())
+  }
+}
+
+impl std::fmt::Debug for InMemoryFileSystem {
```
alshdavid, line 133: This prints out the structure of the `InMemoryFilesystem` when passed into `dbg!(&fs)`. The content of the files is replaced with their size rather than printing their contents.

crates/atlaspack_filesystem/src/in_memory/file_system.rs
```diff
@@ -0,0 +1,267 @@
...
+
+#[cfg(target_os = "windows")]
+fn root_dir() -> PathBuf {
+  PathBuf::from("C:/")
+}
+
+/// In memory implementation of a file-system entry
+#[derive(Debug)]
+enum InMemoryFileSystemEntry {
+  File { contents: Vec<u8> },
```
alshdavid, line 28: This was changed from an `String` to a `Vec<u8>` because Atlaspack can work with non text files - like images. This is needed for plugin support

crates/atlaspack_filesystem/src/in_memory/metadata.rs
alshdavid, line 1: https://doc.rust-lang.org/std/fs/struct.Metadata.html

crates/atlaspack_filesystem/src/lib.rs
alshdavid, line 1: Making the library entry file function as only an index of the public exports

crates/atlaspack_filesystem/src/os/canonicalize.rs
alshdavid, line 1: This can probably be generalized on top of the `FileSystem` trait rather than being fs-specific

crates/atlaspack_filesystem/src/utils/find_ancestor.rs
alshdavid, line 1: TODO: Make this part of the `FileSystemExt` trait

crates/atlaspack_napi_helpers/src/js_callable/js_callable.rs
alshdavid, line 1: Could probably just use `Arc<JsCallable>` at the call site :thinking:

crates/node-bindings/src/file_system/file_system_napi.rs
alshdavid, line 1: TODO: Add tests for the Napi filesystem

crates/atlaspack_filesystem/src/file_system.rs
```diff
@@ -0,0 +1,210 @@
...
+
+  // #[deprecated] // TODO: remove usages
+  /// Write a String to a file
+  fn write_file(&self, path: &Path, data: String) {
+    self.write(path, data.as_bytes()).unwrap()
+  }
+
+  // ------------------------
+  // Trait mapping std::fs::*
+  // ------------------------
```
alshdavid, line 68: https://doc.rust-lang.org/std/fs/index.html

crates/atlaspack_filesystem/src/file_system.rs
```diff
@@ -0,0 +1,210 @@
...
+
+  fn symlink_metadata(&self, _path: &Path) -> io::Result<Box<dyn Metadata>> {
+    unimplemented!()
+  }
+
+  fn write(&self, _path: &Path, _contents: &[u8]) -> io::Result<()> {
+    unimplemented!()
+  }
+}
+
```
alshdavid, line 137: https://doc.rust-lang.org/std/fs/struct.Metadata.html

crates/atlaspack_filesystem/src/file_system.rs
```diff
@@ -0,0 +1,210 @@
...
+  fn permissions(&self) -> Box<dyn Permissions> {
+    unimplemented!()
+  }
+
+  fn len(&self) -> u64 {
+    unimplemented!()
+  }
+}
+
```
alshdavid, line 175: https://doc.rust-lang.org/std/fs/struct.Permissions.html

crates/atlaspack_filesystem/src/file_system.rs
```diff
@@ -0,0 +1,210 @@
...
+  fn mode(&self) -> u32 {
+    unimplemented!()
+  }
+
+  #[cfg(target_family = "unix")]
+  fn set_mode(&mut self, _mode: u32) {
+    unimplemented!()
+  }
+}
+
```
alshdavid, line 195: https://doc.rust-lang.org/std/fs/struct.FileType.html

crates/atlaspack_filesystem/src/file_system.rs
```diff
@@ -0,0 +1,210 @@
...
+pub trait FileSystem {
+  // -----------------------------------
+  // Custom methods on top of std::fs::*
+  // -----------------------------------
+
+  /// Get the current working directory of the process
+  fn cwd(&self) -> io::Result<PathBuf> {
+    unimplemented!()
+  }
```
alshdavid, line 23: Question: Why is `cwd` part of the filesystem? That's more of an environment thing

crates/node-bindings/src/resolver.rs
```diff
@@ -94,11 +95,7 @@ pub struct JsFileSystem {
 }
```
alshdavid, line 96: TODO: Should probably move `JsFileSystem` into the `atlaspack_filesystem` crate and share the implementation with `resolver_old`

packages/utils/node-resolver-rs/benches/node_resolver_bench.rs
alshdavid, line 1: TODO: Should move `PreloadingFileSystem` into the `atlaspack_filesystem` crate

crates/atlaspack_filesystem/src/file_system.rs
```diff
@@ -0,0 +1,210 @@
...
+
+  fn create_dir(&self, _path: &Path) -> io::Result<()> {
+    unimplemented!()
+  }
+
+  fn create_dir_all(&self, _path: &Path) -> io::Result<()> {
+    unimplemented!()
+  }
+
+  fn hard_link(&self, _original: &Path, _link: &Path) -> io::Result<()> {
```
mattcompiles, line 89: Why are implementing the entire FS API? Wouldn't it be easier to just leave out the ones we don't use?

crates/atlaspack_config/src/atlaspack_rc_config_loader.rs
yamadapc: The `OsFileSystem` canonicalize implementation is different to std::fs::canonicalize and going through some caches. 

I reckon it does make some diff. to performance. The caching side of things should be moved out of the file-system trait.

crates/atlaspack_filesystem/src/file_system.rs
```diff
@@ -0,0 +1,210 @@
...
+  fn exists(&self, path: &Path) -> io::Result<bool> {
+    let Ok(metadata) = self.metadata(path) else {
+      return Ok(false);
+    };
+    Ok(metadata.is_dir() || metadata.is_file())
+  }
+
+  /// Open a handle to a file
+  fn open(&self, _path: &Path) -> io::Result<Box<dyn File>> {
+    unimplemented!()
```
yamadapc, line 35: Remove all unimplemented defaults.

If we have it on the trait, implement for all file-system implementations. If we do not need it, then do not add to the trait.

crates/atlaspack_filesystem/src/file_system.rs
yamadapc: I think we need to set cwd / check cwd on some places no? Definitely for in-memory

crates/atlaspack_filesystem/src/file_system.rs
yamadapc: Do we really need a trait for this? Is there usage?

crates/atlaspack_filesystem/src/in_memory/file_system.rs
yamadapc: I think we shouldn't support abstract file handles because that gets very complicated.

If we need to stream file contents, then let's just return readers / streams.

What are you using raw handles for?

crates/atlaspack_filesystem/src/file_system.rs
alshdavid: The trait simply documents the methods but they all default to `unimplemented!()`. The full fs methods are listed to avoid misnaming methods in future, ensuring correctness moving forward.

The `OsFileSystem` is the only implementation that implements all of the methods and only because it's effortless to forward the calls. Given all the tests use the `InMemoryFileSystem`, that's the real target for implementation and method implementations should be added there as needed.

crates/atlaspack_filesystem/src/lib.rs
yamadapc: Nice that's good

crates/atlaspack_config/src/atlaspack_rc_config_loader.rs
alshdavid: Is the output different? Or does it produce the same output, the difference is that it uses a cache internally?

crates/atlaspack_filesystem/src/in_memory/file_system.rs
alshdavid: It's part of the plugin API to access/modify files via a reader/writer. Here, `File` is just an `io::Reader + io::Writer`.

It does add complexity though and I am ignoring that for now and will load & pass the entire file contents to consumers, just wrapping them to be readable

crates/atlaspack_filesystem/src/file_system.rs
alshdavid: I'm more questioning if this is the right place for it.

Considering the FileSystem is just a map of filepaths -> contents, it has no concept of the current position of the user so it feels weird that it's a property of the filesystem itself.

Surely the user tracks their own cwd?

crates/atlaspack_filesystem/src/file_system.rs
alshdavid: There is usage in the form of `FileSystemExt.exists()`, which must built on top of `FileType`

crates/atlaspack_filesystem/src/file_system.rs
alshdavid: I placed all of the `std::fs` methods in the trait to ensure we do not add custom methods in future (or if we do, they get added to `FileSystemExt`).

They are unimplemented by default because it gives us visibility of the fs methods during development (intellisense), and when failures occur at runtime, we are prompted to satisfy the implementations.  

`OsFileSystem` forwards all the `FileSystem` methods to `std::fs` because that's easy to do. `InMemoryFileSystem`, implements them as-required because that's harder.

Lastly, minor point but, having all of the methods allows the `atlaspack_filesystem` crate to be used by plugin developers and other Rust projects within Atlassian, helping with the interoperability of Atlaspack.

crates/atlaspack_config/src/atlaspack_rc_config_loader.rs
yamadapc: It should have the same output 🤞, but it's trying to minimize nº of syscalls (should have less `realpath` calls once the caches are populated). Whereas `std::fs::canonicalize` is just `libc::realpath`.

packages/utils/node-resolver-rs/benches/node_resolver_bench.rs
yamadapc: This is just example code

yamadapc (changes requested):
Please remove methods we aren't using or that aren't implemented for all back-ends.

crates/atlaspack_config/src/atlaspack_rc_config_loader.rs
alshdavid: If it has the same signature, even if the implementation uses a cache internally, I feel it should not have it's own method.

Unless you're advocating for adding something like `FileSystemExt.canonicalize_cached()`? Or perhaps a `CachedFileSystem` that wraps `FileSystem` and can be applied to all of the fs implementations?

alshdavid:
> Please remove methods we aren't using or that aren't implemented for all back-ends.

I disagree with this request for the reasons stated in this [reply](https://github.com/atlassian-labs/atlaspack/pull/50#discussion_r1740222101). Can you please elaborate on why my points are invalid in this case?

crates/atlaspack_filesystem/src/in_memory/file_system.rs
mattcompiles: Where do we use this in a plugin that we need to support thought? Just cos you can do it today doesn't mean we should just blindly support it.

alshdavid: I'm not adding it into plugins atm, this is just here for completeness and for future reference.

Plugins will get the complete binary contents of their target file and `Asset.getStream()` will just return a `Reader` that reads over the entire blob

yamadapc:
There's no benefit in supporting intellisense to functions that are unimplemented and will lead to crashes at runtime.

Please either:

* implement the functions for all back-ends
* remove functions that we aren't using

crates/atlaspack_filesystem/src/utils/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod find_ancestor;
```
MonicaOlejniczak, line 1: Can you revert this? `utils` is not as descriptive as `search` and sucks as a name

crates/atlaspack_filesystem/src/file_system.rs
MonicaOlejniczak: It's mostly for convenience, and because the original node filesystem contains cwd

crates/atlaspack_filesystem/src/file_system.rs
MonicaOlejniczak: Personally I don't think we need to create a filesystem implementation that matches the built-in rust impl 1:1. It's meant to be an atlaspack filesystem, so we should put things in the trait that make sense to be in there.

crates/atlaspack_filesystem/src/in_memory/canonicalize.rs
```diff
@@ -0,0 +1,35 @@
+use std::path::Component;
+use std::path::Path;
+use std::path::PathBuf;
+
+use parking_lot::RwLock;
+
+pub fn canonicalize_impl(current_working_directory: &RwLock<PathBuf>, path: &Path) -> PathBuf {
```
MonicaOlejniczak, line 7: I would remove `_impl`, it's a standalone function

crates/atlaspack_filesystem/src/file_system.rs
alshdavid: My thinking is that, assuming we are expecting to migrate a bunch of plugins to native, we'd expect some external contributions. 

Having a filesystem interface that aligns with the standard library reduces friction during development for those contributors - plus IMO it's easier to just do what the standard lib does rather than inventing our own call signatures.

crates/atlaspack_filesystem/src/lib.rs
MonicaOlejniczak: Where will the boundary be for the cache? My concern is that it's unclear when the cache will be used and for which methods on the file system

PR #52, opened 2024-08-31
[V3] Update JSX test to not use babel

mattcompiles:
## Motivation

Previously this tests relied of babel to parse the JSX. Updated to use the core transformer so it passes against V3. 

## Checklist


packages/core/integration-tests/test/scope-hoisting.js
```diff
@@ -2278,17 +2278,19 @@ describe('scope hoisting', function () {
...
       let output = await run(b);
-      assert.deepEqual(output, {
-        children: 'Test',
-        props: null,
-        type: 'span',
-      });
+
+      assert.equal(output.children, 'Test');
+      assert.equal(output.type, 'span');
+      assert.equal(output.props.test, 'working');
```
mattcompiles, line 2293: I updated the assertions as the new jsx stuff inserts debug stuff that we don't need to validate here.

yamadapc (approved):

MonicaOlejniczak (approved):

PR #53, opened 2024-09-02
Refactor bundle-text, json, toml, and yaml tests

MonicaOlejniczak:
## Motivation

The JavaScript integration tests are extremely large and difficult to parse through. It also includes some tests that fail in V3, that can be extracted into their own test suites since they require additional plugins to function correctly.

## Changes

* Extract bundle-text, json, toml, and yaml tests into separate test files and refactor them into fs fixtures
* Remove an unnecessary `.babelrc` and use the swc transformer instead

## Checklist


mattcompiles (approved):
Nice one

PR #54, opened 2024-09-02
[V3] Read js transformer package config

MonicaOlejniczak:
## Motivation

The Rust JS transformer currently does not read its own package config, which is a piece of functionality that exists in the JavaScript version

## Changes

* Read `@atlaspack/transformer-js` config from the closest `package.json`, later this should be in the `.atlaspackrc` config imo but is out of scope
* Pass through missing env from input options to rust via `resolvedOptions`
* Update environment variable handling to match the logic in `JSTransformer` and factor in `inline_environment`
* Add missing `inline_fs` and `unstable_inlineConstants` handling
* Add missing `package.json` files expected to exist in the transformer (note that when this fails, it generally indicates an invalid state in tests)
* Enable a few tests for v3

## Checklist


crates/atlaspack_plugin_transformer_js/src/js_transformer.rs
```diff
@@ -124,19 +138,21 @@ impl TransformerPlugin for AtlaspackJsTransformerPlugin {
...
           .unwrap_or_default(),
-        env: self
-          .options
-          .env
-          .clone()
-          .unwrap_or_default()
-          .iter()
-          .map(|(key, value)| (key.as_str().into(), value.as_str().into()))
-          .collect(),
+        env: filter_env_variables(
```
mattcompiles, line 141: Not a blocker, but we could potentially run this only once per build rather than per file. Would just need two configs, one for `is_source: true` and one for `is_source: false`.

mattcompiles (approved):
Nice work 👏

crates/atlaspack_plugin_transformer_js/src/js_transformer.rs
MonicaOlejniczak: Sure

PR #55, opened 2024-09-04
Add no-op importCond API

JakeLane:
## Motivation

Add the conditional bundling API. By adding the API outside of the feature flag, we can start with simple no-op behaviour (making the return value the false branch of the condition). I'll follow up with iterative PRs to add the conditional behaviour behind the feature flag.

## Changes

New import called `importCond` which (when implemented) will allow developers to conditionally load _sync_ dependencies based on the result of the condition param.

### DIFF
--- a/packages/core/conditional-import-types/index.d.ts
+++ b/packages/core/conditional-import-types/index.d.ts
@@ -5,6 +5,8 @@ type ErrorMessage =
 type ConditionalImport<CondT, CondF> = CondT | CondF;
 
 /**
+ * **IMPORTANT: This API is currently a no-op. Do not use until this message is removed.**
+ *
  * Conditionally import a dependency, based on the specified condition.
  *
  * This is a synchronous import that differs from conditionally loading a dynamic import (`import()`)
--- a/packages/core/conditional-import-types/index.d.ts
+++ b/packages/core/conditional-import-types/index.d.ts
@@ -1,8 +1,15 @@
-type ModuleRef<_> = string;
-type ErrorMessage =
-  "You must annotate type with \"<typeof import('a'), typeof import('b')>\"";
+type ModuleRef<_PhantomModuleType> = string;
 
-type ConditionalImport<CondT, CondF> = CondT | CondF;
+type NoImportErrorMessage =
+  "You must annotate type with \"<typeof import('a'), typeof import('b')>\"";
+type NoDefaultErrorMessage = 'Conditional imports must have a default export';
+
+type ESModuleWithDefaultExport = {[key: string]: any; default: any};
+
+type ConditionalImport<
+  CondT extends ESModuleWithDefaultExport,
+  CondF extends ESModuleWithDefaultExport,
+> = CondT['default'] | CondF['default'];
 
 /**
  * **IMPORTANT: This API is currently a no-op. Do not use until this message is removed.**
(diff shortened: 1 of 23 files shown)


packages/core/conditional-import-types/index.d.ts
```diff
@@ -0,0 +1,28 @@
+type ModuleRef<_> = string;
```
tomgasson, line 1:
```suggestion

type ModuleRef<_PhantomModuleType> = string;
```

packages/core/conditional-import-types/index.d.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * This is a synchronous import that differs from conditionally loading a dynamic import (`import()`)
+ *
+ * This function requires server to guarantee the dependency is loaded.
+ *
+ * @param condition Condition evaluated by the server
+ * @param ifTrueDependency Dependency returned if the condition is true
+ * @param ifFalseDependency Dependency returned if the condition is false
+ */
+declare function importCond<CondT, CondF>(
```
tomgasson, line 20: If we look at the type of the _entire module_, it's a very confusing problem. They'll have to exactly match (all exports) types to be the same, else they'll narrow differently depending on the access pattern. What we actually want is to make sure the _one value_ (the `default export`) matches type.

```
// IfTrue
export const a = "test"
export const b = 123
```

```
// IfFalse
export const a = "test"
export const b = "other"
```
What's the types of these?

```
{a: string, b: number} | {a: string, b: string}
// or
{ a: string, b: number | string }
```


```
// IfTrue
export const a = "test"
export const b = 123
export default 123
```

```
// IfFalse
export const a = "test"
export const b = "other"
export default "hello"
```
What's the types of these default exports?

```
number | string
```
And we're doing _value refinement_ directly. Must simpler.


We'll need to ensure that the target of the import is an _open_ module which has a default export (that's the value we're after), otherwise we'll later face a problem of not having the place to do `theModule => theModule.namedExport` unwrapping.

- Open module: means use `interface ESModuleWithDefaultExport<T> { default: T }` not `type ESModuleWithDefaultExport<T> = { default: T } & Record<PropertyKey, never>`.
- We want them to be able to export _whatever they want_ without restriction, differently from each side. Only the _actual thing we're after_ (the default export) is what we care about
- You'll notice in the below code how it's forced you to name both things `.Feature` in each side. That's a whole extra constraint we don't want, where _the caller_ would implicate the _the definition_. Simpler to use a default name ("default") universally, since that constraint will be viral anyway.


```suggestion
declare function importCond<CondT extends ESModuleWithDefaultExport<any> | void = void, CondF extends ESModuleWithDefaultExport<any> | void = void>(

...

): CondT.default | CondF.default
```

packages/transformers/js/core/src/dependency_collector.rs
```diff
@@ -422,14 +422,35 @@ impl<'a> Fold for DependencyCollector<'a> {
...
       Callee::Import(_) => DependencyKind::DynamicImport,
       Callee::Expr(expr) => {
         match &**expr {
+          Ident(ident)
+            if !self.config.conditional_bundling
+              && ident.sym.to_string().as_str() == "importCond" =>
+          {
+            // Conditional bundling disabled, treat like a require import
+
+            // Change false dep to the require arg
```
tomgasson, line 441: So this transforms
```tsx
return importCond<
    typeof import('./async-feature-enabled'),
    typeof import('./async-feature-disabled')
  >(
    'feature.async.condition',
    './async-feature-enabled.ts',
    './async-feature-disabled.ts',
  )
```
into
```tsx
return require('./async-feature-disabled.ts')
```
Correct?

Next step after this might be to replace it with actually the equivalent _runtime_ behaviour
```tsx
return ff('feature.async.condition') ? require('./async-feature-enabled.ts') : require('./async-feature-disabled.ts')
```

tomgasson (approved):

packages/transformers/js/core/src/dependency_collector.rs
JakeLane: Yep we end up with `return require('./async-feature-disabled.ts')`

For the next step, I'm undecided if that's the next approach as that requires context to jira-frontend which is not available in github. It's probably not worth the trouble to handle this condition unless we want to allow conditions from multiple sources (not just statsig)?

We could have something like:

```
const conditions = {
   'feature.async.condition': () => ff('feature.async.condition')
} as const;
```

which is compiled into the server template

PR #56, opened 2024-09-04
[V3] Add inline string transformer

MonicaOlejniczak:
## Motivation

Some tests fail because we do not have an inline string transformer. These changes port the transformer to Rust and fix some pre-existing bugs that were highlighted with the change.

## Changes

* Add `atlaspack_plugin_transformer_inline_string` crate
* Update plugins to use the rust inline string transformer when `@atlaspack/transformer-inline-string` is encountered
* Add missing `pipeline` to assets
* Add in missing `bundle_behavior` to dependencies
* Remove `has_symbols` in favour of `Option<Vec<Symbol>>` since there were a lot of inconsistencies and bugs around when it was set
* Use `source_asset_id` instead of `source_asset_path` for the hash to match v2
* Enable some `bundle-text:` tests, and refactor a couple from the JavaScript suite into here

## Checklist

### DIFF
--- a/crates/atlaspack_plugin_transformer_js/src/js_transformer/conversion.rs
+++ b/crates/atlaspack_plugin_transformer_js/src/js_transformer/conversion.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 
 use atlaspack_core::diagnostic;
 use indexmap::IndexMap;
-use swc_core::atoms::Atom;
+use swc_core::atoms::{Atom, JsWord};
 
 use atlaspack_core::plugin::{PluginOptions, TransformResult};
 use atlaspack_core::types::engines::EnvironmentFeature;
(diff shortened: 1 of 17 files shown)


crates/atlaspack_plugin_transformer_js/src/js_transformer/conversion.rs
```diff
@@ -44,56 +44,64 @@ pub(crate) fn convert_result(
...
     );
     dependency_by_specifier.insert(dependency.specifier.as_str().into(), dependency);
   }
 
+  let mut asset_symbols = Vec::new();
+
   if let Some(hoist_result) = result.hoist_result {
     // Has symbols is currently needed to differentiate between assets with no symbols vs assets
     // which haven't had symbols analyzed yet.
     // TODO: replace `asset.symbols` with `Option<Vec<...>>`
```
mattcompiles, line 61: We should remove this comment now

mattcompiles (approved):
Awesome work!

crates/atlaspack_plugin_transformer_js/src/js_transformer/conversion.rs
MonicaOlejniczak: Oh that's where it went! I was looking for it earlier

PR #58, opened 2024-09-05
[V3] Add json transformer

MonicaOlejniczak:
## Motivation

These changes port the json transformer to Rust so that more functionality works in v3

## Changes

* Add `atlaspack_plugin_transformer_json` crate
* Use json transformer when `@atlaspack/transformer-json` is encountered
* Enable json and json5 integration tests

## Checklist

### DIFF
--- a/crates/atlaspack/src/plugins/config_plugins.rs
+++ b/crates/atlaspack/src/plugins/config_plugins.rs
@@ -31,6 +31,7 @@ use atlaspack_plugin_rpc::plugin::RpcTransformerPlugin;
 use atlaspack_plugin_rpc::RpcWorkerRef;
 use atlaspack_plugin_transformer_inline_string::AtlaspackInlineStringTransformerPlugin;
 use atlaspack_plugin_transformer_js::AtlaspackJsTransformerPlugin;
+use atlaspack_plugin_transformer_json::AtlaspackJsonTransformerPlugin;
 use atlaspack_plugin_transformer_raw::AtlaspackRawTransformerPlugin;
 
 use super::Plugins;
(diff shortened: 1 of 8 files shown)


crates/atlaspack/src/plugins/config_plugins.rs
```diff
@@ -224,6 +225,11 @@ impl Plugins for ConfigPlugins {
         continue;
       }
 
+      if transformer.package_name == "@atlaspack/transformer-json" {
```
mattcompiles, line 228: Probably need something more elegant for this at some stage 😅

crates/atlaspack_plugin_transformer_json/src/json_transformer.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl TransformerPlugin for AtlaspackJsonTransformerPlugin {
+  fn transform(&mut self, asset: Asset) -> Result<TransformResult, Error> {
+    let mut asset = asset.clone();
+
+    let code = std::str::from_utf8(asset.code.bytes())?;
+    let code = json5::from_str::<serde_json::Value>(code)?;
+    let code = json5::to_string(&code)?;
+
+    asset.code = Arc::new(Code::from(format!("module.exports = {code};")));
```
mattcompiles, line 25: The old transformer uses `JSON.parse` here. I think we should stick with that as it's actually faster for browsers that way.

MonicaOlejniczak: Sure, it requires another dependency tho

mattcompiles (approved):

PR #61, opened 2024-09-05
Add force skipping for wrapped dependencies

marcins:
## Motivation

In the scope hoisting packager, when an asset is reachable through a lazy (async) dependency, that asset is "wrapped" - that it has a `parcelRegister('pubid', ...)` put around it. Assets that are wrapped won't be scope hoisted into their parent asset.

Asset wrapping is "viral", in that in propagates to all child assets of the wrapped asset within that bundle.

In the case of the Jira route map, if we make that async it effectively becomes "de-optimised".. because it's a giant object with 500+ properties for all the different routes, these additional register/require calls in this case have a _negative_ impact on performance (normally "lazy require" might be faster, but in this case everything ends up being required to produce the object).

We know (or think we do) that the "async route map" will be the only place these routes are defined, and that only the top level `getRoute` will ever be referenced, so in theory it _is_ safe to allow the child assets of the route map to be scope hoisted.

## Changes

In order to allow us to hint to the packager that "we (think we) know what we're doing", I have added an `unstable_forceSkipWrapAssets` config property where an array of relative asset paths can be specified. If an asset in this array is reached while "virally wrapping assets", then the traversal will be skipped.

Note this doesn't affect assets that were _directly_ added to the wrapped set because they met the conditions, only child assets.

## Checklist


packages/packagers/js/src/index.js
```diff
@@ -66,6 +73,8 @@ export default (new Packager({
       unstable_asyncBundleRuntime: Boolean(
         conf?.contents?.unstable_asyncBundleRuntime,
       ),
+      unstable_forceSkipWrapAssets:
```
mattcompiles, line 76: I'd suggest a feature flag rather than an unstable config?

marcins: I considered that, but still need to be able to configure the list of paths though? So seems redundant to have both.

mattcompiles: Right I getcha

JakeLane (approved):

packages/core/integration-tests/test/scope-hoisting.js
```diff
@@ -6317,4 +6317,90 @@ describe('scope hoisting', function () {
       assert.fail('Expected the empty bundle to still run');
     }
   });
+
+  // Will work in v3 once HTML entries are supported
```
MonicaOlejniczak, line 6321: /nit comment is no longer relevant

MonicaOlejniczak (approved):

rust-toolchain
```diff
@@ -1 +1 @@
-stable
```
yamadapc, line 1: Why was this pinned @marcins?

marcins: `1.81` failed to build so "last known good" was pinned at the time.

PR #62, opened 2024-09-05
[V3] Add basic Node plugin support

mattcompiles:
## Motivation

## Changes

## Checklist

### DIFF
--- a/crates/atlaspack_core/src/types/asset.rs
+++ b/crates/atlaspack_core/src/types/asset.rs
@@ -3,7 +3,6 @@ use std::hash::{Hash, Hasher};
 use std::num::NonZeroU32;
 use std::path::PathBuf;
 use std::sync::Arc;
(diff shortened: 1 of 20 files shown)


crates/atlaspack_core/src/types/asset.rs
```diff
@@ -30,6 +29,10 @@ impl Code {
     self.inner.as_bytes()
   }
 
+  pub fn string(&self) -> String {
```
MonicaOlejniczak, line 32: Just noting that this format might need to change for binaries / images, which I was looking into soon

MonicaOlejniczak: I merged this change in, so it will affect what you have somewhat

mattcompiles:
Closing this as I'm splitting it up into separate PRs

PR #64, opened 2024-09-06
[V3] Add image transformer

MonicaOlejniczak:
## Motivation

These changes port the image transformer to Rust so that more functionality works in v3. It also removes the overhead of calling into JavaScript plugins by using Rust directly.

## Changes

* Represent `Code` type as bytes instead of a string for binary data
* Add `atlaspack_plugin_transformer_image` crate which supports most of the pre-existing types in the original image transformer. Note that the sharp config file is not supported since it uses a different underlying implementation. The `image` dependency may be revisited at a later time if necessary.
* Use image transformer when `@atlaspack/transformer-image` is encountered
* Add supported image file types to `FileType`
* Add `read` function to `FileSystem` trait to handle binary data when reading an asset file
* Add missing query to resolver resolution
* Update `readFile` in JavaScript filesystem to handle binary data when there is no encoding
* Enable relevant tests and expand test cases

## Checklist

### DIFF
--- a/crates/atlaspack_plugin_transformer_image/src/image_transformer.rs
+++ b/crates/atlaspack_plugin_transformer_image/src/image_transformer.rs
@@ -1,0 +1,138 @@
+use std::io::Cursor;
+use std::sync::Arc;
+
+use anyhow::Error;
+use atlaspack_core::diagnostic_error;
+use atlaspack_core::plugin::TransformResult;
+use atlaspack_core::plugin::{PluginContext, TransformerPlugin};
+use atlaspack_core::types::{Asset, BundleBehavior, Code, FileType};
+use image::imageops::FilterType;
+use image::{ImageFormat, ImageReader};
+use url_search_params::parse_url_search_params;
+
+#[derive(Debug)]
+pub struct AtlaspackImageTransformerPlugin {}
+
+impl AtlaspackImageTransformerPlugin {
+  pub fn new(_ctx: &PluginContext) -> Result<Self, Error> {
+    Ok(AtlaspackImageTransformerPlugin {})
+  }
+}
+
+impl TransformerPlugin for AtlaspackImageTransformerPlugin {
+  fn transform(&mut self, asset: Asset) -> Result<TransformResult, Error> {
+    let mut asset = asset.clone();
+
+    if asset.bundle_behavior == BundleBehavior::None {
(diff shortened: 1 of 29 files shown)


crates/atlaspack_plugin_transformer_image/src/image_transformer.rs
```diff
@@ -0,0 +1,138 @@
...
+      } else if let Some(height) = height {
+        img.resize(img.width(), height, filter)
+      } else {
+        img
+      };
+
+      let mut bytes: Vec<u8> = Vec::new();
+      img.write_to(&mut Cursor::new(&mut bytes), target_format)?;
+
+      asset.code = Arc::new(Code::new(bytes));
```
yamadapc, line 69: We should refactor things so that these bytes are not kept in memory. We should read images on this file and then ideally flush them out.

packages/core/core/src/atlaspack-v3/fs.js
```diff
@@ -15,9 +15,14 @@ export function toFileSystemV3(fs: ClassicFileSystem): FileSystem {
     canonicalize: jsCallable((path: FilePath) => fs.realpathSync(path)),
     createDirectory: jsCallable((path: FilePath) => fs.mkdirp(path)),
     cwd: jsCallable(() => fs.cwd()),
-    readFile: jsCallable((path: string, encoding?: Encoding) =>
-      fs.readFileSync(path, encoding ?? 'utf8'),
-    ),
+    readFile: jsCallable((path: string, encoding?: Encoding) => {
+      if (!encoding) {
+        // $FlowFixMe
+        return [...fs.readFileSync(path)];
```
yamadapc, line 21: I think if we convert the buffer to array buffer / typed array etc. then there's no need to copy

yamadapc (approved):

crates/atlaspack_plugin_transformer_image/src/image_transformer.rs
MonicaOlejniczak: I'll add it as an improvement item

packages/core/core/src/atlaspack-v3/fs.js
MonicaOlejniczak: Hmm, I tried the underlying buffer which is an array buffer and forcing a uint8 array but they both fail. Will leave it for now.

PR #66, opened 2024-09-08
Add initial rust inline requires optimiser implementation

yamadapc:
This change adds a rust implementation of the inline requires optimiser. This should avoid passing the AST back and forward to rust ; also since this is operating on large output files, the implementation might be a bit more efficient.

I've changed parts of the core SWC transformer wrapper code so that it can be used as a testing library on the new optimiser crate and so that it can run outputting source-maps.

At the moment there's waste, and even if source-maps aren't enabled they'll be partially generated. Some source-map file-naming might need to be patched as well as an anonymous file name is used.

## Motivation

The motivation for this change is to improve performance of the packaging phase of bundling.

## Checklist

There are no integration tests for inline requires that I could find.

There's a minor difference in output code as I'm not wrapping the inlined expressions in `(0, ...)`

### DIFF
--- a/crates/atlaspack_plugin_optimizer_inline_requires/src/lib.rs
+++ b/crates/atlaspack_plugin_optimizer_inline_requires/src/lib.rs
@@ -1,7 +1,7 @@
 mod inlining_visitor;
 
-use std::collections::HashSet;
 use crate::inlining_visitor::IdentifierReplacementVisitor;
+use std::collections::HashSet;
 use swc_core::atoms::atom;
 use swc_core::atoms::Atom;
 use swc_core::common::Mark;
@@ -330,7 +330,7 @@ impl VisitMut for InlineRequiresOptimizer {
 
 #[cfg(test)]
 mod tests {
-  use atlaspack_js_swc_core::test_utils::{run_visit, RunVisitResult};
+  use atlaspack_js_swc_core::test_utils::{run_test_visit, RunVisitResult};
 
   use super::*;
 
@@ -344,7 +344,7 @@ function doWork() {
     "#
     .trim();
 
-    let RunVisitResult { output_code, .. } = run_visit(code, |ctx| InlineRequiresOptimizer {
+    let RunVisitResult { output_code, .. } = run_test_visit(code, |ctx| InlineRequiresOptimizer {
       unresolved_mark: ctx.unresolved_mark,
       ..Default::default()
     });
@@ -373,7 +373,7 @@ parcelRequire.register('moduleId', function(require, module, exports) {
     "#
     .trim();
 
-    let RunVisitResult { output_code, .. } = run_visit(code, |ctx| InlineRequiresOptimizer {
+    let RunVisitResult { output_code, .. } = run_test_visit(code, |ctx| InlineRequiresOptimizer {
       unresolved_mark: ctx.unresolved_mark,
       ..Default::default()
     });
@@ -397,7 +397,7 @@ const parcelHelpers = require("@atlaspack/transformer-js/src/esmodule-helpers.js
     "#
     .trim();
 
-    let RunVisitResult { output_code, .. } = run_visit(code, |ctx| InlineRequiresOptimizer {
+    let RunVisitResult { output_code, .. } = run_test_visit(code, |ctx| InlineRequiresOptimizer {
       unresolved_mark: ctx.unresolved_mark,
       ..Default::default()
     });
@@ -416,7 +416,7 @@ const sideEffects = require("side-effects");
     "#
     .trim();
 
-    let RunVisitResult { output_code, .. } = run_visit(code, |ctx| InlineRequiresOptimizer {
+    let RunVisitResult { output_code, .. } = run_test_visit(code, |ctx| InlineRequiresOptimizer {
       unresolved_mark: ctx.unresolved_mark,
       ignore_patterns: vec![IgnorePattern::ModuleId(atom!("side-effects"))],
       ..Default::default()
@@ -441,7 +441,7 @@ function run() {
     "#
     .trim();
 
-    let RunVisitResult { output_code, .. } = run_visit(code, |ctx| InlineRequiresOptimizer {
+    let RunVisitResult { output_code, .. } = run_test_visit(code, |ctx| InlineRequiresOptimizer {
       unresolved_mark: ctx.unresolved_mark,
       ..Default::default()
     });
--- a/crates/atlaspack_plugin_optimizer_inline_requires/src/lib.rs
+++ b/crates/atlaspack_plugin_optimizer_inline_requires/src/lib.rs
@@ -1,0 +1,459 @@
+mod inlining_visitor;
+
+use std::collections::HashSet;
+use crate::inlining_visitor::IdentifierReplacementVisitor;
+use swc_core::atoms::atom;
+use swc_core::atoms::Atom;
+use swc_core::common::Mark;
+use swc_core::ecma::ast::{CallExpr, Expr, Id, Ident, Lit, VarDecl, VarDeclarator};
+use swc_core::ecma::utils::ExprExt;
+use swc_core::ecma::visit::{VisitMut, VisitMutWith};
+
+/// Represents a `const i = require('module-id')` statement that has been found.
+#[derive(Debug)]
+pub struct RequireInitializer {
+  /// The variable `i` swc [`Id`] for matching it respecting scope
+  pub variable_id: Id,
+  /// The imported package atom `'module-id'`
+  pub imported_package: Atom,
+  /// The entire `require(...)` call for replacement
+  pub call_expr: CallExpr,
+}
+
+/// Default match patterns for parcel
+fn default_require_matchers() -> Vec<RequireMatcher> {
+  vec![
+    RequireMatcher::Global(atom!("require")),
+    RequireMatcher::Global(atom!("parcelRequire")),
+  ]
+}
+
+/// Default ignore patterns for parcel
+fn default_ignore_patterns() -> Vec<IgnorePattern> {
+  vec![IgnorePattern::IdentifierSymbol(atom!("parcelHelpers"))]
+}
+
+/// Extract [RequireInitializer] information from a declarator or return None.
+fn match_require_initializer(
+  decl: &VarDeclarator,
+  unresolved_mark: Mark,
+  require_matchers: &[RequireMatcher],
+  ignore_patterns: &[IgnorePattern],
+) -> Option<RequireInitializer> {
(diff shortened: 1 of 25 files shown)


packages/core/rust/test/InlineRequiresNative.test.js
```diff
@@ -0,0 +1,32 @@
+// @flow strict-local
+
+import assert from 'assert';
+import {runInlineRequiresOptimizer} from '..';
+
+describe('runInlineRequiresOptimizer', () => {
+  it('replaces inline code on source', () => {
```
MonicaOlejniczak, line 7: You should consider adding a test for synchronous `import`

packages/optimizers/inline-requires/src/InlineRequires.js
```diff
@@ -57,6 +59,26 @@ module.exports = new Optimizer<empty, BundleConfig>({
...
+        });
+        const sourceMapResult = result.sourceMap;
+        if (sourceMapResult != null) {
+          sourceMap = new SourceMap(options.projectRoot);
+          sourceMap.addVLQMap(JSON.parse(sourceMapResult));
+          if (originalMap) {
+            sourceMap.extends(originalMap);
+          }
+        }
+        return {contents: result.outputCode, map: originalMap};
```
MonicaOlejniczak, line 79: /nit `.code` is pretty common for transformations rather than `outputCode`

packages/optimizers/inline-requires/src/InlineRequires.js
```diff
@@ -57,6 +59,26 @@ module.exports = new Optimizer<empty, BundleConfig>({
     }
 
     try {
+      if (getFeatureFlag('fastOptimizeInlineRequires')) {
+        let sourceMap = null;
+        const result = runInlineRequiresOptimizer({
```
MonicaOlejniczak, line 64: Did you consider or try out how much faster this would be if it were part of the transformation phase?

packages/optimizers/inline-requires/src/InlineRequires.js
```diff
@@ -57,6 +59,26 @@ module.exports = new Optimizer<empty, BundleConfig>({
...
+          inputCode: contents.toString(),
```
MonicaOlejniczak, line 65: /nit `code` is sufficient as an input or it could be the first argument

packages/transformers/js/core/src/runner.rs
```diff
@@ -0,0 +1,187 @@
...
+    })?;
+  Ok(RunVisitResult {
+    output_code,
+    visitor,
+    source_map,
+  })
+}
+
+/// Same as `run_visit` but for `Fold` instances
+#[allow(unused)]
```
MonicaOlejniczak, line 53: It shouldn't be unused since it's used in the tests?

packages/transformers/js/core/src/runner.rs
```diff
@@ -0,0 +1,187 @@
...
+    })?;
+  Ok(RunVisitResult {
+    output_code,
+    visitor,
+    source_map,
+  })
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum RunWithTransformationError {
```
MonicaOlejniczak, line 72: A lot of these don't look used?

packages/transformers/js/core/src/runner.rs
```diff
@@ -0,0 +1,187 @@
...
+  #[error("Failed to parse")]
```
MonicaOlejniczak, line 73: Could be more descriptive, `Failed to parse module`

packages/transformers/js/core/src/runner.rs
```diff
@@ -0,0 +1,187 @@
...
+  SwcParse(swc_ecma_parser::error::Error),
+  #[error("IO Error: {0}")]
+  IoError(#[from] std::io::Error),
+  #[error("Non-utf 8 output: {0}")]
+  NonUtfOutput(#[from] FromUtf8Error),
+  #[error("Source-map failure")]
```
MonicaOlejniczak, line 79: `Source map` without the dash, also can be worded a bit better. Something like `Failed to generate source map`

packages/transformers/js/core/src/runner.rs
```diff
@@ -0,0 +1,187 @@
...
+  })
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum RunWithTransformationError {
+  #[error("Failed to parse")]
+  SwcParse(swc_ecma_parser::error::Error),
+  #[error("IO Error: {0}")]
+  IoError(#[from] std::io::Error),
+  #[error("Non-utf 8 output: {0}")]
```
MonicaOlejniczak, line 77: This can be structured a bit better, depending on the exact context.

```suggestion
  #[error("Invalid utf-8 output: {0}")]
  InvalidUtf8Output(#[from] FromUtf8Error),
```

crates/node-bindings/src/optimizers/inline_requires_optimizer.rs
```diff
@@ -0,0 +1,49 @@
...
+        input
+          .assets_to_ignore
+          .into_iter()
+          .map(|s| Atom::new(s))
+          .collect(),
+      ))
+      .build();
+    visitor
+  })
+  .map_err(|err| napi::Error::from_reason(format!("[napi] failed to run optimizer: {}", err)))?;
```
MonicaOlejniczak, line 36:
```suggestion
  .map_err(|err| napi::Error::from_reason(format!("[napi] Failed to run inline require optimizer: {}", err)))?;
```

crates/node-bindings/src/optimizers/inline_requires_optimizer.rs
```diff
@@ -0,0 +1,49 @@
...
+
+  Ok(InlineRequiresOptimizerResult {
+    output_code: result.output_code,
+    source_map: if input.source_maps {
+      let source_map = String::from_utf8(result.source_map).map_err(|err| {
+        napi::Error::from_reason(format!("[napi] Non-utf8 source-map output: {}", err))
```
MonicaOlejniczak, line 42:
```suggestion
        napi::Error::from_reason(format!("[napi] Invalid utf-8 source map output: {}", err))
```

crates/node-bindings/src/optimizers/inline_requires_optimizer.rs
```diff
@@ -0,0 +1,49 @@
+use atlaspack_js_swc_core::runner::run_visit;
+use atlaspack_plugin_optimizer_inline_requires::IgnorePattern;
+use napi_derive::napi;
+use swc_core::atoms::Atom;
+
+#[napi(object)]
+pub struct InlineRequiresOptimizerInput {
+  pub input_code: String,
+  pub source_maps: bool,
+  pub assets_to_ignore: Vec<String>,
```
MonicaOlejniczak, line 10: `assets_to_ignore` is a bit too specific imo. Maybe name it more closely to the ignore pattern type like `ignore_module_ids` or perhaps change it to a generic `ignore_patterns` delineated by `type`

crates/atlaspack_plugin_optimizer_inline_requires/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+  /// The variable `i` swc [`Id`] for matching it respecting scope
+  pub variable_id: Id,
+  /// The imported package atom `'module-id'`
+  pub imported_package: Atom,
+  /// The entire `require(...)` call for replacement
+  pub call_expr: CallExpr,
+}
+
+/// Default match patterns for parcel
+fn default_require_matchers() -> Vec<RequireMatcher> {
```
MonicaOlejniczak, line 24: How come you opted for a function instead of implementing `Default` when it's in this crate? Same for ignore patterns

packages/optimizers/inline-requires/src/InlineRequires.js
yamadapc: I think that'd be ideal but quite a bit more work because it'd change how the current hoisting transformer / packager work.

packages/transformers/js/core/src/runner.rs
yamadapc: They are used ; just they are cast automatically from the functions that error out

packages/core/rust/test/InlineRequiresNative.test.js
yamadapc: I'm not sure if I'm handling ESM imports at all because we can't inline them, so it's assumed we're already on a later transformation phase.

packages/optimizers/inline-requires/src/InlineRequires.js
marcins: > Did you consider or try out how much faster this would be if it were part of the transformation phase?

More likely you'd implement it directly in the `ScopeHoistingPackager` where all the import variable stuff is already done today. At the time I avoided this as it was complex and we needed to get a solution in, but that would avoid an AST parse entirely (though it would be more of the same regex / find and replace magic the existing packager already does).

packages/core/rust/test/InlineRequiresNative.test.js
MonicaOlejniczak: I thought the JS version changed imports, but maybe since it's in optimising phase it's okay?

MonicaOlejniczak (approved):

packages/transformers/js/core/src/runner.rs
```diff
@@ -0,0 +1,186 @@
...
+      let mut module = module.fold_with(&mut resolver(unresolved_mark, global_mark, false));
+
+      let context = RunContext {
+        source_map: source_map.clone(),
+        global_mark,
+        unresolved_mark,
+      };
+      let result = transform(context, &mut module);
+
+      let mut line_pos_buffer = vec![];
```
marcins, line 119: Currently after the inline requires optimiser runs, the next optimiser is the SWC minifier, which again has to parse the bundle to minify it. If this code supported minified output (and could use the same options as the existing `@atlaspack/optimizer-swc`) you could avoid another parse/print pass by removing that optimiser from the config when the inline requires optimiser is used.

packages/transformers/js/core/src/runner.rs
```diff
@@ -0,0 +1,186 @@
...
+      let mut output_buffer = vec![];
+      let writer = JsWriter::new(
```
marcins, line 121: Currently after the inline requires optimiser runs, the next optimiser is the SWC minifier, which again has to parse the bundle to minify it. If this code supported minified output (and could use the same options as the existing `@atlaspack/optimizer-swc`) you could avoid another parse/print pass by removing that optimiser from the config when the inline requires optimiser is used.

PR #74, opened 2024-09-12
Use inline requires when importCond flag is disabled

JakeLane:
## Motivation

Allows usage of importCond in jira with support for a condition. This allows us to actually use `importCond` functionally in production, without the bundling implementation.

## Changes

- New module `ConditionalImportsFallback` to visit earlier than the dependency collector

### DIFF
--- a/packages/transformers/js/core/src/conditional_imports_fallback.rs
+++ b/packages/transformers/js/core/src/conditional_imports_fallback.rs
@@ -1,0 +1,123 @@
+use swc_core::common::{Mark, DUMMY_SP};
+use swc_core::ecma::ast::{
+  ComputedPropName, CondExpr, Expr, MemberExpr, MemberProp, OptChainExpr, ParenExpr,
+};
+use swc_core::ecma::visit::VisitMut;
+
+use crate::utils::{create_require, match_str};
+use crate::Config;
+
+pub struct ConditionalImportsFallback<'a> {
+  pub config: &'a Config,
+  pub unresolved_mark: Mark,
+}
+
+impl<'a> VisitMut for ConditionalImportsFallback<'a> {
+  fn visit_mut_expr(&mut self, node: &mut Expr) {
+    match node {
+      Expr::Call(call) => match &call.callee {
+        swc_core::ecma::ast::Callee::Expr(expr) => match &**expr {
+          Expr::Ident(ident) if ident.sym.to_string().as_str() == "importCond" => {
+            match (match_str(&call.args[1].expr), match_str(&call.args[2].expr)) {
+              (Some((if_true, _if_true_span)), Some((if_false, _if_false_span))) => {
+                if !self.config.conditional_bundling {
+                  // Found importCond, if flag off replace an inline require import
+                  // importCond('CONDITION', 'IF_TRUE', 'IF_FALSE');
+                  // =>
+                  // (globalThis.__MOD_COND?.['CONDITION'] ? require('IF_TRUE') : require('IF_FALSE')).default;
+                  *node = Expr::Member(MemberExpr {
+                    obj: ParenExpr {
+                      expr: CondExpr {
+                        test: OptChainExpr {
+                          base: Box::new(
+                            MemberExpr {
+                              obj: Box::new(
+                                MemberExpr {
+                                  obj: Box::new(Expr::Ident("globalThis".into())),
+                                  prop: MemberProp::Ident("__MOD_COND".into()),
+                                  span: DUMMY_SP,
+                                }
+                                .into(),
+                              ),
+                              prop: MemberProp::Computed(ComputedPropName {
+                                expr: call.args[0].expr.clone(),
+                                span: DUMMY_SP,
+                              }),
+                              span: DUMMY_SP,
+                            }
+                            .into(),
+                          ),
+                          optional: true,
+                          span: DUMMY_SP,
+                        }
+                        .into(),
+                        cons: Box::new(create_require(if_true, self.unresolved_mark).into()),
+                        alt: Box::new(create_require(if_false, self.unresolved_mark).into()),
+                        span: DUMMY_SP,
+                      }
+                      .into(),
+                      span: DUMMY_SP,
+                    }
+                    .into(),
+                    prop: MemberProp::Ident("default".into()),
+                    span: DUMMY_SP,
+                  })
+                }
+              }
+              _ => {}
+            };
+          }
+          _ => {}
+        },
+        _ => {}
+      },
+      _ => {}
+    }
+  }
+}
+
+#[cfg(test)]
+mod tests {
+  use super::*;
+  use crate::test_utils::{remove_code_whitespace, run_test_visit, RunContext, RunVisitResult};
+
+  fn make_conditional_imports<'a>(
+    context: RunContext,
+    config: &'a Config,
+  ) -> ConditionalImportsFallback<'a> {
+    ConditionalImportsFallback {
+      config,
+      unresolved_mark: context.unresolved_mark,
+    }
+  }
+
+  fn make_config() -> Config {
+    let mut config = Config::default();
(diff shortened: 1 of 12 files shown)


packages/transformers/js/core/src/conditional_imports_fallback.rs
```diff
@@ -0,0 +1,168 @@
...
+        Callee::Expr(expr) => match &**expr {
+          Expr::Ident(ident) if ident.sym.to_string().as_str() == "importCond" => {
+            match (match_str(&call.args[1].expr), match_str(&call.args[2].expr)) {
+              (Some((if_true, if_true_span)), Some((if_false, if_false_span))) => {
+                if !self.config.conditional_bundling {
+                  // Found importCond, if flag off replace an inline require import
+                  // importCond('CONDITION', 'IF_TRUE', 'IF_FALSE');
+                  // =>
+                  // (globalThis.__MOD_COND && globalThis.__MOD_COND['CONDITION'] ? require('IF_TRUE') : require('IF_FALSE')).default;
+                  *node = Expr::Member(MemberExpr {
```
yamadapc, line 29: Have you tried building this with the macro? Sometimes even if you build an AST with placeholders and just replace variables that's nicer

packages/transformers/js/core/src/conditional_imports_fallback.rs
```diff
@@ -0,0 +1,168 @@
...
+                    obj: ParenExpr {
+                      expr: CondExpr {
+                        test: Box::new(
+                          BinExpr {
+                            op: BinaryOp::LogicalAnd,
+                            left: MemberExpr {
+                              obj: Box::new(Expr::Ident("globalThis".into())),
+                              prop: MemberProp::Ident("__MOD_COND".into()),
+                              span: call.span(),
```
yamadapc, line 38: I don't think you need to set these

packages/transformers/js/core/src/conditional_imports_fallback.rs
```diff
@@ -0,0 +1,168 @@
...
+use crate::Config;
+
+pub struct ConditionalImportsFallback<'a> {
+  pub config: &'a Config,
+  pub unresolved_mark: Mark,
+}
+
+impl<'a> VisitMut for ConditionalImportsFallback<'a> {
+  fn visit_mut_expr(&mut self, node: &mut Expr) {
+    match node {
```
yamadapc, line 18: It's nicer to write this code with conditional patterns `let ... = ... else { return; }`

yamadapc (approved):

packages/transformers/js/core/src/conditional_imports_fallback.rs
```diff
@@ -0,0 +1,168 @@
...
+      Expr::Call(call) => match &call.callee {
+        Callee::Expr(expr) => match &**expr {
+          Expr::Ident(ident) if ident.sym.to_string().as_str() == "importCond" => {
```
yamadapc, line 21: Compare to `atom!("importCond")` directly instead

packages/transformers/js/core/src/conditional_imports_fallback.rs
```diff
@@ -0,0 +1,168 @@
...
+            match (match_str(&call.args[1].expr), match_str(&call.args[2].expr)) {
```
yamadapc, line 22: what is this, why not `as_str`

packages/transformers/js/core/src/conditional_imports_fallback.rs
```diff
@@ -0,0 +1,168 @@
...
+                              obj: Box::new(
+                                MemberExpr {
+                                  obj: Box::new(Expr::Ident("globalThis".into())),
+                                  prop: MemberProp::Ident("__MOD_COND".into()),
+                                  span: call.span(),
+                                }
+                                .into(),
+                              ),
+                              prop: MemberProp::Computed(ComputedPropName {
+                                expr: call.args[0].expr.clone(),
```
yamadapc, line 51: We need to check that the arg is there otherwise the bundler will panic on `require()`

packages/transformers/js/core/src/conditional_imports_fallback.rs
yamadapc: Please use the macro to build this even if you just build part of the AST

packages/transformers/js/core/src/conditional_imports_fallback.rs
JakeLane: `as_str` will lose the span context, which is required lower down. I've refactored a bit which hopefully makes the intention clearer.

packages/transformers/js/core/src/conditional_imports_fallback.rs
JakeLane: Yep I verifed that you are correct, the helper behaviour only checks the `require` string specifier

packages/transformers/js/core/src/conditional_imports_fallback.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    let Callee::Expr(callee_expr) = &call.callee else {
+      return;
+    };
+
+    let Expr::Ident(callee_ident) = &**callee_expr else {
+      return;
+    };
+
+    if callee_ident.sym != atom!("importCond") && call.args.len() != 3 {
```
yamadapc, line 34: I forgot that you should be checking the unresolved mark, so that you aren't matching on shadowed `importCond`

yamadapc: https://github.com/atlassian-labs/atlaspack/blob/main/crates/atlaspack_plugin_optimizer_inline_requires/src/lib.rs#L118-L120

packages/transformers/js/core/src/conditional_imports_fallback.rs
```diff
@@ -0,0 +1,113 @@
...
+      // Not an importCond
+      return;
+    }
+
+    if callee_ident.span.ctxt.outer() != self.unresolved_mark {
+      // Don't process importCond more than once
```
yamadapc, line 33: Forgot to exit? The idea of this block is to handle this:
```
importCond(...) // <- handle this
function test(importCond) {
   importCond(...); // <- ignore this
}
```

packages/transformers/js/core/src/conditional_imports_fallback.rs
```diff
@@ -0,0 +1,113 @@
...
+        span: DUMMY_SP,
+        type_args: None,
+      }
+      .into()
+    };
+
+    // importCond('CONDITION', 'IF_TRUE', 'IF_FALSE');
+    // =>
+    // (globalThis.__MCOND && globalThis.__MCOND['CONDITION'] ? require('IF_TRUE') : require('IF_FALSE')).default;
+    let new_node = quote!(
```
yamadapc, line 69: Is this what you want when flag is off?

packages/transformers/js/core/src/dependency_collector.rs
```diff
@@ -2190,46 +2171,4 @@ document.body.appendChild(img);
...
-    "#;
-
-    let RunVisitResult { output_code, .. } = run_test_fold(input_code, |context| {
-      make_dependency_collector(context, &mut items, &mut diagnostics, &config)
-    });
-
-    let hash = make_placeholder_hash("b", DependencyKind::Require);
-    let expected_code = format!(
-      r#"
-      const x = require("{}");
```
yamadapc, line 2211: placeholder no longer looks like this

packages/transformers/js/core/src/conditional_imports_fallback.rs
JakeLane: yep, when the flag is on we want to enable the changes in the bundling algorithm in a future PR. When flag is off, we want sync imports (like a require) so the condition function works without the bundle changes (in production code).

PR #75, opened 2024-09-16
Sort the connections back into topological order after traversing runtimes

benjervis:
A while back in Parcel, we [reversed the order](https://github.com/parcel-bundler/parcel/pull/8837) that runtimes were processed in, so that child runtimes could be processed before parent bundles who would need to know about them. That didn't quite cut it, so we moved to [reverse topological order](https://github.com/parcel-bundler/parcel/pull/8981).

This works fine for Jira, but in Confluence it causes an issue where a bundle gets loaded thinking that some of its dependencies are already loaded when they aren't.

We thought we'd fixed this with `connections.reverse()`, but that doesn't _actually_ restore topological order, and it seems that Confluence has found the edge case.

The fix here is to actually restore topological order, by processing the bundles in reverse, and then re-traversing and building up the connection list that way.

### DIFF
--- a/packages/core/core/src/applyRuntimes.js
+++ b/packages/core/core/src/applyRuntimes.js
@@ -32,6 +32,7 @@ import createAssetGraphRequest from './requests/AssetGraphRequest';
 import {createDevDependency, runDevDepRequest} from './requests/DevDepRequest';
 import {toProjectPath, fromProjectPathRelative} from './projectPath';
 import {tracer, PluginTracer} from '@atlaspack/profiler';
+import {DefaultMap} from '@atlaspack/utils';
 
 type RuntimeConnection = {|
   bundle: InternalBundle,
@@ -82,7 +83,6 @@ export default async function applyRuntimes<TResult: RequestResult>({
   configs: Map<string, Config>,
 |}): Promise<Map<string, Asset>> {
   let runtimes = await config.getRuntimes();
-  let connections: Array<RuntimeConnection> = [];
 
   // As manifest bundles may be added during runtimes we process them in reverse topological
   // sort order. This allows bundles to be added to their bundle groups before they are referenced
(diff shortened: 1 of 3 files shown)


mattcompiles (approved):

packages/core/core/src/applyRuntimes.js
```diff
@@ -82,18 +112,17 @@ export default async function applyRuntimes<TResult: RequestResult>({
   configs: Map<string, Config>,
 |}): Promise<Map<string, Asset>> {
   let runtimes = await config.getRuntimes();
-  let connections: Array<RuntimeConnection> = [];
 
-  // As manifest bundles may be added during runtimes we process them in reverse topological
-  // sort order. This allows bundles to be added to their bundle groups before they are referenced
-  // by other bundle groups by loader runtimes
+  // Sort bundles into reverse topological order
```
yamadapc, line 116: This is topological order, not reverse

packages/core/core/src/applyRuntimes.js
```diff
@@ -192,8 +221,14 @@ export default async function applyRuntimes<TResult: RequestResult>({
     }
   }
 
-  // Correct connection order after generating runtimes in reverse order
-  connections.reverse();
+  // Sort the bundles back into forward topological order
+  let connections: Array<RuntimeConnection> = [];
+
+  bundleGraph.traverseBundles({
```
yamadapc, line 227: This isn't reverse topological order, I don't think

packages/core/core/src/applyRuntimes.js
```diff
@@ -192,8 +221,14 @@ export default async function applyRuntimes<TResult: RequestResult>({
     }
   }
 
-  // Correct connection order after generating runtimes in reverse order
-  connections.reverse();
+  // Sort the bundles back into forward topological order
```
yamadapc, line 224: You're now going for reverse topological order

packages/core/core/src/applyRuntimes.js
yamadapc: This should be
```
const connections = [];
bundleGraph.traverseBundles({
    exit(bundle) {
        connections.push(...connectionMap.get(bundle));
    }
});
connections.reverse();
```

PR #76, opened 2024-09-17
Limit number of rust threads on tests to 2

yamadapc:
This improves runtime of running integration tests at the moment.

A better fix is to pre-compile these tests.


yamadapc:
There's significant improvement the more cores you have.

mattcompiles (approved):

PR #77, opened 2024-09-17
Start work on Rust HTML transformer

yamadapc:
There are 2 changes:

* Add rust HTML transformer and start to write test-suite for JavaScript HTML transformer
* Fix asset ID alias

HTML transformer implementation is not complete. I chose to use Servo's HTML parser. There's just a very barebones set-up at the moment.

Please see the commit messages for justifications. I'm removing the "non zero u32" alias because it's unused and incompatible with the current code. So I'd rather alias to string until we port all code to use u32.

We shouldn't have an alias that can't be used on the current code. A change to make the IDs u32 should be applied globally to everything.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -416,6 +416,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "atlaspack_plugin_transformer_html"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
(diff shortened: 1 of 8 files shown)


JakeLane (approved):

yamadapc:
Sorry I shouldn't have set auto-merge ;

PR #78, opened 2024-09-17
Always run all V2 tests

yamadapc:
This commit modifies the integration testing rig so that we run all V2 tests on the V3 integration suite, even if they are currently marked as failures.

The purpose of doing this is to:

* Quickly understand what failures are happening from CI build logs
* Be able to build tooling that automatically tries to diagnose current gaps
* Identify tests that are not failing but are currently skipped

This diff adds a few other changes outside of test-utils:

* The rust 'RPC' resolver code, which is meant for plugin support is changed so that it will not panic and crash a thread, since that will otherwise stall the test-suite
* A few tests that should be passing on V3 are re-enabled
* I reckon we should force the asset graph to be built always on the V3 implementation for now. So this adds `force: true` on the runs that are using the rust back-end. This is temporary until we have cache invalidation


mattcompiles (approved):
Looks good. Haven't checked why CI is failing though.

PR #85, opened 2024-09-19
Implement initial pass of conditional bundles

JakeLane:
## Motivation

We want to be able to allow devs to introduce conditional changes (feature gating) without affecting bundle size. We need a new API that allows devs to synchronously import without introducing both branches (true and false cases) into the same bundle.

## Changes

- Using `importCond` with a feature flag now creates a new bundle
- Introduce a new API for the bundle graph, `getConditionalBundleMapping` so we can know which bundles need which scripts, based on the condition
- Introduce a new reporter which creates a manifest via the `getConditionalBundleMapping` function, which we need to make script loading decisions on the server
- When used in a sync bundle, we expect the module to be loaded by the server
- When used in an async bundle, we will load the conditional bundles associated


tomgasson:
Concept LGTM, I'll leave bundling details up to someone else to review

tomgasson (approved):

crates/atlaspack_core/src/types/dependency.rs
```diff
@@ -221,6 +221,8 @@ pub enum Priority {
   Parallel = 1,
   /// The dependency should be placed in a separate bundle that is loaded later
   Lazy = 2,
+  /// The dependency should be placed in a separate bundle that is loaded conditionally
+  Conditional = 3,
```
marcins, line 225: Are the priority numbers used anywhere as a comparison, or is this an "unordered" value?

marcins (approved):
There's a lot of change here and I'm probably not deep enough in it anymore to find any obvious mistakes, I can definitely see that it's heavily based on the code I created late last year like you said. Does this still produce one extra bundle per condition arm? If so, is there a planned approach for merging / minimising these bundles if we end up with a lot of conditions?

crates/atlaspack_core/src/types/dependency.rs
JakeLane: I believe the numbers are used to reduce the size in memory

JakeLane:
> There's a lot of change here and I'm probably not deep enough in it anymore to find any obvious mistakes, I can definitely see that it's heavily based on the code I created late last year like you said. Does this still produce one extra bundle per condition arm? If so, is there a planned approach for merging / minimising these bundles if we end up with a lot of conditions?

Yes this creates a bundle split per conditional import. I have plans to optimise this in a future pass based on how it works out in production. We do get some deduplication of dependencies by the bundling algorithm already (lazy modules can get internalised into other bundles), so I didn't want to overcomplicate the work.

PR #87, opened 2024-09-19
Change transformers to allow multiple assets in result

Nikola-3:
The PR achieves the following:
* implements logic to discover and process new assets with a queue
* adds unit tests for new functionality
* changes the transformers and pipeline APIs to allow for easier mocking and to no longer use hash keyword to avoid colliding with the standard library

The remaining work is to model the `AssetRequestOutput` and related function to update the new assets.

### DIFF
--- a/crates/atlaspack/src/requests/asset_request.rs
+++ b/crates/atlaspack/src/requests/asset_request.rs
@@ -1,7 +1,3 @@
-use std::hash::Hash;
-use std::path::PathBuf;
-use std::sync::Arc;
-
 use atlaspack_core::plugin::AssetBuildEvent;
 use atlaspack_core::plugin::BuildProgressEvent;
 use atlaspack_core::plugin::ReporterEvent;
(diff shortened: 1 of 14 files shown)


yamadapc (approved):

crates/atlaspack/src/requests/asset_request.rs
```diff
@@ -86,40 +90,178 @@ impl Request for AssetRequest {
...
+      Some(vec![Asset::default()]),
+      Some(vec![Dependency::default()]),
+      Some(vec![PathBuf::from("./tmp")]),
+    );
+    let pipeline = TransformerPipeline::new(vec![mock_transformer]);
+    let asset = Asset::default();
+    let plugins = Arc::new(MockPlugins::new());
+    let expected_discovered_assets = vec![Asset::default()];
+    let expected_dependencies = vec![Dependency::default()];
+    let expected_invalidations = vec![PathBuf::from("./tmp")];
```
yamadapc, line 183: it'd be nice to test cases where invalidations / dependencies have multiple entries from the newly discovered assets

Nikola-3: I can do that in a separate PR 👍

PR #89, opened 2024-09-23
Add basic Node JS plugin support

mattcompiles:
## Motivation

This PR adds basic Node JS plugin support to Atlaspack V3. I've stripped back the code to not bother handle dependencies at all. This requires a bunch more changes which I have done in [another branch](https://github.com/atlassian-labs/atlaspack/tree/node-plugin-html) but I'm not sure if we really want to continue supporting dependencies in JS plugins going forward. If we decide we'd like to add this feature we can follow up with that. 

For now this PR is enough to get a basic transformer like `@atlaspack/transformer-svg` working which I have enabled an integration test for in this PR. There's still a bunch more things to add like invalidations and source-maps but this is a good start. 

## Checklist

- [] Existing or new tests cover this change

### DIFF
--- a/crates/atlaspack_plugin_rpc/src/plugin/transformer.rs
+++ b/crates/atlaspack_plugin_rpc/src/plugin/transformer.rs
@@ -75,7 +75,7 @@ impl TransformerPlugin for RpcTransformerPlugin {
       side_effects: result.asset.side_effects,
       is_bundle_splittable: result.asset.is_bundle_splittable,
       is_source: result.asset.is_source,
-      // TODO: Fix or remove the duplicate meta fields. I vote we remove these fields as they're just duplicates of meta
+      // TODO: Fix or remove the duplicate meta fields.
       ..Default::default()
     };
 
--- a/crates/atlaspack_plugin_rpc/src/plugin/transformer.rs
+++ b/crates/atlaspack_plugin_rpc/src/plugin/transformer.rs
@@ -1,5 +1,6 @@
 use std::fmt;
 use std::fmt::Debug;
+use std::sync::Arc;
 
 use anyhow::Error;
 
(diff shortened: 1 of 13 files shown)


packages/core/core/src/atlaspack-v3/worker/worker.js
```diff
@@ -1,26 +1,87 @@
...
+
+    try {
+      if (transformer.parse) {
+        // $FlowFixMe
+        let ast = await transformer.parse({asset: assetCompat});
+        // $FlowFixMe
+        assetCompat.setAST(ast);
+      }
+
+      // $FlowFixMe
```
yamadapc, line 45: What's wrong?

yamadapc (approved):

crates/atlaspack_plugin_rpc/src/plugin/transformer.rs
```diff
@@ -9,40 +10,81 @@ use atlaspack_core::plugin::TransformResult;
...
-    Ok(TransformResult {
+    let asset_env = asset.env.clone();
+    let stats = asset.stats.clone();
+    let run_transformer_opts = RpcTransformerOpts {
+      // TODO: Make this not bad
+      resolve_from: self.plugin.resolve_from.to_str().unwrap().to_string(),
+      specifier: self.plugin.package_name.clone(),
+      // TODO: Pass this just once to each worker
+      options: self.plugin_options.clone(),
       asset,
```
yamadapc, line 57: The nicest thing would be to make this be a non serialized object. 

So we would create an object from Rust with getter/setter fields that modify or access that directly on demand.

mattcompiles: Interesting idea. I'll look into that.

packages/core/core/src/atlaspack-v3/worker/worker.js
mattcompiles: Transformers currently expect a larger API which I'm not yet providing.

PR #90, opened 2024-09-23
HTML transformer improvements

yamadapc:
* Adds native compare environment variable that will diff asset graphs
* Fixes many discrepancies between the graphs
* Starts to add HMR setup to rust HTML transformer
* Add unit-tests to HTML transformer
* Make all IDs be generated from the same function, shared between rust and JS, so that we get consistent IDs between implementations
* Add 'transformation context' type to pass environment into the transformer
* Adopts caniuse database library added in previous PR


crates/atlaspack/src/requests/asset_graph_request.rs
```diff
@@ -167,6 +167,7 @@ impl AssetGraphBuilder {
           code: code.clone(),
           env: dependency.env.clone(),
           file_path: path,
+          project_root: self.request_context.project_root.clone(),
```
yamadapc, line 170: We weren't setting the project root for the asset graph, so paths weren't set absolute/relative properly

crates/atlaspack/src/requests/path_request.rs
```diff
@@ -202,10 +203,6 @@ mod tests {
     }
   }
 
-  impl Hash for ResolvedResolverPlugin {
-    fn hash<H: std::hash::Hasher>(&self, _state: &mut H) {}
```
yamadapc, line 206: We should avoid implementing `Hash` manually, if we do, we need to implement a valid hash function.

crates/atlaspack/src/requests/target_request.rs
```diff
@@ -43,22 +45,14 @@ mod package_json;
...
+#[derive(Debug, Hash)]
 pub struct TargetRequest {
   pub default_target_options: DefaultTargetOptions,
   pub entry: Entry,
-  pub env: Option<HashMap<String, String>>,
+  pub env: Option<BTreeMap<String, String>>,
   pub mode: BuildMode,
 }
 
-impl Hash for TargetRequest {
```
yamadapc, line 54: This hash implementation missed `env` out.

crates/atlaspack/src/requests/target_request.rs
```diff
@@ -43,22 +45,14 @@ mod package_json;
 ///
 /// Targets will be generated from the project package.json file and input Atlaspack options.
 ///
-#[derive(Debug)]
+#[derive(Debug, Hash)]
 pub struct TargetRequest {
   pub default_target_options: DefaultTargetOptions,
   pub entry: Entry,
-  pub env: Option<HashMap<String, String>>,
+  pub env: Option<BTreeMap<String, String>>,
```
yamadapc, line 52: We use `BTreeMap` since that will have a deterministic hash, whereas `HashMap` does not since keys aren't sorted in any way.

crates/atlaspack/src/requests/target_request.rs
```diff
@@ -410,6 +404,23 @@ impl TargetRequest {
     Ok(targets)
   }
 
+  fn get_default_engines_for_context(&self, context: EnvironmentContext) -> Engines {
```
yamadapc, line 407: When we set default engines, we should only set the default engines for the current target.

So if we're targeting node, we'll default to `{ node: 16 }` and if we target browsers we'll set the corresponding browsers default.

crates/atlaspack_core/src/plugin/transformer_plugin.rs
```diff
@@ -27,6 +28,29 @@ pub struct TransformResult {
   pub invalidate_on_file_change: Vec<PathBuf>,
 }
 
+#[derive(Clone)]
+pub struct TransformContext {
```
yamadapc, line 32: Added new `TransformContext` type.

All the values currently passed into the transformer constructor should be moved here. This will avoid future changes to the `Transform` trait signature.

crates/atlaspack_core/src/types/asset.rs
```diff
@@ -55,19 +56,42 @@ impl From<String> for Code {
...
+  pub pipeline: Option<&'a str>,
+  pub file_type: FileType,
+  pub query: Option<&'a str>,
+  /// This should be set to None if it's equal to the asset-id and set by the
+  /// constructor otherwise the values will differ. See [`Asset::new`] for more.
+  pub unique_key: Option<&'a str>,
+}
 
-  env.hash(&mut hasher);
+pub fn create_asset_id(params: CreateAssetIdParams) -> String {
```
yamadapc, line 75: Asset IDs, environment IDs and dependency IDs are moved to be generated from rust functions.

This means that:

* We get stable IDs for everything between v2/v3
* A few issues related to IDs are possibly fixed, due to stricter checks

crates/atlaspack_core/src/types/bundle.rs
```diff
@@ -73,20 +73,19 @@ pub struct Bundle {
...
-  Isolated = 1,
-
-  /// Unspecified bundling behavior
-  None = 255,
+  Isolated,
 }
 
 impl Default for BundleBehavior {
   fn default() -> Self {
-    BundleBehavior::None
```
yamadapc, line 90: We don't overload optional types into the enum, this way we can parse and generate corresponding JSON more easily.

crates/atlaspack_core/src/types/environment.rs
```diff
@@ -172,46 +186,72 @@ impl From<EnvironmentContext> for IncludeNodeModules {
   }
 }
 
-impl Hash for IncludeNodeModules {
-  fn hash<H: Hasher>(&self, state: &mut H) {
-    match self {
-      IncludeNodeModules::Bool(b) => b.hash(state),
-      IncludeNodeModules::Array(a) => a.hash(state),
-      IncludeNodeModules::Map(m) => {
-        for (k, v) in m {
```
yamadapc, line 181: This hash was incorrect and would generate different values for the same map.

Since there is non deterministic order of the entries, this outputs a separate hash for maps that are equal.

crates/atlaspack_core/src/types/environment/browsers.rs
```diff
@@ -1,188 +1,36 @@
...
+use std::str::FromStr;
 
 use browserslist::Distrib;
 
 use super::version::Version;
 
 /// List of targeted browsers
-#[derive(Clone, Default, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Default, Debug, Eq, PartialEq)]
 pub struct Browsers {
```
yamadapc, line 10: This is migrated to use https://github.com/atlassian-labs/atlaspack/pull/88

crates/atlaspack_plugin_transformer_js/src/js_transformer/conversion/loc.rs
```diff
@@ -1,13 +1,17 @@
...
 use atlaspack_core::types::{Location, SourceLocation};
 
 pub(crate) fn convert_loc(
+  project_root: &Path,
   file_path: PathBuf,
   loc: &atlaspack_js_swc_core::SourceLocation,
 ) -> SourceLocation {
   SourceLocation {
-    file_path,
+    file_path: file_path
```
yamadapc, line 11: Fix inconsistency between JS transformer symbols and the node implementation

packages/core/core/src/AssetGraph.js
```diff
@@ -227,7 +227,7 @@ export default class AssetGraph extends ContentGraph<AssetGraphNode> {
         // The passed project path is ignored in this case, because there is no `loc`
         createDependency('', {
           specifier: fromProjectPathRelative(entry.filePath),
-          specifierType: 'esm', // ???
+          specifierType: 'url',
```
yamadapc, line 230: This was an arbitrary choice. It doesn't matter what specifier type we set for the entries, but in Rust land we've set this to URL.

So I've changed this to URL everywhere.

packages/core/core/src/Dependency.js
```diff
@@ -45,23 +46,60 @@ type DependencyOpts = {|
...
-    hashString(
-      (opts.sourceAssetId ?? '') +
-        opts.specifier +
-        opts.env.id +
-        (opts.target ? JSON.stringify(opts.target) : '') +
-        (opts.pipeline ?? '') +
-        opts.specifierType +
-        (opts.bundleBehavior ?? '') +
-        (opts.priority ?? 'sync') +
-        (opts.packageConditions ? JSON.stringify(opts.packageConditions) : ''),
```
yamadapc, line 63: Similar to the hashmap ordering, these IDs are not deterministic due to relying on insertion order of these maps.

This likely lead to build bugs, because the IDs must be consistent for equivalent dependencies for the bundles output to be correct.

packages/core/core/src/requests/BundleGraphRequest.js
```diff
@@ -118,6 +118,24 @@ export default function createBundleGraphRequest(
         },
       );
 
+      // if (input.rustAtlaspack && process.env.NATIVE_COMPARE === 'true') {
```
yamadapc, line 121: I've added diffing code but commented out to simplify getting the PR builds green.

The main purpose of the PR is to minimize this diff

packages/core/core/test/Environment.test.js
```diff
@@ -1,12 +1,15 @@
...
+// $FlowFixMe
+import expect from 'expect';
 import {createEnvironment} from '../src/Environment';
+import {initializeMonitoring} from '../../rust';
 
 describe('Environment', () => {
   it('assigns a default environment with nothing passed', () => {
     assert.deepEqual(createEnvironment(), {
-      id: 'c242f987e3544367',
+      id: 'd821e85f6b50315e',
```
yamadapc, line 12: All IDs have changed, since the hash function now gets different inputs.

packages/core/integration-tests/test/bundler.js
```diff
@@ -1806,16 +1806,16 @@ describe.v2('bundler', function () {
           assets: ['index.html'],
         },
         {
```
yamadapc, line 1808: Because the hash function changed, these bundles have changed.

crates/atlaspack_core/src/plugin.rs
```diff
@@ -49,8 +41,15 @@ pub struct PluginLogger {}
...
+  /// Environment variables
+  pub env: Option<BTreeMap<String, String>>,
   pub log_level: LogLevel,
   pub mode: BuildMode,
   pub project_root: PathBuf,
 }
+
+impl PluginOptions {
+  pub fn should_scope_hoist(&self) -> bool {
+    self.mode == BuildMode::Production
```
mattcompiles, line 53: This should be controlled by the target options passed in from JS. You can def have a prod build that doesn't have scope hoisting enabled.

crates/atlaspack_core/src/types/bundle.rs
```diff
@@ -73,20 +73,19 @@ pub struct Bundle {
...
+    BundleBehavior::Inline
```
mattcompiles, line 87: Can we avoid implementing Default for this now? If it's optional now we shouldn't really need it. It also feels like a footgun because you may accidentally create an Bundle that's inline by defaulting.

packages/core/core/src/CommittedAsset.js
```diff
@@ -17,7 +17,7 @@ export default class CommittedAsset {
   mapBuffer: ?Promise<?Buffer>;
   map: ?Promise<?SourceMap>;
   ast: ?Promise<AST>;
-  idBase: ?string;
+  code: ?string;
```
mattcompiles, line 20: Any context here?

packages/core/core/test/Environment.test.js
mattcompiles: RIP Jira cache 😉

packages/core/core/src/CommittedAsset.js
yamadapc: When there are assets that have only in-memory code (I think assets generated by others), their code is hashed to generate the ID.

Previously the code and path were hashed initially and passed as `idBase`, which then was a part of the overall ID. I've changed this to just pass the values around and use in ID generation, but we pass around that value now instead of `idBase`.

mattcompiles (approved):
Nice job 👍

crates/atlaspack_core/src/plugin/transformer_plugin.rs
MonicaOlejniczak: I disagree, but maybe it will be more apparent later. The main motivation for the constructor is to perform operations that can be cached with obvious one-off inputs. If you place everything into the transform fn, that line is blurred since it gets called for every asset. Having the distinction is useful and creates more performance oriented opportunities imo.

yamadapc: If there is benefit, there's no problem even in passing the values to both methods, this way implementations that don't cache _thing that would be cached_ can just be written as a plain function of input -> output. These are all the implementations at the moment.

If there are implementations that pre-process something to share on the instance, it's okay to just pass the input again.

If we can write things as plain functions with less boilerplate, that's better.

If we have an actual need to do work on a constructor, we can obviously keep/add that.

PR #92, opened 2024-09-26
Fixing serde errors for optional napi properties

alshdavid:
There are some issues with serde and enums.

This isn't the best solution but unblocks the next step of building JFE.

We should probably separate the input types coming from napi with dedicated structures and cast them to internal types to avoid having to propagate napi issues throughout the core.

### DIFF
--- a/crates/atlaspack_core/src/types/atlaspack_options.rs
+++ b/crates/atlaspack_core/src/types/atlaspack_options.rs
@@ -30,7 +30,7 @@ pub struct AtlaspackOptions {
   pub fallback_config: Option<String>,
 
   #[serde(default)]
-  pub log_level: Option<LogLevel>,
+  pub log_level: LogLevel,
 
   #[serde(default)]
   pub mode: BuildMode,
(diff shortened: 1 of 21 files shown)


crates/atlaspack_core/src/types/atlaspack_options.rs
```diff
@@ -74,7 +74,7 @@ impl<'de> Deserialize<'de> for BuildMode {
 #[serde(default, rename_all = "camelCase")]
 pub struct DefaultTargetOptions {
   pub dist_dir: Option<PathBuf>,
-  pub engines: Engines,
+  pub engines: Option<Engines>,
```
yamadapc, line 77: is this actually null? on my branch I've gotten this part to the point that it's exactly equal as previous behaviour

alshdavid: I'm running a JFE build and it is passing in `undefined` there. This change allows the build to continue

yamadapc (approved):

PR #93, opened 2024-09-27
V3: Using nodefs by default for now

alshdavid:
The native filesystem is missing functionality required to resolve `atlaspackrc#extends`. I have set it to use the Nodejs filesystem implementation to move to the next step and discuss this later.

We can't ship using the Nodejs fs from Rust due to performance issues so this is an essential "todo".

I think the best way to proceed is to implement the Flow `FileSystem` interface in Rust and make it available to Nodejs via napi bindings then use that in both Nodejs and Rust. I think that a symmetrical interface will make it easier going forward.

I'll leave that up for discussion next week.


alshdavid:
No longer required

PR #94, opened 2024-09-27
Don't clean-up orphaned nodes on every event

yamadapc:
This should improve performance of handling filesystem events, since there won't be one graph traversal for every event. Since complexity of this particular section is reduced to be roughly linear we expect worst case performance of handling events to improve.


Nikola-3 (approved):

yamadapc:
There's a mistake in this diff because the request graph doesn't always have a root. 

Also a lot of time is spent in the actual for loop body still.

PR #95, opened 2024-09-27
V3: Resolver plugins

alshdavid:
Basic support for resolver plugins

### DIFF
--- a/crates/atlaspack_plugin_rpc/Cargo.toml
+++ b/crates/atlaspack_plugin_rpc/Cargo.toml
@@ -17,4 +17,3 @@ napi = { version = "2.16.4", features = ["serde"], optional = true }
 serde = { version = "1.0.198", optional = true }
 parking_lot = { version = "0.12", optional = true }
 once_cell = { version = "1.19.0", optional = true }
-dyn-hash = "0.*"
--- a/crates/atlaspack_plugin_rpc/Cargo.toml
+++ b/crates/atlaspack_plugin_rpc/Cargo.toml
@@ -17,3 +17,4 @@ napi = { version = "2.16.4", features = ["serde"], optional = true }
 serde = { version = "1.0.198", optional = true }
 parking_lot = { version = "0.12", optional = true }
 once_cell = { version = "1.19.0", optional = true }
+dyn-hash = "0.*"
--- a/crates/atlaspack_plugin_rpc/Cargo.toml
+++ b/crates/atlaspack_plugin_rpc/Cargo.toml
@@ -17,3 +17,4 @@ napi = { version = "2.16.4", features = ["serde"], optional = true }
 serde = { version = "1.0.198", optional = true }
 parking_lot = { version = "0.12", optional = true }
 once_cell = { version = "1.19.0", optional = true }
+tracing = "0.1.40"
--- a/crates/atlaspack_plugin_rpc/src/nodejs/rpc_conn_nodejs.rs
+++ b/crates/atlaspack_plugin_rpc/src/nodejs/rpc_conn_nodejs.rs
@@ -1,4 +1,4 @@
-use atlaspack_core::plugin::Resolution;
+use atlaspack_core::plugin::Resolved;
 use atlaspack_napi_helpers::anyhow_from_napi;
 use atlaspack_napi_helpers::js_callable::JsCallable;
 use napi::{JsObject, JsUnknown};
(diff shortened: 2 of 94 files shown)


crates/atlaspack_plugin_rpc/src/nodejs/rpc_conns_nodejs.rs
```diff
@@ -19,15 +21,17 @@ impl NodejsWorkerFarm {
...
-      *current_index = *current_index + 1;
-    }
-    current_index.clone()
+    self
+      .current_index
+      .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |value| {
+        Some((value + 1) % self.workers.len())
+      })
+      .expect("Could not get worker")
+  }
```
alshdavid, line 31: :racehorse:

crates/atlaspack_plugin_rpc/src/plugin/resolver.rs
```diff
@@ -1,14 +1,33 @@
...
+use crate::RunResolverResolve;
+
+pub struct RpcResolverPlugin {
+  resolve_from: PathBuf,
+  specifier: String,
+  started: bool,
+  rpc_worker: RpcWorkerRef,
+}
+
+impl std::hash::Hash for RpcResolverPlugin {
```
yamadapc, line 24: prefer to derive

packages/core/core/src/atlaspack-v3/worker/worker.js
```diff
@@ -1,16 +1,97 @@
...
+        invalidations: [],
+        resolution: {type: 'unresolved'},
+      };
+    }
+
+    return {
+      invalidations: [],
+      resolution: {
+        type: 'resolved',
+        filePath: '',
```
yamadapc, line 83: I'm confused why this is empty

crates/atlaspack_plugin_rpc/src/plugin/resolver.rs
alshdavid: I have to exclude RPC worker from the hash calculation. I can move this to the RPC worker and ensure the trait implements Hash.

packages/core/core/src/atlaspack-v3/worker/worker.js
alshdavid: Missed that, thank you 🙏

crates/atlaspack_plugin_rpc/src/plugin/resolver.rs
yamadapc: Implementing these traits manually will lead to wrong implementations really. There's no benefit from implementing it manually. 

In this case we actually want to use a manual id method rather than the hash for identification of the plugin, similar to what nikola did for transformers

yamadapc: What I mean by no benefit is that if we need a cache key that has a subset of fields, we can just introduce a new type. 

There's no point in implementing hash with slightly different semantics than std library as an alternative to just splitting key types

crates/atlaspack_plugin_rpc/Cargo.toml
```diff
@@ -17,3 +17,4 @@ napi = { version = "2.16.4", features = ["serde"], optional = true }
 serde = { version = "1.0.198", optional = true }
 parking_lot = { version = "0.12", optional = true }
 once_cell = { version = "1.19.0", optional = true }
+dyn-hash = "0.*"
```
yamadapc, line 20: let's set a bound

crates/atlaspack_plugin_rpc/src/nodejs/rpc_conn_nodejs.rs
```diff
@@ -1,20 +1,34 @@
...
 /// RpcConnectionNodejs wraps the communication with a
 /// single Nodejs worker thread
 pub struct NodejsWorker {
   ping_fn: JsCallable,
+  load_plugin_fn: JsCallable,
+  run_resolver_resolve_fn: JsCallable,
   transformer_register_fn: JsCallable,
 }
 
+impl std::hash::Hash for NodejsWorker {
```
yamadapc, line 19: can we not implement these ? just use a different construct ; don't use `.hash` (e.g.: add a method and implement that)

crates/atlaspack_plugin_rpc/src/plugin/resolver.rs
alshdavid: Updated

crates/atlaspack_plugin_rpc/Cargo.toml
alshdavid: Removed this and replaced it with the `id()` method

yamadapc (approved):

crates/atlaspack_plugin_rpc/src/plugin/resolver.rs
```diff
@@ -42,6 +46,15 @@ impl RpcResolverPlugin {
 }
 
 impl ResolverPlugin for RpcResolverPlugin {
+  fn id(&self) -> u64 {
```
yamadapc, line 49: LGTM

PR #96, opened 2024-09-29
Add VCS aware file change event querying

yamadapc:
This PR adds a prototype of VCS aware change lists queries.

It adds a new filesystem implementation which uses a rust implementation of git diffing / yarn lock/state reading to get the list of changed files.

There is a `README.md` explaining details including how to roll-out / next steps. As it is, this branch does not consume the event lists returned by VCS, as I'd like to measure the accuracy of the changes.

Additionally, an example binary is provided which can be used to list changes.


crates/atlaspack_vcs/src/yarn_integration/yarn_lock_parser.rs
yamadapc, line 1: We don't need to support this

crates/atlaspack_vcs/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+  exclude_patterns: &[String],
+) -> Result<Vec<VCSFile>, anyhow::Error> {
+  let mut command = Command::new("git");
+  command
+    .arg("ls-files")
+    // .arg("--format=%(objectname)%(path)")
+    .arg("--deleted")
+    .arg("--modified")
+    .arg("--others")
+    .arg("-z") // We separate rows by \0 to avoid issues with newlines
```
tomgasson, line 237: If you are able to distinguish CI builds (which won't have a working tree changes) then you can shortcut a bunch of time spent here.

yamadapc:
I'll remove the yarn v1 parser - https://gist.github.com/yamadapc/285bc0614027aadf25fd22154dae1538

mattcompiles (approved):

PR #97, opened 2024-09-30
Create babel-plugin-transform-conditional-imports for babel transform envs

JakeLane:
## Motivation

Some environments (like webpack and jest) need to transform `importCond` without errors. This change is similar to what's in feature flag off state for conditional bundling.

## Changes

New babel plugin for use in environments where atlaspack is not transforming code.

## Checklist


JakeLane:
I ended up implementing a swc plugin that can be used externally. Unfortunately, caused some issues when trying to slot into swc-loader and jest, so I've had to drop this. Given we reuse the code here we can just migrate over in the future and delete the babel plugin.

packages/utils/babel-plugin-transform-conditional-imports/test/babel-plugin-transform-conditional-imports.test.js
```diff
@@ -0,0 +1,47 @@
...
+    assert(
+      transformed ===
+        `const _CONDITION$IF_TRUE$IF_FALSE = {
+  ifTrue: require('IF_TRUE').default,
+  ifFalse: require('IF_FALSE').default
+};
+Object.defineProperty(_CONDITION$IF_TRUE$IF_FALSE, "load", {
+  get: () => globalThis.__MCOND && globalThis.__MCOND('CONDITION') ? _CONDITION$IF_TRUE$IF_FALSE.ifTrue : _CONDITION$IF_TRUE$IF_FALSE.ifFalse
+});
+_CONDITION$IF_TRUE$IF_FALSE.load;`,
```
tomgasson, line 44: I'm not sure I get this - it's immediately called?

crates/atlaspack_conditional_imports/src/lib.rs
```diff
@@ -0,0 +1,346 @@
...
+    &mut self,
+    cond: Atom,
+    if_true: Atom,
+    if_true_span: Span,
+    if_false: Atom,
+    if_false_span: Span,
+  ) -> Expr {
+    // importCond('CONDITION', 'IF_TRUE', 'IF_FALSE');
+    // =>
+    // (globalThis.__MCOND && globalThis.__MCOND['CONDITION'] ? require('IF_TRUE') : require('IF_FALSE')).default;
```
tomgasson, line 81: Can we just hard fail if `globalThis.__MCOND` isn't defined? I don't want this going into unknown places... better we actually set them up

packages/utils/babel-plugin-transform-conditional-imports/test/babel-plugin-transform-conditional-imports.test.js
JakeLane: `importCond` isn't assigned to a variable so it just becomes a lone identifier in this state of the pipeline

packages/transformers/js/core/src/lib.rs
```diff
@@ -359,7 +357,18 @@ pub fn transform(
 
               // Replace conditional imports with requires when flag is off
               if !config.conditional_bundling {
-                module.visit_mut_with(&mut ConditionalImportsFallback { unresolved_mark });
+                module.visit_mut_with(
```
tomgasson, line 360: This looks like it's the wrong way around.  !conditional_bundling, yet we're visiting it with the conditional imports visitor?

JakeLane: This visitor is used for the fallback without bundling behaviour. Perhaps I can make this clearer by renaming the visitor?

tomgasson (approved):

crates/atlaspack_swc_transform_utils/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "atlaspack_swc_transform_utils"
```
MonicaOlejniczak, line 2: Can this be named more specifically so we have more focused abstractions? Something like `atlaspack_swc_runner`

crates/atlaspack_swc_transform_utils/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2021"
+description = "Atlaspack Bundler"
```
MonicaOlejniczak, line 5: This should be filled out

package.json
```diff
@@ -10,11 +10,12 @@
   "workspaces": [
     "crates/caniuse_database",
     "crates/lmdb-js-lite",
+    "crates/atlaspack_contextual_imports_swc_plugin",
```
MonicaOlejniczak, line 13: /nit make it alphabetical

packages/utils/babel-plugin-transform-contextual-imports/tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+  "compilerOptions": {
+    "target": "es2016",
+    "module": "commonjs",
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
+    "skipLibCheck": true
+  },
+  "include": ["src/*"],
+  "exclude": ["node_modules", ".vscode-test"]
```
MonicaOlejniczak, line 11: Is `.vscode-test` needed? Can you remove it?

JakeLane: This is just duplicated from the other TS package (`packages/utils/atlaspackforvscode/tsconfig.json`). Makes sense to drop it

MonicaOlejniczak (approved):

PR #98, opened 2024-10-01
Support loading CJS plugins

alshdavid:
ESM plugins are available under `module.default`
CJS plugins may be under `module.default.default`

Support both import patterns

### DIFF
--- a/packages/core/core/src/atlaspack-v3/worker/worker.js
+++ b/packages/core/core/src/atlaspack-v3/worker/worker.js
@@ -58,7 +58,6 @@ export class AtlaspackWorker {
     [RunResolverResolveOptions],
     Promise<RunResolverResolveResult>,
   > = jsCallable(async ({key, dependency, specifier}) => {
-    // console.log({threadId, key, specifier, resolvers: this.#resolvers })
     const resolver = this.#resolvers.get(key);
     if (!resolver) {
       throw new Error(`Resolver not found: ${key}`);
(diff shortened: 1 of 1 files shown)


marcins (approved):

yamadapc (approved):

packages/core/core/src/atlaspack-v3/worker/worker.js
```diff
@@ -32,9 +32,23 @@ export class AtlaspackWorker {
       // $FlowFixMe
       let resolvedModule = await import(resolvedPath);
 
+      let instance = undefined;
+      if (resolvedModule.default && resolvedModule.default[CONFIG]) {
+        instance = resolvedModule.default[CONFIG];
+      } else if (
+        resolvedModule.default &&
+        resolvedModule.default.default[CONFIG]
```
yamadapc, line 40: hm

yamadapc: what is this for?

alshdavid: Just realised I am missing a check there but either way;

When we use a native esm `import()` in Nodejs and import a cjs module, it puts `module.exports` into `default`.

```javascript
// foo.cjs
module.exports = { foo: 'foo' }
```

```javascript
// main.mjs
import foo from './foo.cjs'
import * as foo2 from './foo.cjs'
// import { foo } from './foo.cjs' // <-- illegal

console.log(foo.foo)
console.log(foo.default.foo)
```
Preprocessors that target cjs will typically convert esm exports to
```typescript
export default 'foo'
export const bar = 'bar'
```
Into 
```javascript
module.exports = {}
module.exports.default = 'foo'
module.exports.bar = 'bar'
```

So if I use an esm import, to access the default export I need to get it with 

```javascript
import foo1 from './foo.cjs'
import * as foo2 from './foo.cjs'
const foo3 = await import('./foo.cjs')

console.log(foo1.default)
console.log(foo2.default.default)
console.log(foo3.default.default)
```

PR #99, opened 2024-10-01
Migrate repository to typescript

yamadapc:
This pull request runs https://github.com/stripe-archive/flow-to-typescript-codemod

And also makes a few fixes to get it to work:

* Remove inline interfaces 6fc0af8936c91b86e025bc2414356273d89dd641
* Remove import extensions ce99540b23810e97bd6418e7811952a22f6daf32
* Use arrow parens 2e556f94063449b6b166a6feff2ce581e054f308 729c1198759d3e67df6dfdf3c275058ab7020865
* dfba9c915100b87f736934efff61c468ac23ae91
  * Update package.json source entries  
  * Update a few broken imports 
  * Keep the `hmr-rutime.js` file the same as before


yamadapc:
In order to make this workable we should use https://github.com/atlassian-labs/atlaspack/tree/origin/turbo

PR #100, opened 2024-10-01
Set-up turbo repo and move towards ahead of time compiled files

yamadapc:
Required for https://github.com/atlassian-labs/atlaspack/pull/99


crates/atlaspack_plugin_resolver/src/atlaspack_resolver.rs
```diff
@@ -361,7 +361,9 @@ impl ResolverPlugin for AtlaspackResolver {
       (atlaspack_resolver::Resolution::External, _query) => {
         if let Some(_source_path) = &ctx.dependency.source_path {
           if ctx.dependency.env.is_library && ctx.dependency.specifier_type != SpecifierType::Url {
-            todo!("check excluded dependency for libraries");
+            return Err(anyhow::anyhow!(
```
MonicaOlejniczak, line 364: Is this meant to be here? Also a lot of the checks for this code path in the JS version seem a bit overkill to me 🤷

packages/core/core/src/Atlaspack.js
```diff
@@ -130,7 +130,7 @@ export default class Atlaspack {
       rustAtlaspack = new AtlaspackV3({
         ...options,
         corePath: path.join(__dirname, '..'),
-        threads: process.env.NODE_ENV === 'test' ? 2 : undefined,
+        threads: process.env.NODE_ENV === 'test' ? 0 : undefined,
```
MonicaOlejniczak, line 133: Revert?

packages/core/integration-tests/test/BundleGraph.js
```diff
@@ -11,10 +11,10 @@ import {
 } from '@atlaspack/test-utils';
 import type {BundleGraph, BundleGroup, PackagedBundle} from '@atlaspack/types';
 
-describe.v2('BundleGraph', () => {
+describe('BundleGraph', () => {
```
MonicaOlejniczak, line 14: All these test changes should be in a separate PR to reduce noise, including the join `/` removal

PR #101, opened 2024-10-01
V3: Include stack traces for napi threadsafe functions

alshdavid:
This adds stack traces to error thrown in Napi threadsafe functions

I will use better types when we move to TypeScript

This requires that the Rust `JsCallable` is paired with the JavaScript `JsCallable`

```javascript
const reportError = jsCallable(() => {
  throw new Error('Hello World')
})
```

### DIFF
--- a/crates/atlaspack_napi_helpers/src/js_callable/js_callable.rs
+++ b/crates/atlaspack_napi_helpers/src/js_callable/js_callable.rs
@@ -103,16 +103,23 @@ impl JsCallable {
       ThreadsafeFunctionCallMode::NonBlocking,
       move |JsValue(value, env)| {
         let container: JsObject = value.try_into()?;
-        let then_fn: JsFunction = container.get_named_property("then")?;
+        let then_fn: JsFunction = match container.get_named_property("then") {
+          Ok(then_fn) => then_fn,
+          Err(error) => {
+            tx.send(Err(anyhow::anyhow!(error))).unwrap();
+            return Ok(());
+          }
+        };
 
         let then_callback = env.create_function_from_closure("callback", move |ctx| {
-          // Return [bool, promise]
-          let container = ctx.get::<JsObject>(0)?;
-
-          let error = container.get_element::<JsUnknown>(0)?;
+          // Return
+          //   [error, null]
+          //   [null, value]
+          let result = ctx.get::<JsObject>(0)?;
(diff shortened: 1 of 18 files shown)


crates/atlaspack_napi_helpers/src/js_callable/js_callable.rs
```diff
@@ -78,90 +79,65 @@ impl JsCallable {
...
-
-    Ok(())
-  }
-
-  pub fn call_serde<Params>(&self, params: Params) -> napi::Result<()>
-  where
-    Params: Serialize + Send + Sync + 'static,
-  {
-    self.call(map_params_serde(params))
-  }
```
alshdavid, line 106: We were not using this and it has footguns so I thought it was best to remove it

packages/core/core/src/atlaspack-v3/worker/worker.js
```diff
@@ -116,59 +109,63 @@ export class AtlaspackWorker {
...
+      let output = await transformer.generate({
         // $FlowFixMe
-        let output = await transformer.generate({
-          // $FlowFixMe
-          asset: assetCompat,
-          // $FlowFixMe
-          ast: assetCompat.getAST(),
-        });
+        asset: assetCompat,
         // $FlowFixMe
```
alshdavid, line 132: Wasn't able to figure out the flow type for this, will update when [this](https://github.com/atlassian-labs/atlaspack/pull/99) is merged :eyes:

packages/core/core/src/atlaspack-v3/fs.js
```diff
@@ -1,41 +1,46 @@
...
+    isFile: jsCallable((path: string) => {
       try {
         return fs.statSync(path).isFile();
       } catch {
         return false;
       }
-    },
-    isDir: (path: string) => {
+    }),
+    // $FlowFixMe migrate to TypeScript
```
alshdavid, line 37: Wasn't able to figure out the flow type for this, will update when https://github.com/atlassian-labs/atlaspack/pull/99 is merged 👀

yamadapc: That might take quite some time, so it's better to type things properly. Also untyped code will continue to be untyped, whereas typed code may be migrated with type-safety.

alshdavid:
I'll come back to this later

PR #104, opened 2024-10-02
V3: Resolver.loadConfig() and some arguments

alshdavid:
- Added `Resolver.loadConfig()`
- Added partial `PluginOptions`
- Added partial `PluginConfig`
- [TEMP] Initialize `NodeFS` for workers
- [TEMP] Initialize `PackageManger` in worker

NOTE: This would not pass integration tests because the `NodeFS` does not work there. Discussion on this might be to rewrite the integration tests to use the real FS for integration tests

### DIFF
--- a/packages/core/core/src/atlaspack-v3/worker/worker.js
+++ b/packages/core/core/src/atlaspack-v3/worker/worker.js
@@ -1,12 +1,17 @@
 // @flow
 import assert from 'assert';
 import * as napi from '@atlaspack/rust';
+import {NodeFS} from '@atlaspack/fs';
+import {NodePackageManager} from '@atlaspack/package-manager';
 import type {
   Resolver,
   Transformer,
   PluginOptions,
   Dependency,
   FilePath,
+  FileSystem,
+  ConfigResultWithFilePath,
+  PackageJSON,
 } from '@atlaspack/types';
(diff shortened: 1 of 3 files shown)


packages/core/core/src/atlaspack-v3/worker/worker.js
```diff
@@ -57,54 +66,150 @@ export class AtlaspackWorker {
...
+          get logLevel() {
+            throw new Error('Resolver.resolve.options.logLevel');
+          },
+          projectRoot,
+          get cacheDir() {
+            throw new Error('Resolver.resolve.options.cacheDir');
+          },
+          inputFS: this.#fs,
+          outputFS: this.#fs,
+          packageManager,
```
alshdavid, line 118: Inject this from napi, probably the same for `fs`

alshdavid: @mattcompiles, this is what I was saying earlier about making the Rust file system match the JavaScript one. It would allow us to inject it here in integration tests but use the NodeFS for real builds.

From what I can tell, plugins in JFE use the FS from `options` so it might be required

mattcompiles (approved):

MonicaOlejniczak (approved):

PR #106, opened 2024-10-04
Fix reading feature-flags when resolving options

yamadapc:
PR https://github.com/atlassian-labs/atlaspack/pull/72 added a feature-flag to switch to the new LMDB back-end, however that flag isn't correctly read because it is not set when resolving options.

This PR changes the feature-flags code to set feature-flags before anything else.

`AFB-786`

### DIFF
--- a/packages/packagers/js/src/ScopeHoistingPackager.js
+++ b/packages/packagers/js/src/ScopeHoistingPackager.js
@@ -24,6 +24,7 @@ import ThrowableDiagnostic, {
 } from '@atlaspack/diagnostic';
 import globals from 'globals';
 import path from 'path';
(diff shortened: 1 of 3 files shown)


packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -24,6 +24,7 @@ import ThrowableDiagnostic, {
 } from '@atlaspack/diagnostic';
 import globals from 'globals';
 import path from 'path';
+import {getFeatureFlag} from '@atlaspack/feature-flags';
```
yamadapc, line 27: Add missing import statement

JakeLane (approved):

PR #107, opened 2024-10-08
Improve write bundles performance

yamadapc:
Improves write bundles performance by rewriting hash_ref replacement in rust.

Removes streaming from this part of the code since the bundles aren't huge and most of the time they were loaded in memory already. This should reduce the memory footprint due to avoiding copies caused by piping into paused streams.

Performance should be at least 5x better, hopefully the memory usage drop will result in a reasonable improvement on overall build times.


packages/core/core/test/requests/WriteBundleRequest.test.js
```diff
@@ -0,0 +1,132 @@
...
+      if (err) reject(err);
+      else resolve(null);
+    });
+  });
+
+  const output = writable.getBuffer();
+  return output;
+}
+
+describe.only('replaceStream', () => {
```
yamadapc, line 69: remove

crates/node-bindings/Cargo.toml
```diff
@@ -29,6 +30,7 @@ mockall = "0.12.1"
 napi-derive = "2.16.3"
 num_cpus = "1.16.0"
 parking_lot = "0.12"
+regex = "1.11"
```
yamadapc, line 33: remove

packages/core/rust/index.js.flow
```diff
@@ -7,6 +7,8 @@ import type {
   PackageManager,
 } from '@atlaspack/types';
 
+import {decl} from 'postcss';
```
yamadapc, line 10: remove

yamadapc:
https://github.com/atlassian-labs/atlaspack/actions/runs/11249158483

PR #109, opened 2024-10-08
Update caniuse build and readme

MonicaOlejniczak:
## Motivation

These changes modify the caniuse database slightly to improve readability and usability ergonomics

## Changes

* Add caniuse database to workspaces and exclude it from flow checks
* Update readme wording slightly
* Update generate script in `package.json` to use `node` instead of `zx` so it can be more easily used in the project
* Improve generate script readability
  * Directly import `zx` dependency
  * Use import instead of require as per mjs standard
  * Proxy `console.log` to `write` so that every output fits on a single line and is easier to grok without `.`
  * Add do not modify comment header
  * Add empty lines between generated fns
  * Ensure output indent is two spaces to match the project settings
  * Refactor some `forEach` to be more concise


crates/caniuse_database/src/generated.rs
```diff
@@ -1,3 +1,9 @@
+/*
```
yamadapc, line 1: ultra nit is to use `//!` which is module level doc

crates/caniuse_database/README.md
```diff
@@ -9,9 +9,9 @@ version of file is vendored into the binary and is around 364KB.
 Alternatively the full database can be parsed and used to check support at
```
yamadapc, line 9: for what it's worth, this entire `README.md` is generated from the `src/lib.rs`

I used `cargo readme > ./README.md` to do this. The nice thing of doing that is that if we publish these crates or build the docs they have nice crate level docs.

The other nice thing is that the examples in this file are actually tests that run on our unit-tests jobs.

So it'd be nicer maybe to update the wording `src/lib.rs` then update this file with `cargo-readme`

This - https://crates.io/crates/cargo-readme

yamadapc (approved):

crates/caniuse_database/src/generated.rs
MonicaOlejniczak: I didn't want to use a module level doc, but happy to change it

crates/caniuse_database/README.md
MonicaOlejniczak: Oh I see, thanks! I'll update it

PR #111, opened 2024-10-09
[V3] Add native yaml transformer

MonicaOlejniczak:
## Motivation

These changes port the yaml transformer to Rust so that more functionality works in v3. It also removes the overhead of calling into JavaScript plugins by using Rust directly.

## Changes

* Add `atlaspack_plugin_transformer_yaml` crate
* Use yaml transformer when `@atlaspack/transformer-yaml` is encountered
* Enable yaml tests

## Checklist


yamadapc (approved):
nice!

PR #114, opened 2024-10-09
Fix bug in dependency ID generation

yamadapc:
This commit fixes a bug where on certain cases different dependencies would be given the same ID.

The root cause of this problem was that the `Target` type on JavaScript and rust have a mismatch, and two fields were missing on the Rust type. For this reason, the hashes were not considering the `pipeline` and `source` fields, which were previously considered.

There are changes made to address this:

* A catch-all `extra` value is added to the rust `Target`, so that any extra properties are considered
* Long term, just adding the missing fields will handle them from rust and javascript alike.

And:

* Where we were passing `Environment` as an object we just pass the `environment_id`; this:
  * saves some copying
  * prevents this problem which would be possible on all nested object parameters and hard to spot


mattcompiles (approved):
Looks good mate. You have some failing unit tests due to the changes ids though.

PR #115, opened 2024-10-10
V3: Resolver API

alshdavid:
All Resolvers in JFE now pass

- Added compat types for napi structures that need to be used in JavaScript
  - `Dependency`
  - `Environment`
  - `Target`
  - `MutableDependencySymbols`
- Added compat mappings for bit types (`BundleBehavior`, etc)

### DIFF
--- a/packages/core/core/src/atlaspack-v3/worker/compat/dependency.js
+++ b/packages/core/core/src/atlaspack-v3/worker/compat/dependency.js
@@ -1,0 +1,83 @@
+// @flow
+
+import type {Dependency as NapiDependency} from '@atlaspack/rust';
+import type {
+  Dependency as ClassicDependency,
+  DependencySpecifier,
+  SpecifierType,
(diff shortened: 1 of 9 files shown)


packages/core/core/src/atlaspack-v3/worker/compat/dependency.js
```diff
@@ -0,0 +1,83 @@
...
+import {MutableDependencySymbols} from './asset-symbols';
+import {
+  bundleBehaviorMap,
+  specifierTypeMap,
+  dependencyPriorityMap,
+  packageConditionsMap,
+} from './bitflags';
+
+export class Dependency implements ClassicDependency {
+  env: ClassicEnvironment;
```
alshdavid, line 28: `IEnvironment` or `EnvironmentInterface`?

MonicaOlejniczak (approved):

PR #116, opened 2024-10-10
Bump SWC to 1.7.26

yamadapc:
Bump SWC to 1.7.26, only where used from JavaScript.

### DIFF
--- a/packages/core/integration-tests/test/macros.js
+++ b/packages/core/integration-tests/test/macros.js
@@ -595,7 +595,7 @@ describe.v2('macros', function () {
     });
 
     res = await overlayFS.readFile(b.getBundles()[0].filePath, 'utf8');
-    let match2 = res.match(/output=(\d+)/);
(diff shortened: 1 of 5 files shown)


yamadapc:
I haven't updated types ; since we're on `flow` this is not a type-safe upgrade at all.

`flowgen` fails to parse the latest version of `@swc/types`.

JakeLane (approved):

packages/core/integration-tests/test/macros.js
```diff
@@ -595,7 +595,7 @@ describe.v2('macros', function () {
     });
 
     res = await overlayFS.readFile(b.getBundles()[0].filePath, 'utf8');
-    let match2 = res.match(/output=(\d+)/);
+    let match2 = res.match(/output=(.*);/);
```
yamadapc, line 598: The new SWC output just looks like so the regex here needed to be updated since before this nº likely was in decimal.
```
output=0x19278afe4dc;
```

PR #118, opened 2024-10-11
V3: Made RPC more Nodejs specific

alshdavid:
Previously, rpc was one struct with all the methods plugins use on it.

This PR changes the rpc system so the rpc worker returns plugins directly allowing each plugin to handle its calls individually.

The new interface looks like this;

```rust
pub trait RpcFactory: Send + Sync {
  fn start(&self) -> anyhow::Result<Arc<dyn RpcWorker>>;
}

pub trait RpcWorker: Send + Sync {
  fn create_bundler(&self, ctx: &PluginContext, plugin: &PluginNode) -> anyhow::Result<Box<dyn BundlerPlugin>>;
  fn create_compressor(&self, ctx: &PluginContext, plugin: &PluginNode) -> anyhow::Result<Box<dyn CompressorPlugin>>;
  // ...
}
```

This allows Nodejs+ plugins to be more closely associated with their bindings. This helps with re-mapping the Rust types from the `atlaspack` crate before sending them into JavaScript land


crates/atlaspack_plugin_rpc/src/nodejs/plugins/nodejs_rpc_resolver.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RunResolverResolve {
+  pub key: String,
+  pub dependency: Dependency,
+  pub specifier: String,
+  pub pipeline: Option<String>,
+  pub project_root: PathBuf,
+}
```
alshdavid, line 97: Now the types for the plugin can sit next to the plugin ^

crates/atlaspack_plugin_rpc/src/nodejs/rpc/nodejs_rpc_worker.rs
alshdavid, line 1: I haven't implemented it here but if we use channels to send types to the Nodejs workers rather than `ThreadSafeFunction`, we can avoid the bindings here and also simplify the worker bootstrap logic. 

It would also allow workers to be bootstrapped lazily, right now the build is blocked until they are initialized (minor, +100ms to a build)

crates/atlaspack_plugin_rpc/src/testing.rs
alshdavid, line 1: Could not use automock here so I wrote a manual mock

crates/atlaspack_plugin_rpc/src/testing.rs
```diff
@@ -0,0 +1,192 @@
...
+    Ok(None)
+  }
+}
+
+impl OptimizerPlugin for TestingRpcPlugin {
+  fn optimize(&self, _ctx: OptimizeContext) -> Result<OptimizedBundle, anyhow::Error> {
+    anyhow::bail!("Mock Optimizer Plugin Incomplete")
+    // Ok(OptimizedBundle {
+    //   contents: fs::File::create(),
+    // })
```
alshdavid, line 141: Cannot create a mock file without hitting the real filesystem. Will revisit this when we get to this stage

crates/atlaspack_plugin_rpc/src/lib.rs
alshdavid, line 1: Future note - I will add this to the main `RpcFactory`:

```rust
pub trait RpcFactory: Send + Sync {
  // ...
  resolve(&self, cwd: &Path, specifier: &str) -> anyhow::Result<PathBuf>
}
```
This will be used in Rust land to resolve plugin specifiers, allowing us to pass an absolute path into `loadPlugin`.

The initial implementation will delegate to Nodejs for resolution, but will be replaced our internal rust resolver.

MonicaOlejniczak (approved):

PR #119, opened 2024-10-11
V3: Integrate native html transformer

MonicaOlejniczak:
## Motivation

These changes introduce more features to the native html transformer. Since most of the functionality is in a working state, it has been integrated into the v3 plugin system, which will eliminate the overhead of calling into JavaScript by using Rust directly.

## Changes

* Fix discovered asset path pipelines to work with inline assets
* Add `new_inline` fn to asset and use it in the html transformer
* Add `htm` file type support
* Walk dom nodes in the correct order using `VecDeque` which ensures that scripts are outputted in the order they are encountered (not reversed)
* Update dependencies visitor to handle most inline attributes and inline scripts
* Fix `EsModule` assignment check in js transformer
* Noop `@atlaspack/transformer-posthtml` for now, as it's not required or used in products
* Use html transformer when `@atlaspack/transformer-html` is encountered
* Enable most html tests, with the exception of tests that include css and svgs
* Manually input html entries instead of using globs in html tests to increase test coverage as the functionality does not work in v3

## Checklist


crates/atlaspack_core/src/types/asset.rs
```diff
@@ -253,22 +254,63 @@ impl Asset {
...
+    let environment_id = env.id();
+    let mut hasher = crate::hash::IdentifierHasher::default();
+
+    environment_id.hash(&mut hasher);
+    code.bytes().hash(&mut hasher);
+    file_type.hash(&mut hasher);
+    unique_key.hash(&mut hasher);
+
+    // Ids must be 16 characters for scope hoisting to replace imports correctly in REPLACEMENT_RE
+    let asset_id = format!("{:016x}", hasher.finish());
```
mattcompiles, line 292: Is there a reason you're not using `create_asset_id`?

MonicaOlejniczak: I originally passed in significantly fewer options to `new_inline`, and can probably reuse it more easily now

alshdavid (approved):

crates/atlaspack_core/src/types/asset.rs
MonicaOlejniczak: Should be fixed now, let me know if the type changes are ok

PR #120, opened 2024-10-11
Fix bug in inline requires

yamadapc:
The rust inline requires visitor wasn't doing anything on output bundles because the `parcelRequire = ...` symbol isn't always an unresolved global.

### DIFF
--- a/packages/core/integration-tests/test/inline-requires.js
+++ b/packages/core/integration-tests/test/inline-requires.js
@@ -31,7 +31,7 @@ import {
             }
 
           other.js:
-            // this is here so that we don't scope hoist a.js
+            // this is here so that we don't scope hoist dependency/index.js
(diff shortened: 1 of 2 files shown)


packages/core/integration-tests/test/inline-requires.js
```diff
@@ -0,0 +1,117 @@
...
+    it('inlines require statements', async () => {
+      await fsFixture(overlayFS, __dirname)`
+        inline-requires
+          dependency/index.js:
+            export function exportedFunction() {
+              throw new Error("Shouldn't be called");
+            }
+
+          other.js:
+            // this is here so that we don't scope hoist a.js
```
Nikola-3, line 34: `a.js`?

JakeLane (approved):

PR #123, opened 2024-10-14
Log on empty files when building with an env flag

Nikola-3:
## Motivation

This is an action to let us debug faster if we have issues with symbol propagation/scope hoisting involving empty files again.

## Changes

* Added a flagged logging statement
* Unit tests both for new value being set and logging being omitted when flag not set
* Removed some redundant code
* Moved utility function where I thought more appropriate and change usages to import it from new location

## Checklist


yamadapc (approved):

packages/transformers/js/core/src/lib.rs
```diff
@@ -503,6 +507,23 @@ pub fn transform(
                 config.conditional_bundling,
               );
               module.visit_with(&mut collect);
+
+              if collect.is_empty_or_empty_export {
+                // The above should be false almost always and we already have the value
+                // so it's better to skip fetching the flag when unnecessary
+                let should_look_for_empty_files = std::env::var("ATLASPACK_SHOULD_LOOK_FOR_EMPTY_FILES");
+                if should_look_for_empty_files.is_ok() && should_look_for_empty_files.unwrap() == "true" {
```
yamadapc, line 515:
```suggestion
                if should_look_for_empty_files.ok() == Some("true") {
```

PR #125, opened 2024-10-14
Remove egregious cases of code duplication in SymbolProp

Nikola-3:
## Motivation

As discussed in the sync I will be doing a short series of obvious refactors to move us to a nicer `SymbolPropagation.js` version that will be slightly easier to maintain until we rewrite it in rust

## Changes

Simple refactor removing some obvious code duplication and simplifying a for loop.

### DIFF
--- a/packages/core/core/src/SymbolPropagation.js
+++ b/packages/core/core/src/SymbolPropagation.js
@@ -558,16 +558,8 @@ function propagateSymbolsDown(
     } else if (node.type === 'asset' && node.usedSymbolsDownDirty) {
       visit(
         node,
-        assetGraph.getIncomingDependencies(node.value).map((d) => {
-          let dep = assetGraph.getNodeByContentKey(d.id);
-          invariant(dep && dep.type === 'dependency');
-          return dep;
-        }),
-        outgoing.map((dep) => {
-          let depNode = nullthrows(assetGraph.getNode(dep));
-          invariant(depNode.type === 'dependency');
-          return depNode;
-        }),
+        incomingDependencyNodesToAsset(assetGraph, node.value),
+        dependencyNodesFromIds(assetGraph, outgoing),
       );
       node.usedSymbolsDownDirty = false;
     }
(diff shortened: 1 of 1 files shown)


packages/core/core/src/SymbolPropagation.js
```diff
@@ -558,16 +558,8 @@ function propagateSymbolsDown(
...
-          let dep = assetGraph.getNodeByContentKey(d.id);
-          invariant(dep && dep.type === 'dependency');
-          return dep;
-        }),
-        outgoing.map((dep) => {
-          let depNode = nullthrows(assetGraph.getNode(dep));
-          invariant(depNode.type === 'dependency');
-          return depNode;
-        }),
+        incomingDependencyNodesToAsset(assetGraph, node.value),
```
MonicaOlejniczak, line 561: This is to a dependency, so should probably be named more like `incomingDependencyNodes`

packages/core/core/src/SymbolPropagation.js
```diff
@@ -645,43 +637,24 @@ function propagateSymbolsUp(
 
     const nodeVisitor = (nodeId) => {
       let node = nullthrows(assetGraph.getNode(nodeId));
-      let outgoing = assetGraph.getNodeIdsConnectedFrom(nodeId);
 
-      for (let childId of outgoing) {
-        let child = nullthrows(assetGraph.getNode(childId));
-        if (node.type === 'asset') {
-          invariant(child.type === 'dependency');
+      if (node.type === 'asset') {
```
MonicaOlejniczak, line 641: 👍 for moving this up

packages/core/core/src/SymbolPropagation.js
Nikola-3: Nice find 😅, I think I would prefer to go with `incomingDependencyNodesFromAsset`. I find there is an issue with constant conflation of nodes and node ids and so it gets a bit confusing for me since intuitively, a node can be equated to it's id since it's always one step away, but when working with the code there is a lack of clarity what a variable represents. This is partly on flow not properly working with my IDE but even still, having the clarity be an explicit part of the code helps when reading the code in my opinion.
If I wasn't aiming to keep the diff super small, I would've renamed a ton of the variables in this file to add extra understanding of what they represent.

MonicaOlejniczak (approved):

PR #128, opened 2024-10-15
Skip toProjectRoot relative paths when already relative

MonicaOlejniczak:
## Motivation

These changes are similar to https://github.com/atlassian-labs/atlaspack/pull/81 but instead uses code that has been applied in Confluence for a while. Specifically when the cwd and project root are different, paths within symbols and loc would be processed twice resulting in invalid paths.

## Changes

Skip `toProjectPath` handling when the path is already relative from Atlaspack's point of view

## Checklist

### DIFF
--- a/packages/core/core/src/projectPath.js
+++ b/packages/core/core/src/projectPath.js
@@ -9,14 +9,22 @@ import {relativePath, normalizeSeparators} from '@atlaspack/utils';
 export opaque type ProjectPath = string;
 
 function toProjectPath_(projectRoot: FilePath, p: FilePath): ProjectPath {
+  // If the file path is not provided, then treat it as though it is already from the project root
   if (p == null) {
     return p;
   }
 
-  // If the file is outside the project root, store an absolute path rather
-  // than a relative one. This way if the project root is moved, the file
-  // references still work. Accessing files outside the project root is not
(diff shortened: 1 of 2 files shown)


yamadapc (approved):

packages/core/core/src/projectPath.js
```diff
@@ -9,14 +9,21 @@ import {relativePath, normalizeSeparators} from '@atlaspack/utils';
...
   if (p == null) {
     return p;
   }
 
-  // If the file is outside the project root, store an absolute path rather
-  // than a relative one. This way if the project root is moved, the file
-  // references still work. Accessing files outside the project root is not
-  // portable anyway.
+  // If the file path is already relative and it does not begin with . then treat the path as it is
+  // already from the project root. This prevents double relative paths from occurring within
```
yamadapc, line 18: What is a double relative path? Is it possible to unit-test what this is fixing?

yamadapc: Okay to merge this first ; to remove patches from usages.

MonicaOlejniczak: I'll reword it a tad, I tried testing it but it's a bit of a pain

PR #129, opened 2024-10-15
[V3] Add the native CSS transformer

mattcompiles:
## Motivation

This PR adds the initial implementation of the native CSS transformer. Its is mostly a direct port of the existing CSS transformer as they both use lightningcss as their parser. There are still some missing pieces which I'm going to immediately follow up on but I wanted to get this open as the PR was getting bigger and there's value in it's current state. 

I've also disable the `postcss` transformer in native as it blocks a lot of the tests. I think we should remove this completely, or at least massively reduce it's scope as it currently has it's own CSS modules implementation. 

The following is what's remaining:

- Inline style attribute support e.g. `<div style="display: block" />`
  - Probably not high value, a lot of tests use this feature which blocks enabling them 
- CSS modules is working in the transformer, but the asset_graph logic for them is currently broken. 
  - This is my priority to fix next as it's needed for AFM and should unlock a bunch more integration tests.
- Proper CSS module config support. Right now I've just implemented some defaults that should work for AFM, but it shouldn't be too hard to support the rest of the config options.

## Changes

- Added native CSS transformer

## Checklist


crates/atlaspack_plugin_transformer_css/src/css_transformer.rs
```diff
@@ -0,0 +1,439 @@
...
+    asset: Asset,
+  ) -> Result<TransformResult, Error> {
+    // TODO: Hardcoded AFM defaults for now, could support proper config here if
+    // need be
+    let is_style_tag = asset
+      .meta
+      .get("type")
+      .is_some_and(|meta_type| *meta_type == json!("tag"));
+    let is_css_module = asset.is_source && !is_style_tag && {
+      let file_path_string = asset
```
MonicaOlejniczak, line 47: /nit lossy conversion should be okay and more succinct for this check

```rust
asset
  .file_path
  .file_name()
  .is_some_and(|name| name.to_string_lossy().ends_with(".module.css"));
```

crates/atlaspack_plugin_transformer_css/src/css_transformer.rs
```diff
@@ -0,0 +1,436 @@
...
+    types::JSONObject,
+  };
+  use atlaspack_filesystem::in_memory_file_system::InMemoryFileSystem;
+
+  use super::*;
+
+  #[test]
+  fn supports_css_imports() {
+    let file_system = Arc::new(InMemoryFileSystem::default());
+    let mut plugin = AtlaspackCssTransformerPlugin::new(&PluginContext {
```
MonicaOlejniczak, line 354: I would make the boilerplate a function for both tests

crates/atlaspack_plugin_transformer_css/src/css_transformer.rs
```diff
@@ -0,0 +1,436 @@
...
+      },
+      ..asset.env.deref().clone()
+    });
+
+    let preserve_imports = asset
+      .meta
+      .get("hasDependencies")
+      .map_or_else(|| true, |value| value != false);
+
+    let css = stylesheet.to_css(PrinterOptions {
```
MonicaOlejniczak, line 97: Is the drafts option not available in lightning? I noticed it in JS, but not anywhere in this file

crates/atlaspack_plugin_transformer_css/src/css_transformer.rs
```diff
@@ -0,0 +1,436 @@
...
+          .collect()
+      })
+      .unwrap_or_default();
+
+    let mut css_code = Vec::new();
+    let mut discovered_assets = Vec::new();
+    let mut asset_symbols = Vec::new();
+
+    if let Some(exports) = css.exports {
+      let unique_key = format!("{}:module", asset.id);
```
MonicaOlejniczak, line 176: The original seems to also mutate the js asset unique key as this?

```
 asset.uniqueKey ??= asset.id;
```

MonicaOlejniczak (approved):

crates/atlaspack_plugin_transformer_css/src/css_transformer.rs
mattcompiles: Yeah I was unsure about this one as well. I'm honestly not sure why. To me it makes sense that only the discovered assets have this set. I decided to leave it out until I found a reason for it.

MonicaOlejniczak: All good, it's weird to me too

crates/atlaspack_plugin_transformer_css/src/css_transformer.rs
mattcompiles: I think the settings all exist but not under `drafts`.

PR #133, opened 2024-10-17
[V3] CSS improvements and fixes

mattcompiles:
## Motivation

This PR improves the CSS transformer and makes changes to the asset graph request to support CSS modules. Most CSS features are now supported except for the following: 
- [CSS module composes](https://github.com/css-modules/css-modules/blob/master/docs/composition.md) which can be added later if needed.
- Config options
  - exclude/include globs
  - [custom-naming-patterns](https://parceljs.org/languages/css/#custom-naming-patterns)
  - [custom-media-queries](https://parceljs.org/languages/css/#custom-media-queries)
  - [pseudo-class-replacement](https://parceljs.org/languages/css/#pseudo-class-replacement)

None of which are currently used in AFM. 

## Changes

- Support CSS transformer config loading
  - Global CSS modules
  - Dashed idents
- Consistent CSS module detection against v2
- Browserslist support for lightning transpilation
- Symbol/Dependency fixes for CSS modules
- Fixed attaching "direct" discovered assets
  - "direct" referring to the fact that they are not pointed to by any generated dependencies and therefore need to be attached to the root dependency directly
- Fixed an infinite loop in the asset graph request when self referencing dependencies are added
- Enable a bunch of tests in the css and css-module integration suites

## Checklist

### DIFF
--- a/crates/atlaspack_core/src/asset_graph.rs
+++ b/crates/atlaspack_core/src/asset_graph.rs
@@ -263,7 +263,10 @@ impl AssetGraph {
           .edges_directed(dep_node, Direction::Outgoing)
           .next()
         {
-          self.propagate_requested_symbols(resolved.target(), dep_node, on_undeferred);
+          // Avoid infintite loops for self references
+          if resolved.target() != asset_node {
+            self.propagate_requested_symbols(resolved.target(), dep_node, on_undeferred);
+          }
         } else {
           on_undeferred(dep_node, Arc::clone(&dependency));
         }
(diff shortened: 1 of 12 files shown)


crates/atlaspack_core/src/asset_graph.rs
```diff
@@ -263,7 +263,10 @@ impl AssetGraph {
           .edges_directed(dep_node, Direction::Outgoing)
           .next()
         {
-          self.propagate_requested_symbols(resolved.target(), dep_node, on_undeferred);
+          // Avoid infintite loops for self references
```
MonicaOlejniczak, line 266: Noting that this might fix the hmr tests for the html transformer, I didn't investigate the infinite loop any further at the time

MonicaOlejniczak (approved):

PR #134, opened 2024-10-18
Fix worker paths and add missing filename

MonicaOlejniczak:
## Motivation

When migrating to TypeScript, all of the worker paths break due to the change in extension. Additionally, the babel dependency in the eslint plugin changes slightly and requires a filename in the tests.

## Changes

* Resolve worker paths without an extension
* Use a public export name for the worker paths using the development condition
* Add filename to eslint parser

## Checklist


yamadapc (approved):
awesome

PR #139, opened 2024-10-21
[V3] Handle should optimize false correctly

mattcompiles:
## Motivation

Prior to this change `should_optimize` would be set to true when a prod build occurred but `--no-optimize` was passed.

## Checklist

### DIFF
--- a/crates/atlaspack/src/requests/target_request.rs
+++ b/crates/atlaspack/src/requests/target_request.rs
@@ -384,7 +384,10 @@ impl TargetRequest {
             .default_target_options
             .output_format
             .unwrap_or_else(|| fallback_output_format(context)),
-          should_optimize: self.default_target_options.should_optimize,
+          should_optimize: self
+            .default_target_options
+            .should_optimize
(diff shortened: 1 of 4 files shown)


crates/atlaspack/src/requests/target_request.rs
```diff
@@ -486,6 +489,7 @@ impl TargetRequest {
     let is_library = target_descriptor
       .is_library
       .unwrap_or_else(|| self.default_target_options.is_library.unwrap_or(false));
+    println!("isLibrary {}", is_library);
```
yamadapc, line 492: trace if'll keep it

crates/atlaspack/src/requests/target_request.rs
```diff
@@ -654,6 +654,7 @@ mod tests {
   use atlaspack_filesystem::in_memory_file_system::InMemoryFileSystem;
 
   use crate::test_utils::{request_tracker, RequestTrackerTestOptions};
+  use pretty_assertions::assert_eq;
```
yamadapc, line 657: 🆒

yamadapc (approved):

PR #141, opened 2024-10-22
[V3] Small refactor napi plugins

alshdavid:
Contributing small cleanups that were part of my config loader branch, excluding the config loader stuff

**Changelog**
- Updated generics on `find_ancestor` helpers
- Added lazily initialized `InitializedState` to napi plugins
  - For `RpcNodejsResolverPlugin`
  - For  `NodejsRpcTransformerPlugin`
- Renamed flow types in binding layer from `ClassicFoo` to use `IFoo` naming convention


crates/atlaspack_filesystem/src/search.rs
```diff
@@ -41,24 +45,37 @@ pub fn find_ancestor<'a, P: AsRef<Path>>(
   None
 }
```
alshdavid, line 46: Added this because it allows `from` and `root` to be different types

crates/atlaspack_plugin_rpc/src/nodejs/plugins/nodejs_rpc_resolver.rs
```diff
@@ -38,54 +42,60 @@ impl Debug for RpcNodejsResolverPlugin {
...
+    nodejs_workers: Arc<NodeJsWorkerCollection>,
     ctx: &PluginContext,
-    plugin: &PluginNode,
+    plugin_node: &PluginNode,
   ) -> Result<Self, anyhow::Error> {
-    let plugin_options = RpcPluginOptions {
-      hmr_options: None,
-      project_root: ctx.options.project_root.clone(),
-      mode: ctx.options.mode.clone(),
-    };
```
alshdavid, line 49: Moved this into the one-time plugin init under `self.get_or_init_state()`

packages/core/core/src/atlaspack-v3/fs.js
```diff
@@ -4,15 +4,15 @@ import type {FileSystem, JsCallable} from '@atlaspack/rust';
 import type {
   Encoding,
   FilePath,
-  FileSystem as ClassicFileSystem,
-} from '@atlaspack/types-internal';
+  FileSystem as IFileSystem,
```
MonicaOlejniczak, line 7: What was wrong with classic file system?

alshdavid: Just trying to be consistent in the naming the JS interfaces, can revert this if it's contentious

MonicaOlejniczak (approved):

PR #142, opened 2024-10-22
Change fallback behaviour for __MCOND and fix dependency for runtime

JakeLane:
## Motivation
We need a safe rollout when `globalThis.__MCOND` is not defined (if we mess something up on the control branch of the rollout). We also have an issue where some dependencies don't get loaded correctly when prod mode is on.

## Changes

Two major changes here:

- Check if `globalThis.__MCOND` exists when configured, so we can have a safe rollout when feature flag is disabled
- Fix the dependency in scope hoisting, as we don't pick up the correct dep

I also added a new integration test and fixed some tests not using the correct data. This can probably allow some refactoring in the style of the new test. Dev mode is broken so I skip the test I fixed (as it's now red). 

## Checklist

### DIFF
--- a/crates/atlaspack_contextual_imports/src/lib.rs
+++ b/crates/atlaspack_contextual_imports/src/lib.rs
@@ -19,13 +19,17 @@ use swc_core::{
 #[derive(Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct ContextualImportsConfig {
-  pub server: Option<bool>,
+  #[serde(default)]
+  pub server: bool,
+  #[serde(default)]
+  pub default_if_undefined: bool,
 }
 
 impl Default for ContextualImportsConfig {
   fn default() -> ContextualImportsConfig {
     ContextualImportsConfig {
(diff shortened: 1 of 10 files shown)


crates/atlaspack_contextual_imports/Cargo.toml
```diff
@@ -8,7 +8,7 @@ description = "Atlaspack contextual imports transformation"
 [dependencies]
 serde = { version = "1.0.200", features = ["derive"] }
 serde_json = "1.0.116"
-swc_core = { version = "0.96", features = ["ecma_plugin_transform"] }
+swc_core = { version = "0.96", features = ["ecma_plugin_transform", "ecma_quote", "ecma_utils"] }
```
MonicaOlejniczak, line 11: you'll need to rebase

JakeLane: thanks for the heads up, fixed now

MonicaOlejniczak (approved):

PR #147, opened 2024-10-23
V3: Temporarily use require.resolve() as Transformer resolve() function for JS workers

alshdavid:
Temporarily use require.resolve() as the resolve() function in JavaScript workers

### DIFF
--- a/packages/core/core/src/atlaspack-v3/worker/worker.js
+++ b/packages/core/core/src/atlaspack-v3/worker/worker.js
@@ -189,8 +189,10 @@ export class AtlaspackWorker {
       const ast = await transformer.parse({
         asset: mutableAsset,
         config,
-        get resolve() {
-          throw new Error('Transformer.parse.resolve()');
+        async resolve(specifier: string, options: any) {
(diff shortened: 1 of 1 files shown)


packages/core/core/src/atlaspack-v3/worker/worker.js
```diff
@@ -189,8 +189,10 @@ export class AtlaspackWorker {
       const ast = await transformer.parse({
         asset: mutableAsset,
         config,
-        get resolve() {
-          throw new Error('Transformer.parse.resolve()');
+        async resolve(specifier: string, options: any) {
+          // TODO Temp, replace with resolver pipeline
+          // eslint-disable-next-line no-return-await
+          return await require.resolve(specifier, options);
```
mattcompiles, line 195: The atlaspack resolve API is `(from, to) => string` so you need to map to something like:
```
require.resolve(to, {paths: [from]})
```

alshdavid: Updated

mattcompiles (approved):

PR #150, opened 2024-10-23
Add single file output option

benjervis:
Some SSR runtime environments are limited to only loading a single JS file to do rendering, which means Atlaspack needs to have a way to combine all assets into a single bundle.

To achieve this, I've added a monolithic bundler algorithm to the default bundler. The `bundle` method looks through the targets and separates them into single file and regularly bundled entries based on the new `__unstable_singleFileOutput` option.

The createIdealGraph function creates new bundle graphs which then get legacy decorated, but the single file bundler just mutates the existing bundle graph. This means that the single file entries need to run after the ideal graph entries, so that the ideal graph has a clean bundle graph to work from.

### DIFF
--- a/packages/transformers/js/src/JSTransformer.js
+++ b/packages/transformers/js/src/JSTransformer.js
@@ -863,7 +863,7 @@ export default (new Transformer({
           specifierType: dep.kind === 'Require' ? 'commonjs' : 'esm',
           loc: convertLoc(dep.loc),
           priority:
-            dep.kind === 'DynamicImport' && !asset.env.unstableSingleFileOutput
+            dep.kind === 'DynamicImport'
               ? 'lazy'
               : dep.kind === 'ConditionalImport'
               ? 'conditional'
--- a/packages/transformers/js/src/JSTransformer.js
+++ b/packages/transformers/js/src/JSTransformer.js
@@ -863,7 +863,7 @@ export default (new Transformer({
           specifierType: dep.kind === 'Require' ? 'commonjs' : 'esm',
           loc: convertLoc(dep.loc),
           priority:
-            dep.kind === 'DynamicImport'
+            dep.kind === 'DynamicImport' && !asset.env.unstableSingleFileOutput
               ? 'lazy'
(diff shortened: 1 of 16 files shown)


packages/transformers/js/src/JSTransformer.js
```diff
@@ -863,7 +863,7 @@ export default (new Transformer({
           specifierType: dep.kind === 'Require' ? 'commonjs' : 'esm',
           loc: convertLoc(dep.loc),
           priority:
-            dep.kind === 'DynamicImport'
+            dep.kind === 'DynamicImport' && !asset.env.unstableSingleFileOutput
```
yamadapc, line 866: It'd be better/more reliable if we write a bundler that puts everything in one file, it's relatively straightforward I think to write

benjervis: Hey @yamadapc , I've updated it to happen in the bundler step now

yamadapc:
This looks great! Ideally we would move towards completely splitting this functionality onto its own bundler.
So you'd have a new totally separate bundler much like the `LibraryBundler`. The reason for this is:

* It's easier to test
* It's clearer that you've a separate algorithm that does not depend on the mutations done in ideal graph and decorate functions

Since you've mentioned that we are already using the same configuration for the monolithic bundles and the others, and hence this quality suggestion is not possible to implement.

It'd be good to make it explicit that the added functionality is not dependant on those values. Since this default bundler file is already a 1800LOC file that has bugs we have a hard time debugging and understanding, it'd be preferable that you split the implementation moving towards a state where there are two isolated bundling implementations.

You've already done all the work, this is going to look like:
```
// MonolithicBundler.js

export function bundleMonolithicTargets(...) {
    // same as in the DefaultBundler or pass as a parameter
    let targetMap = ...;
    let graphs = [];

    for (let entries of targetMap.values()) {
       // build single file entries
       // run createJSMonolithGraph
    }
} 
```

On `DefaultBundler.js` to the opposite filtering.

yamadapc (approved):

PR #154, opened 2024-10-24
WIP: V3: Static resolve() function for JS plugins

alshdavid:
WIP

Adds static resolve() function to plugins


crates/atlaspack/src/plugins/config_plugins.rs
```diff
@@ -216,6 +216,7 @@ impl Plugins for ConfigPlugins {
         resolvers.push(rpc_worker.create_resolver(&self.ctx, resolver)?);
       }
 
+      self.ctx.static_resolver.init(resolvers.clone());
```
alshdavid, line 219: Super hacky :/

alshdavid:
Think of this as an RFC - will come back to it if it's needed

PR #156, opened 2024-10-27
Optimize resolver with threadlocal caches

yamadapc:
On https://github.com/atlassian-labs/atlaspack/pull/10 we have removed `dashmap` from this project dependencies.

The reason for that is we were observing an alarming amount of crashes
related to memory bugs on that crate on developer machines. We have
confirmed that those bugs were in the `dashmap` crate and they stopped
happening after it was removed.

However, the solution put in place made resolution around 3-5x slower
than what it was before with `dashmap` caches. That is due to contention
on the `RwLock` hashmap used for resolver internal caches.

This change moves those caches to use a thread-local map, which achieves
close performance to the `dashmap` solution, around 30% slower.

The entire resolution task can be completed very quickly on a large
project (few seconds), hence we prefer to have a more stable
implementation for now.

## Motivation

* Improve performance of the resolver

## Changes

Move resolver internal caches to use thread-local storage.

## Checklist


crates/atlaspack_filesystem/Cargo.toml
```diff
@@ -8,6 +8,8 @@ description = "FileSystem wrapper trait for use in Atlaspack codebase."
 mockall = { workspace = true }
 xxhash-rust = { workspace = true, features = ["xxh3"] }
 anyhow = { workspace = true }
+thread_local = { workspace = true }
+atlaspack_shared_map = { path = "../atlaspack_shared_map" }
```
MonicaOlejniczak, line 12: /nit alphabetise the deps

crates/atlaspack_filesystem/src/lib.rs
```diff
@@ -16,8 +15,10 @@ pub mod os_file_system;
 /// This should be `OsFileSystem` for non-testing environments and `InMemoryFileSystem` for testing.
 pub type FileSystemRef = Arc<dyn FileSystem + Send + Sync>;
 
+pub type DefaultHasher = xxhash_rust::xxh3::Xxh3Builder;
+
 pub type FileSystemRealPathCache =
-  RwLock<HashMap<PathBuf, Option<PathBuf>, xxhash_rust::xxh3::Xxh3Builder>>;
+  atlaspack_shared_map::ThreadLocalHashMap<PathBuf, Option<PathBuf>, DefaultHasher>;
```
MonicaOlejniczak, line 21: does this really need to be namespaced? the line is quite long and definitions follow through easily

crates/atlaspack_shared_map/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
```
MonicaOlejniczak, line 1: feel free to add a desc and yourself as the author

packages/utils/node-resolver-rs/Cargo.toml
```diff
@@ -27,6 +28,7 @@ thiserror = { workspace = true }
 url = { workspace = true }
 xxhash-rust = { workspace = true, features = ["xxh3"] }
 tracing = { workspace = true }
+thread_local = { workspace = true }
```
MonicaOlejniczak, line 31: /nit above tracing

crates/atlaspack_filesystem/src/os_file_system/canonicalize.rs
```diff
@@ -28,9 +28,7 @@ pub fn canonicalize(path: &Path, cache: &FileSystemRealPathCache) -> std::io::Re
           ret.push(c);
 
           // First, check the cache for the path up to this point.
-          let read = cache.read().unwrap();
-          let cached = read.get(&ret).cloned();
-          drop(read);
+          let cached = cache.get(&ret);
```
MonicaOlejniczak, line 31: This is a lot nicer now 👍

MonicaOlejniczak (approved):

PR #157, opened 2024-10-28
Pass the LMDB database reference back into rust to share cache read/write

yamadapc:
We can avoid serialising data back to JavaScript from the rust asset graph builder by writing it directly into the cache.

Currently there is a rust based LMDB wrapper implementation under the `useLmdbJsLite`. This change is rolling out with good results.

This diff couples the rollout of `atlaspackV3` with `useLmdbJsLite` (using V3 will enable the rust LMDB cache), and demonstrates the set-up so we can share the LMDB environment between rust and JavaScript.

The LMDB wrapper is designed such that all JavaScript writes are centralised in a single writer thread.

We can use this writer thread from Rust, as well as implement an async API through that handle. This PR however is storing the direct LMDB handle into the `Atlaspack` rust struct. Will perform reads/writes on the calling thread instead of going through this writer thread.

## Checklist


crates/lmdb-js-lite/src/writer.rs
```diff
@@ -254,6 +254,19 @@ impl<'a, 'b> Transaction<'a, 'b> {
   }
 }
 
+/// Wraps a LMDB database environment.
+///
+/// This is thread-safe and can be shared across threads. LMDB itself will
+/// manage locks.
+///
+/// It's important that batch writes are done within a single write transaction.
```
yamadapc, line 262: For performance

crates/lmdb-js-lite/src/writer.rs
```diff
@@ -254,6 +254,19 @@ impl<'a, 'b> Transaction<'a, 'b> {
...
+///
+/// Entries are individually compressed on read/write. In the future we may
+/// want to expose batch write methods that run compression in multiple threads.
```
yamadapc, line 265: This is done for the JavaScript batch write API

crates/lmdb-js-lite/src/lib.rs
```diff
@@ -336,12 +353,25 @@ impl LMDB {
...
+  pub fn get_database(&self) -> Result<&Arc<DatabaseHandle>, GetDatabaseError> {
     let inner = self
       .inner
       .as_ref()
-      .ok_or_else(|| napi::Error::from_reason("Trying to use closed DB"))?;
+      .ok_or_else(|| GetDatabaseError::ClosedDatabase)?;
+    Ok(inner)
+  }
+
+  pub fn get_database_napi(&self) -> napi::Result<&Arc<DatabaseHandle>> {
```
MonicaOlejniczak, line 371: Is this extra function really needed? The `get_database` is only used once in this file, and it just maps the napi error anyway. Additionally, can you write these bindings in the node bindings crate?

MonicaOlejniczak: I do prefer the bindings to be separate, but it might get repetitive. It's not a big deal

MonicaOlejniczak (approved):

PR #158, opened 2024-10-28
Spike clippy usage in atlaspack

Nikola-3:
The PR:
* applies `cargo clippy --fix --lib -p <package_name>` for each package with issues when run with `cargo clippy`
* adds manual fixes or exceptions depending on initial complexity of refactoring
* fixes the `caniuse` script
* adds `cargo clippy` as a task to the ci lint job
* all current changes are suggestions that should be discussed within the team before adopting the convention

## Motivation

As we write more rust code, we would like to have a linter with an agreed upon set of rules for best coding practices.

## Changes

Min set of changes needed to get `cargo clippy` and `yarn build-native(-release)` to both work at the same time.

## Checklist


.github/workflows/ci.yml
```diff
@@ -25,6 +25,7 @@ jobs:
           components: rustfmt
       - run: yarn --frozen-lockfile
       - run: yarn lint
+      - run: cargo clippy
```
alshdavid, line 28: You can incorporate `cargo clippy` into `yarn lint` by adding it to the end of the command with `&&`

crates/atlaspack_contextual_imports/Cargo.toml
```diff
@@ -8,7 +8,7 @@ description = "Atlaspack contextual imports transformation"
 [dependencies]
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true}
-swc_core = { workspace = true, features = ["ecma_plugin_transform"] }
+swc_core = { workspace = true, features = ["ecma_plugin_transform", "__utils", "ecma_quote"] }
```
yamadapc, line 11: We shouldn't need these features ; I reckon `__...` is private

yamadapc (approved):
lgtm but will cause conflicts with everything ; https://github.com/atlassian-labs/atlaspack/pull/161 is also a large diff

I reckon you've ran auto-fix for these changes right? So we can prioritise other branches to merge first?

JakeLane (approved):

crates/atlaspack_contextual_imports/Cargo.toml
Nikola-3: In each of these cases, I went into the imported module and found that it was being gated behind a feature I added. For example this import: https://github.com/atlassian-labs/atlaspack/blob/bd0c80bc7f1b09dcaa37a01ae4bbee5d16ce7d6a/crates/atlaspack_contextual_imports/src/lib.rs#L8
```
use swc_core::ecma::utils::private_ident;
```
resolves to
`.cargo/registry/src/index.crates.io-6f17d22bba15001f/swc_core-0.96.9/src/lib.rs`
```
#[cfg(feature = "__utils")]
    #[cfg_attr(docsrs, doc(cfg(feature = "__utils")))]
    pub mod utils {
        pub use swc_ecma_utils::*;
    }
```

Nikola-3:
Most of the changes are auto-fix, some are manual, almost all of which are the ignores and the addition of features to toml files.

crates/atlaspack_napi_helpers/src/js_callable/js_callable.rs
```diff
@@ -101,7 +101,7 @@ impl JsCallable {
     let send = {
       let name = self.fn_name.clone();
       move |error_message: &str, value: napi::Result<Return>| -> napi::Result<()> {
-        if let Err(_) = tx.send(value) {
+        if tx.send(value).is_err() {
```
Nikola-3, line 104: https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pattern_matching - do we care about the drop order here?

MonicaOlejniczak (approved):

JakeLane (approved):

PR #161, opened 2024-10-28
V3: Tokio

alshdavid:
Move to async rust so that we can coordinate Rust to JavaScritpt calls without blocking worker threads.

- Introduced `tokio`
- Introduced `async-trait`
- Moved `RequestTracker` from using `rayon` to `tokio` for request spawning
- Migrated `atlaspack::RequestTracker::run_request` to async
- Migrated `atlaspack_core::plugin::Resolver` to async
- Migrated `atlaspack_core::plugin::Transformer` to async
- Migrated Resolver and Transformer plugins to async
- Updated `NodeJsRpcTransformer` and `NodejsRpcResolver` to use async calls when communicating with Nodejs
- Updated `NodejsRpcWorkerfarm` load balancer to not block on calls to the workers
- Made `JsCallable::call()` async
- Moved original `JsCallablbe::call()` to `JsCallable::call_blocking()`

- Updated `atlaspack::Atlaspack::state(&mut self) -> State` to return a new state for each call (no mutation of parent struct)
  - Call is now `Atlaspack::state(&self) -> State` 
- Made `RpcHost` mandatory
  - Not essential for this PR but nice to have


crates/atlaspack_napi_helpers/src/js_callable/js_callable.rs
```diff
@@ -13,11 +13,305 @@ use napi::JsObject;
...
+            &name, error_message
+          )));
+        }
+        Ok(())
+      }
+    };
+
+    let threadsafe_function = self.threadsafe_function.clone();
+
+    tokio::spawn(async move {
```
yamadapc, line 116: no need to do this ; the call is non-blocking already

crates/atlaspack_napi_helpers/src/js_callable/js_callable.rs
```diff
@@ -1,4 +1,4 @@
-use std::sync::mpsc::channel;
+use std::sync::mpsc::channel as std_channel;
```
yamadapc, line 1: nit: It'd rather qualify names than rename things.

crates/atlaspack_napi_helpers/src/js_callable/js_callable.rs
```diff
@@ -13,11 +13,305 @@ use napi::JsObject;
...
+  {
+    Ok(
+      self
+        .call(map_params_serde(params), map_return_serde())
+        .await?,
+    )
+  }
+}
+
+/*
```
yamadapc, line 290: Please delete this

crates/atlaspack_napi_helpers/src/js_callable/oneshot.rs
```diff
@@ -0,0 +1,49 @@
+//! This module contains a oneshot channel that can be cloned
```
yamadapc, line 1: why can't we just use oneshot?

crates/atlaspack/src/requests/path_request.rs
```diff
@@ -66,11 +68,13 @@ impl Request for PathRequest {
     let mut invalidations = Vec::new();
 
     for resolver in request_context.plugins().resolvers()?.iter() {
-      let resolved = resolver.resolve(ResolveContext {
-        dependency: Arc::clone(&self.dependency),
-        pipeline: parsed_pipeline.clone(),
-        specifier: String::from(specifier),
-      })?;
+      let resolved = resolver
+        .resolve(ResolveContext {
```
yamadapc, line 72: To get good performance, we then need to take each sync resolver implementation and move it out of the tokio loop.

That will be both for the `transformers` and the `resolvers`;

So where we call `resolver.resolve_with_options` we need to either wrap that with a rayon task on a thread-pool or with `spawn_blocking`

yamadapc (approved):
Looks good to me but can you remove the commented out code

crates/atlaspack/src/requests/path_request.rs
alshdavid: Agreed, there's quite a few blocking calls throughout the code base that we have to address. 

Will do so in following PRs to avoid blowing this one up any further

PR #163, opened 2024-10-28
[V3] Resolver fixes for Jira

mattcompiles:
## Motivation

This PR resolves some resolver specific issues encountered while building Jira.

## Changes

- Run the next resolver when the current one fails
- Update the GlobResolver to handle TS files
  - Not sure why this doesn't fail in V2. Likely due to V3 using different `sourceAssetType` values

## Checklist


crates/atlaspack/src/requests/path_request.rs
```diff
@@ -66,11 +66,21 @@ impl Request for PathRequest {
...
+      let result = resolver.resolve(ResolveContext {
         dependency: Arc::clone(&self.dependency),
         pipeline: parsed_pipeline.clone(),
         specifier: String::from(specifier),
-      })?;
+      });
+
+      let resolved = match result {
+        Ok(result) => result,
+        Err(error) => {
```
yamadapc, line 77: hm.... that's not good, why do we need to eat these errors?

mattcompiles: We need to update our diagnostic type to allow multiple errors/sub-errors.

yamadapc (approved):

PR #165, opened 2024-10-29
Make integration tests run faster

yamadapc:
Don't shutdown worker threads between integration tests / builds.

### DIFF
--- a/packages/core/core/src/atlaspack-v3/AtlaspackV3.js
+++ b/packages/core/core/src/atlaspack-v3/AtlaspackV3.js
@@ -14,25 +14,49 @@ export type AtlaspackV3Options = {|
   ...AtlaspackNapiOptions['options'],
 |};
 
+type WorkerMessage = {|type: 'workerRegistered'|} | {|type: 'pong'|};
+
 class WorkerPool {
   workerPool: Worker[] = [];
   currentUsedWorkers: number = 0;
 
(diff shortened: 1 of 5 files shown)


packages/core/core/src/atlaspack-v3/AtlaspackV3.js
```diff
@@ -14,6 +14,60 @@ export type AtlaspackV3Options = {|
   ...AtlaspackNapiOptions['options'],
 |};
 
+type WorkerMessage = {|type: 'workerRegistered'|} | {|type: 'pong'|};
+
+class WorkerPool {
```
MonicaOlejniczak, line 19: /nit put this in a sibling `WorkerPool.js` file

packages/core/core/src/atlaspack-v3/AtlaspackV3.js
```diff
@@ -14,6 +14,60 @@ export type AtlaspackV3Options = {|
...
+  workerPool: Worker[] = [];
+  currentUsedWorkers: number = 0;
+
+  waitMessage(worker: Worker, type: string): Promise<WorkerMessage> {
```
MonicaOlejniczak, line 23: Consider using `#` for the private methods, otherwise it's difficult to understand the public api at a glance

MonicaOlejniczak (approved):

PR #167, opened 2024-10-29
WIP: V3: Asset Contents via Buffer

alshdavid:
WIP


crates/atlaspack_plugin_rpc/src/nodejs/plugins/nodejs_rpc_transformer.rs
```diff
@@ -106,19 +111,45 @@ impl TransformerPlugin for NodejsRpcTransformerPlugin {
...
+    let (result, contents) = self
       .nodejs_workers
       .next_worker()
       .transformer_register_fn
-      .call_serde::<_, RpcTransformerResult>(run_transformer_opts)
+      .call(
+        move |env| {
+          let run_transformer_opts = env.to_js_value(&run_transformer_opts)?;
+
+          let bytes = std::mem::take(asset.code.get_mut());
```
yamadapc, line 128: I don't see the point in doing this vs. passing the value. Also I don't think we should clear the code value, it'd be better to copy or pass a reference.

Is this actually a bottle-neck?

alshdavid: The idea here is to move the bytes of the `Asset` contents into JavaScript then move it back when the plugin is done.

There is no need to copy the data for this, so in this case I am simply taking the data from the Rust `Asset` - replacing the `Asset` contents with an empty buffer temporarily while the JavaScript plugin is using the contents.

I think I can pass a mutable reference of the `Asset` contents buffer to JavaScript but that is a more complex control flow and is less safe given it will be sent across threads and napi-rs doesn't give us safety errors for this.

This is fine IMO and this one change improves JFE build times by 11%

alshdavid:
https://github.com/atlassian-labs/atlaspack/pull/195

PR #173, opened 2024-11-01
Chore: .vscode/settings.json

alshdavid:
Added this to the workspace settings for vscode users

```json
{
  // Fix for broken NAPI macros
  "rust-analyzer.procMacro.ignored": { "napi-derive": ["napi"] },
  // Disable type checking for Flow
  "javascript.validate.enable": false,
  // Misc
  "rust-analyzer.cargo.features": "all",
  "[rust]": {
    "editor.defaultFormatter": "rust-lang.rust-analyzer",
    "editor.formatOnSave": true,
    "editor.formatOnType": true,
  }
}

```


marcins:
Doesn't VSCode have a way to provide repo specific settings? Or does that not work well for this kind of thing?

alshdavid:
> Doesn't VSCode have a way to provide repo specific settings? Or does that not work well for this kind of thing?

Yep, this script sets up the workspace settings. 

Haha, before you can configure your workspace settings you have to first clone the repo - but then you have to look up what settings to set and it's a bit of a nuisance to look up.

Alternatively, I can add these settings to the docs and vscode users can add them manually

MonicaOlejniczak:
@alshdavid I would just commit `.vscode/settings.json` and remove it from the `.gitignore`, instead of writing a script. If someone has custom settings already, they should consider merging it for everyone or apply it as a user setting.

alshdavid:
> @alshdavid I would just commit `.vscode/settings.json` and remove it from the `.gitignore`, instead of writing a script. If someone has custom settings already they should consider merging it in for everyone, or apply it as a user setting (which is a file outside of the repo).

Sounds good to me, updated

MonicaOlejniczak (approved):

PR #174, opened 2024-11-01
V3: Support for TypeScript plugins

alshdavid:
- Added `esbuild-register` to workers

Note: Plugins written in TypeScript that are `type: module` will not work until `@atlaspack/core` is migrated to either TypeScript or `type: module`.

I looked at converting just the worker code to be esm - but that's not viable due to the Flow. I guess we don't have any esm plugins anyway so it's probably fine for now.


JakeLane (approved):

alshdavid:
Moved this to the consumer site

PR #175, opened 2024-11-04
add async/await to runtime plugins

Nikola-3:
This is the first in a series of PRs that move over atlaspack rust plugins to async/await syntax to make them run asynchronously. The PRs are separated by plugin type.


crates/atlaspack_core/Cargo.toml
```diff
@@ -28,3 +28,4 @@ serde-value = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 xxhash-rust = { workspace = true, features = ["xxh3"] }
+tokio = "1.41.0"
```
alshdavid, line 31: Try this to pull the version specified in the root `Cargo.toml`
```toml
tokio = { workspace = true }
```

alshdavid: Actually, `tokio` isn't used in this crate so this can be removed

Nikola-3: Oh right that was a miss from the test auto complete 👍

crates/atlaspack_core/src/plugin/runtime_plugin.rs
```diff
@@ -21,9 +22,10 @@ pub struct RuntimeAsset {
 }
 
 /// Programmatically insert assets into bundles
+#[async_trait]
 pub trait RuntimePlugin: Debug {
```
alshdavid, line 26: In Rust, a type needs to be marked as "safe" to be sent across threads. to do this you need to add the `Send` and sometimes `Sync` type bounds.

The reasons async types are not constrained as thread safe by default is Rust allows developers to spawn Futures on the current thread (does not need to be thread safe) or multiple threads (needs to be thread safe).

```rust
pub trait RuntimePlugin: Debug + Send + Sync {}
```

Nikola-3: Done

PR #177, opened 2024-11-04
Added three-js benchmark

alshdavid:
Added a three-js benchmark. Will integrate this into CI in a follow up PR

Usage:
```
yarn build-native-release
yarn build

cd benchmarks/three-js

# V2 with no plugins, 30 copies of 3js
yarn start:v2 

# V3 with no plugins (native only)
yarn start:v3

# V2 with 10 JS plugins
yarn start:v2:plugins

# V3 with 10 JS plugins
yarn start:v3:plugins
```

Works by

- Creating a temporary directory
- Copying benchmark to directory
- Creating 30 copies of 3js (can increase or decrease with `ATLASPACK_BENCH_COPIES=50` env var
- Links Atlaspack to local copy
- Running build on generated sources
- Deleting temporary directory

The benchmark cannot be run in place because Atlaspack uses the root `package.json` and fails to build


benchmarks/three-js/vendor/three-js.tar.gz
alshdavid, line 1: I vendored three-js into the repo as a tar file because it saves the time needed to clone a specific commit of three-js.

This is the same version of 3js that is used by esbuild for their benchmarks

PR #180, opened 2024-11-04
Clean-up cache optimization feature-flag

yamadapc:
This change, added in https://github.com/atlassian-labs/atlaspack/pull/105, has been rolled-out.

It improves cache invalidation time by several orders of magnitude on the worst case, after changing branches or reinstalling dependencies.


yamadapc:
There's something wrong with the change

PR #185, opened 2024-11-04
Fix dev dep request missing 'invalidateOnCreate'

yamadapc:
This commit fixes a long-standing bug in the dev dep request where the 'invalidateOnCreate' property was missing.

The reason this happened was due to an optimisation in the worker to main-thread communication, which ommitted certain fields from the DevDepRequest type when the values were thought to be unnecessary.

This commit fixes the issue by:

* Ensuring the value exists at a type-level more clearly

A test that reproduces the bug is added.

## Checklist


packages/core/core/src/Atlaspack.js
```diff
@@ -563,14 +564,24 @@ export default class Atlaspack {
...
   async unstable_invalidate(): Promise<void> {
     await this._init();
   }
 
   /**
    * Build the asset graph
    */
-  async unstable_buildAssetGraph(): Promise<void> {
+  async unstable_buildAssetGraph(
+    writeToCache: boolean = true,
```
MonicaOlejniczak, line 583: Should the cli be updated to support this?

MonicaOlejniczak (approved):

PR #195, opened 2024-11-05
V3: Transferring NodejsAsset contents via buffer

alshdavid:
- Migrated `Asset.code` from `Arc<Code>` to just `Code`
- Copying bytes from rust into JavaScript buffer
- Transferring buffer from Rust to JavaScript
- Changed nodejs `worker.js#runTransformerTranform()` to accept the buffer as the contents

**Benchmarks**
16 cores 16 workers
30 copies 3js
10 javascript plugins

|branch|time|%|
|-|-|-|
|`main:v2`|7.49s|~|
|`main:v3`|19.53s|~|
|`alsh/transformer-zero-copy-buffer`|15.40s|-21%|


crates/atlaspack_napi_helpers/src/zero_copy_buffer.rs
```diff
@@ -0,0 +1,43 @@
+use napi::*;
+
+pub struct ZeroCopyBuffer(JsBuffer);
```
yamadapc, line 3: why can't you just use `Buffer`?

Cargo.toml
```diff
@@ -7,6 +7,9 @@ members = [
   "packages/utils/dev-dep-resolver"
 ]
 
+[profile.release]
+debug = true
```
yamadapc, line 11: don't want to add this

packages/core/core/src/atlaspack-v3/worker/worker.js
```diff
@@ -159,122 +159,135 @@ export class AtlaspackWorker {
...
+            pipeline: mutableAsset.pipeline,
+            query: mutableAsset.query.toString(),
+            symbols: mutableAsset.symbols.intoNapi(),
+            uniqueKey: mutableAsset.uniqueKey,
+            sideEffects: mutableAsset.sideEffects,
+            isBundleSplittable: mutableAsset.isBundleSplittable,
+            isSource: mutableAsset.isSource,
+          },
+        },
+        await mutableAsset.getBuffer(),
```
MonicaOlejniczak, line 287: This structure doesn't make sense to me, because multiple assets can be returned. Can't you just add a `code` property alongside `asset`?

crates/atlaspack_napi_helpers/src/zero_copy_buffer.rs
alshdavid: Can use `napi::bindgen_prelude::Buffer` but it needs unsafe Rust to coerce it to a usable value.

As an alternative, napi-rs has `JsBuffer` and `JsBufferValue` which you create from `env` which is safe - but you need to jump between those values to use them making them unergonomic to use (cannot be used (at least not ergonomically) in a `napi(object)` or as an argument to `#[napi]` functions). 

`ZeroCopyBuffer` is a wrapper on top of those types to make them more ergonomic to use. I can get rid of this abstraction if you'd like but it means more code at the call site to coerce the types to the desired types.

packages/core/core/src/atlaspack-v3/worker/worker.js
alshdavid: That's fair, just updated this to the following shape:
```typescript
type RpcAssetResult = [
  RpcAssetResult,
  Buffer,
]
```
Where multiple assets would be:
```typescript
return [
  [ asset1, contents1 ],
  [ asset2, contents2 ],
  // etc
]
```
This is because I am trying to get around the fact that we use serde to deserialize values returned from JavaScript.

I'm using a tuple here because it's cheaper to operate on from napi (numeric keys on objects/arrays are faster to access than string keys)

packages/core/core/src/atlaspack-v3/worker/worker.js
```diff
@@ -355,6 +366,4 @@ type RunTransformerTransformOptions = {|
   asset: napi.Asset,
 |};
 
-type RunTransformerTransformResult = {|
-  asset: napi.RpcAssetResult,
-|};
+type RunTransformerTransformResult = [napi.RpcAssetResult, Buffer];
```
MonicaOlejniczak, line 369: Shouldn't this be an array now?

alshdavid: At the moment Rust the nodejs transformer doesn't support returning multiple assets so JavaScript only returns one `Asset` back to rust.

One asset is typed as `[serializableAsset, contentsBuffer]`

Will update the transformer soon to enable returning multiple assets

MonicaOlejniczak (approved):

crates/atlaspack_napi_helpers/src/zero_copy_buffer.rs
alshdavid: Basically:
```rust
#[napi]
fn foo(env: Env, buf: JsBuffer) -> Result<JsBuffer> {
  // Buffer -> Vec<u8>
  let buf = buf.into_value()?;
  let buf = buf.to_vec();
  
  // Vec<u8> -> Buffer
  let buf = env.create_buffer_with_data(buf)?;
  let buf = buf.into_raw();
  
  Ok(buf)
}
```

vs

```rust
#[napi]
fn foo(env: Env, buf: ZeroCopyBuffer) -> Result<ZeroCopyBuffer> {
  // Buffer -> Vec<u8>
  let buf = buf.to_vec()?;
  
  // Vec<u8> -> Buffer
  let buf = ZeroCopyBuffer::new(&env, buf)?;
  
  Ok(buf)
}
```

yamadapc:
can we use `get_element` instead of unchecked version

and just use `Buffer` directly?

``` 
    let mut result = env.create_buffer(buffer.len())?;
    // This is faster than moving the vector in
    result.copy_from_slice(&buffer);
    result.into_unknown()
```

alshdavid:
> can we use `get_element` instead of unchecked version

Updated. We don't really need runtime type checking because we control both sides of the integration, but I'm not fussed.

> and just use `Buffer` directly?

Unfortunately `JsCallable` doesn't take a `Vec<impl NapiRaw>` because it's tricky to use traits there so instead it takes an `Vec<JsUnknown>` which most napi types can be converted into.

I can look at addressing that in a future PR

yamadapc (approved):

PR #196, opened 2024-11-05
V3: Use lmdb to remove code serialisation

MonicaOlejniczak:
## Motivation

Currently we serialise the asset code back through napi, while instead we can use lmdb to avoid this process. It has shown an approximate 5-10% improvement when building an small product entry.

## Changes

* Add rust flags to ensure unit tests work for macos, and are the same as the existing linux flags
* Store asset code in lmdb at the end of the request, as the performance was not much different to doing this within the asset request and is less complex
* Skip serialising asset code, this is primarily used when the asset graph is returned
* Fix lmdb-js-lite `get_sync` method so it does not fail on empty files
* Add recovery to lmdb-js-lite that clears the cache when encountering an invalid folder
* Modify lmdb-js-lite `getBlob` to use async for better stability when errors are thrown
* Add serialize / deserialize types for easier use
* Remove cache write in rust asset graph request
* Always use lmdb in tests / runtime for v3

## Checklist


packages/core/cache/src/LMDBLiteCache.js
```diff
@@ -139,12 +143,9 @@ export class LMDBLiteCache implements Cache {
...
 
-  getBlob(key: string): Promise<Buffer> {
-    try {
-      return Promise.resolve(this.getBlobSync(key));
-    } catch (err) {
-      return Promise.reject(err);
-    }
+  // eslint-disable-next-line require-await
+  async getBlob(key: string): Promise<Buffer> {
+    return this.getBlobSync(key);
```
yamadapc, line 148: This change is not right ; we need to wrap it in a promise

crates/lmdb-js-lite/src/writer.rs
```diff
@@ -286,26 +287,44 @@ impl DatabaseWriter {
...
-      flags.set(EnvFlags::NO_META_SYNC, options.async_writes);
-      let mut env_open_options = EnvOpenOptions::new();
       env_open_options.flags(flags);
-      // http://www.lmdb.tech/doc/group__mdb.html#gaa2506ec8dab3d969b0e609cd82e619e5
-      // max DB size that will be memory mapped
-      if let Some(map_size) = options.map_size {
-        env_open_options.map_size(map_size as usize);
+
+      let mut env = env_open_options.open(path);
+      if path.ends_with(".atlaspack-cache") || path.ends_with(".parcel-cache") {
```
yamadapc, line 312: why is it conditional on the path? this isn't a bundler specific crate

crates/lmdb-js-lite/src/writer.rs
```diff
@@ -286,26 +287,44 @@ impl DatabaseWriter {
...
+        if let Err(heed::Error::Mdb(MdbError::Invalid)) = env {
+          // Remove invalid v2 caches and retry opening the database
+          tracing::debug!("Clearing incompatible cache {}", path.display());
```
yamadapc, line 315: let's warn instead of debug

crates/atlaspack/src/atlaspack.rs
```diff
@@ -130,7 +135,20 @@ impl Atlaspack {
       let request_result = request_tracker.run_request(AssetGraphRequest {}).await?;
 
       let asset_graph = match request_result {
-        RequestResult::AssetGraph(result) => result.graph,
+        RequestResult::AssetGraph(result) => {
+          let mut txn = self.db.environment().write_txn().unwrap();
```
yamadapc, line 139: Let's write tests and put this into methods

packages/core/core/src/atlaspack-v3/worker/worker.js
```diff
@@ -364,6 +366,7 @@ type RunTransformerTransformOptions = {|
   env: napi.Environment,
   options: RpcPluginOptions,
   asset: napi.Asset,
+  code: String,
```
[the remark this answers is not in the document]
MonicaOlejniczak: Yep, it's from a bad rebase. Will remove it

packages/core/core/src/atlaspack-v3/AtlaspackV3.js
```diff
@@ -12,10 +12,7 @@ export type AtlaspackV3Options = {|
   nodeWorkers?: number,
   packageManager?: AtlaspackNapiOptions['packageManager'],
   threads?: number,
-  /**
-   * A reference to LMDB lite's rust object
-   */
-  lmdb: Lmdb | null,
+  lmdb: Lmdb,
```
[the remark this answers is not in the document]
MonicaOlejniczak: I changed it to a different type and the comment was not relevant at the time, but later reverted the change. Will reinstate it.

packages/core/core/src/PackagerRunner.js
```diff
@@ -446,21 +448,25 @@ export default class PackagerRunner {
...
-      );
-      this.devDepRequests.set(
-        `${name}:${fromProjectPathRelative(resolveFrom)}`,
-        devDepRequest,
-      );
     }
+
+    // Add dev dependency for the packager. This must be done AFTER running it due to
+    // the potential for lazy require() that aren't executed until the request runs.
+    let devDepRequest = await createDevDependency(
```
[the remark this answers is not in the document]
MonicaOlejniczak: It's in the description. For some reason when lmdb fails within here, the process hangs indefinitely. I believe it might be behaviour that's part of the spec. Regardless, I tested the behaviour wrt to the comment and the ordering is the same within the request.

packages/core/core/src/Atlaspack.js
```diff
@@ -137,14 +137,12 @@ export default class Atlaspack {
       // eslint-disable-next-line no-unused-vars
       let {entries, inputFS, outputFS, ...options} = this.#initialOptions;
 
-      const lmdb =
-        resolvedOptions.cache instanceof LMDBLiteCache
-          ? resolvedOptions.cache.getNativeRef()
-          : null;
+      // $FlowFixMe
+      const lmdb: Lmdb = (resolvedOptions.cache: any).getNativeRef();
```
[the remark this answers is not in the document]
MonicaOlejniczak: Sure

packages/core/cache/src/LMDBLiteCache.js
MonicaOlejniczak: It's wrapped via `async`

crates/lmdb-js-lite/src/writer.rs
MonicaOlejniczak: The intention was to make sure it's the right cache as I'm unsure whether the error can be triggered in other scenarios. I can remove it, not a big deal.

crates/lmdb-js-lite/src/writer.rs
MonicaOlejniczak: Sure

packages/core/core/src/PackagerRunner.js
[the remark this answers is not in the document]
yamadapc: > Move package runner finally block outside of it, as it makes no runtime difference and enables lmdb to not hang when errors occur within here

I don't understand the reason it has been moved though. Why did this cause hangs?

- - -

The main problem is that there's a behaviour difference here, the dev dependency was previously created and registered in `this.devDepRequests` even when exceptions were thrown.

Now it'll only be registered if there is no exception.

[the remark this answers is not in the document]
MonicaOlejniczak: It was difficult to understand, but it seemed like an issue with how node processes errors then returns results within a finally block whenever lmdb keys were missing. I can try revisiting to see if there's something else to be done, but don't think there will be.

Won't the entire build crash when an exception is thrown here regardless? My understanding is that the dev dep request wouldn't actually do anything when triggered during an error, but correct me if I'm wrong.

packages/core/cache/src/LMDBLiteCache.js
yamadapc: hm.... okay

packages/core/core/src/PackagerRunner.js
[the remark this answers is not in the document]
yamadapc: This is changing state though ; I don't know whether that is safe, but I don't think we should be making that change.

- - -

I don't understand why we're changing it though, there's no LMDB on this code, what problem did this cause?

crates/atlaspack/src/atlaspack.rs
MonicaOlejniczak: Are you after an entire integration test, since that's essentially what the JS one's already demonstrate (they would fail if this code block did not work).

Or are you wanting unit tests for the commit operation?

alshdavid:
*Benchmarks*
Three-js 
30 copies
10 plugins
16 physical cores, 16 threads, 16 workers

```
V2:     7835ms
V3:     15871ms
Branch: 7735ms
```

crates/atlaspack/src/atlaspack.rs
yamadapc: Everything should have both integration and exhaustive unit tests.

packages/core/core/src/PackagerRunner.js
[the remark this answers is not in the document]
MonicaOlejniczak: We paired on this, and discovered the issue with the `PromiseQueue` and unhandled rejections that was causing workers to be recreated indefinitely. @yamadapc will be opening a PR to fix the underlying issue.

JakeLane (approved):

yamadapc (approved):

PR #197, opened 2024-11-06
´Fix PromiseQueue to remove unhandled rejections

yamadapc:
The current `PromiseQueue` API is build such that the following code will raise an unhandled rejection error:

```js
const queue = new PromiseQueue();
queue.add(() => Promise.reject(new Error('foo')));
await queue.run();
```

For this reason, most of the codebase uses `PromiseQueue#add` with a catch block around the promise, including its own tests.

As such, the correct usage of the API is:
```js
const queue = new PromiseQueue();
queue.add(() => Promise.reject(new Error('foo'))).catch((err) => {});
await queue.run();
```

All the other usages, that haven't been modified by this commit, were missing this handling and may lead to unhandled rejections crashing threads/workers.

This commit changes the API so that this can't happen, the API can't trigger unhandled exceptions anymore, and the promise is not returned.

The only case that did require a promise to be returned was the `AssetGraphRequest` which is updated to use the new API.

### DIFF
--- a/packages/core/utils/test/PromiseQueue.test.js
+++ b/packages/core/utils/test/PromiseQueue.test.js
@@ -19,11 +19,24 @@ describe('PromiseQueue', () => {
     assert(someBooleanToBeChanged);
   });
 
-  it('run() should reject if the function throws a sync exception', async () => {
+  it('run() should reject if any of the async functions in the queue failed', async () => {
     let error = new Error('some failure');
     try {
       let queue = new PromiseQueue();
       queue.add(() => Promise.reject(error));
+      await queue.run();
+    } catch (e) {
+      assert.equal(e, error);
+    }
+  });
+
+  it('run() should reject if any of the async functions in the queue throwo', async () => {
+    let error = new Error('some failure');
+    try {
+      let queue = new PromiseQueue();
+      queue.add(() => {
+        throw error;
+      });
       await queue.run();
     } catch (e) {
       assert.equal(e, error);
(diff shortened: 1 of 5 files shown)


packages/core/utils/test/PromiseQueue.test.js
```diff
@@ -42,18 +38,9 @@ describe('PromiseQueue', () => {
 
   it(".add() should resolve with the same result when the passed in function's promise resolves", async () => {
     let queue = new PromiseQueue();
-    let promise = queue.add(() => Promise.resolve(42));
-    await queue.run();
-    let result = await promise;
-    assert.equal(result, 42);
-  });
-
-  it(".add() should reject with the same error when the passed in function's promise rejects", async () => {
```
yamadapc, line 51: it shouldn't

packages/core/utils/test/PromiseQueue.test.js
```diff
@@ -42,18 +38,9 @@ describe('PromiseQueue', () => {
 
   it(".add() should resolve with the same result when the passed in function's promise resolves", async () => {
```
yamadapc, line 39: it shouldn't

packages/core/utils/test/PromiseQueue.test.js
```diff
@@ -19,15 +19,11 @@ describe('PromiseQueue', () => {
...
   });
 
-  it('run() should reject if any of the async functions in the queue failed', async () => {
+  it('run() should reject if the function throws a sync exception', async () => {
     let error = new Error('some failure');
     try {
       let queue = new PromiseQueue();
-      queue
-        .add(() => Promise.reject(error))
-        .catch(
```
yamadapc, line 28: not needed anymore

JakeLane (approved):

MonicaOlejniczak (approved):

PR #200, opened 2024-11-07
Rust: Using smart pointers in resolver to avoid cloning tsconfig

alshdavid:
- Using `Arc<RwLock>>` for properties in `TsConfig` and `TsConfigWrapper` to avoid cloning the underlying data

### DIFF
--- a/packages/utils/node-resolver-rs/src/lib.rs
+++ b/packages/utils/node-resolver-rs/src/lib.rs
@@ -24,7 +24,6 @@ use package_json::PackageJson;
 pub use package_json::PackageJsonError;
 use parking_lot::RwLock;
 use parking_lot::RwLockReadGuard;
-use parking_lot::RwLockWriteGuard;
 pub use specifier::parse_package_specifier;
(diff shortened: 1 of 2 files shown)


packages/utils/node-resolver-rs/src/lib.rs
```diff
@@ -1148,10 +1161,10 @@ impl<'a> ResolveRequest<'a> {
 
               if !exists {
                 return Err(ResolverError::TsConfigExtendsNotFound {
-                  tsconfig: tsconfig.compiler_options.path.clone(),
+                  tsconfig: compiler_options.path.clone(),
                   error: Box::new(ResolverError::FileNotFound {
                     relative: path.to_path_buf(),
-                    from: tsconfig.compiler_options.path.clone(),
+                    from: compiler_options.path.clone(),
```
alshdavid, line 1167: Leaving these clones in, unsure if these will be a problem (likely not)

yamadapc (approved):

PR #204, opened 2024-11-11
V3: Simplify asset and dependency processing

MonicaOlejniczak:
## Motivation

Once the rust based asset graph is returned, adding all the nodes to the JavaScript based asset graph is quite slow. These changes reduce the amount of fields that are changed to what is necessary. Both the three-js benchmarks and an internal product build show a ~45-65% improvement on the `getAssetGraph` function with these changes.

| Benchmark  | Before | After | Improvement
| - | - | - | - |
| internal | 11.40s | 4.09s | -7.31s (64.14%) |
| three-js | 1.28s  | 0.68s  | -0.6s (46.88%) |

## Changes

* Mutate asset and dependency values
* Only change what is needed

## Checklist

### DIFF
--- a/packages/core/core/src/requests/AssetGraphRequestRust.js
+++ b/packages/core/core/src/requests/AssetGraphRequestRust.js
@@ -9,7 +9,6 @@ import type {Async} from '@atlaspack/types';
 import AssetGraph, {nodeFromAssetGroup} from '../AssetGraph';
 import type {AtlaspackV3} from '../atlaspack-v3';
 import {ATLASPACK_VERSION} from '../constants';
-import {toProjectPath} from '../projectPath';
 import {requestTypes, type StaticRunOpts} from '../RequestTracker';
 import {propagateSymbols} from '../SymbolPropagation';
 import type {Environment} from '../types';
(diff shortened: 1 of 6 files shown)


packages/core/core/src/requests/AssetGraphRequestRust.js
```diff
@@ -218,42 +194,16 @@ function getAssetGraph(serializedGraph, options) {
...
-          id: getEnvId(dependency.env),
-          sourceType: dependency.env.sourceType,
-          loc: dependency.env.loc ?? undefined,
-        },
-        pipeline: dependency.pipeline ?? undefined,
-        range: dependency.range ?? undefined,
-        resolveFrom: dependency.resolveFrom ?? undefined,
-        target: dependency.target ?? undefined,
-        bundleBehavior:
-          dependency.bundleBehavior === 255 ? null : dependency.bundleBehavior,
```
yamadapc, line 235: not all fields are compatible

MonicaOlejniczak: The `255` value got removed a while ago when `bundleBehavior` became an optional type, I think it's ok

packages/core/core/src/requests/AssetGraphRequestRust.js
```diff
@@ -164,36 +163,13 @@ function getAssetGraph(serializedGraph, options) {
...
-          ...asset.env,
-          loc: asset.env.loc ?? undefined,
-          id: getEnvId(asset.env),
-          sourceType: asset.env.sourceType,
-        },
-        bundleBehavior:
-          asset.bundleBehavior === 255 ? null : asset.bundleBehavior,
-        committed: true,
-        contentKey: id,
-        filePath: toProjectPath(options.projectRoot, asset.filePath),
```
mattcompiles, line 193: Are the file paths already relative?

MonicaOlejniczak: Currently they are absolute. The internal rust code will need a bit of work to handle working with relative paths.

mattcompiles: Its doesn't break anything making these all absolute?

MonicaOlejniczak: Not that I know of, it's meant to be more of an optimisation for later phases but shouldn't be necessary to use functionally speaking

MonicaOlejniczak: I added some setup code for relative-ising the asset file path, but it's not trivial to do the dependency source path unless we move more of the request tracker to async / await.

JakeLane (approved):

yamadapc (approved):

PR #206, opened 2024-11-13
Add manifest path support to conditional reporter

JakeLane:
## Motivation

Add a configuration option to `package.json` so we can configure the path for the manifest. This is important as jira needs a separate directory for non-hashed file paths.

## Changes

- Add filename option to package.json
- Added new eslint package based on eslint-plugin-monorepo which fixes the auto-fixer behaviour

## Checklist

### DIFF
--- a/packages/dev/eslint-plugin/src/rules/no-relative-import.js
+++ b/packages/dev/eslint-plugin/src/rules/no-relative-import.js
@@ -1,0 +1,93 @@
+// Forked from https://github.com/azz/eslint-plugin-monorepo/blob/master/src/rules/no-relative-import.js
+
+/**
+ * MIT License
+
+  Copyright (c) 2017 Lucas Azzola
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy
+  of this software and associated documentation files (the "Software"), to deal
+  in the Software without restriction, including without limitation the rights
+  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+  copies of the Software, and to permit persons to whom the Software is
+  furnished to do so, subject to the following conditions:
+
+  The above copyright notice and this permission notice shall be included in all
+  copies or substantial portions of the Software.
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+  SOFTWARE.
+ */
+
+const moduleVisitor = require('eslint-module-utils/moduleVisitor');
+
+const getPackages = require('get-monorepo-packages');
+const isInside = require('path-is-inside');
+const minimatch = require('minimatch');
+const _path = require('path');
+const resolve = require('eslint-module-utils/resolve');
+
+const getPackageDir = (filePath, packages) => {
+  const match = packages.find((pkg) =>
+    minimatch(filePath, _path.join(pkg.location, '**')),
+  );
+  if (match) {
+    return match.location;
+  }
+};
+
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        'Disallow usage of relative imports instead of package names in monorepo',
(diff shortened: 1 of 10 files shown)


packages/dev/eslint-plugin/src/rules/no-relative-import.js
```diff
@@ -0,0 +1,93 @@
...
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+  SOFTWARE.
+ */
+
+const moduleVisitor = require('eslint-module-utils/moduleVisitor');
+
+const getPackages = require('get-monorepo-packages');
+const isInside = require('path-is-inside');
+const minimatch = require('minimatch');
+const _path = require('path');
```
MonicaOlejniczak, line 32: Why is it underscored?

packages/dev/eslint-plugin/src/rules/no-relative-import.js
```diff
@@ -0,0 +1,93 @@
...
+      const subPackagePath = _path.relative(pkg.location, resolvedPath);
+      context.report({
+        node,
+        message: `Import for monorepo package '${pkg.package.name}' should be absolute.`,
+        fix: (fixer) => {
+          const basePath =
+            '' +
+            pkg.package.name +
+            (subPackagePath !== '.' ? '/' + subPackagePath : '');
+          const path = _path.parse(basePath).dir;
```
MonicaOlejniczak, line 81: Just parse once

packages/dev/eslint-plugin/src/rules/no-relative-import.js
```diff
@@ -0,0 +1,93 @@
...
+        return;
+      }
+
+      const subPackagePath = _path.relative(pkg.location, resolvedPath);
+      context.report({
+        node,
+        message: `Import for monorepo package '${pkg.package.name}' should be absolute.`,
+        fix: (fixer) => {
+          const basePath =
+            '' +
```
MonicaOlejniczak, line 78: Just remove it?

packages/dev/eslint-plugin/src/rules/no-relative-import.js
```diff
@@ -0,0 +1,93 @@
...
+            pkg.package.name +
+            (subPackagePath !== '.' ? '/' + subPackagePath : '');
+          const path = _path.parse(basePath).dir;
+          const name = _path.parse(basePath).name;
+          return fixer.replaceText(
+            node,
+            "'" +
+              (name !== '.' && name !== 'index' ? path + '/' + name : path) +
```
MonicaOlejniczak, line 86:
```suggestion
              (name !== '.' && name !== 'index' ? `${path}/${name}` : path) +
```

packages/reporters/conditional-manifest/src/ConditionalManifestReporter.js
```diff
@@ -1,62 +1,102 @@
...
-        manifest[bundle.target.name] ??= {};
-        manifest[bundle.target.name][
-          relative(bundle.target.distDir, bundle.filePath)
-        ] = bundleInfo;
+type Config = {|
+  filename: string,
+|};
+
+// Exported for testing
+export async function getConfig({
```
MonicaOlejniczak, line 17: /nit this can just be in its own file, then you don't need the comment and you can just read the reporter logic without worrying about config at a glance

packages/reporters/conditional-manifest/src/ConditionalManifestReporter.js
```diff
@@ -1,62 +1,102 @@
...
+  event,
+  options,
+  logger,
+}: {|
+  event: ReporterEvent,
+  options: PluginOptions,
+  logger: PluginLogger,
+  tracer: PluginTracer,
+|}): Async<void> {
+  const {filename} = await getConfig(options);
```
MonicaOlejniczak, line 51: This should only be retrieved on `buildSuccess`, otherwise it will slow down builds since reporter events are emitted all the time. Therefore, you can just have an early return when it's not a success event

packages/reporters/conditional-manifest/src/ConditionalManifestReporter.js
```diff
@@ -1,62 +1,102 @@
...
+    // Replace bundles with file paths
+    const mapBundles = (bundles) =>
+      bundles.map((bundle) => relative(bundle.target.distDir, bundle.filePath));
+
+    const manifest = {};
+    for (const [bundle, conditions] of bundles.entries()) {
+      const bundleInfo = {};
+      for (const [key, cond] of conditions) {
+        bundleInfo[key] = {
+          ifTrueBundles: mapBundles(cond.ifTrueBundles).reverse(),
```
MonicaOlejniczak, line 65: Can you explain why they're reversed in a comment?

packages/dev/eslint-plugin/src/rules/no-relative-import.js
JakeLane: I believe this was because the original creator wanted to use `path` as a variable 😅 .

I might just refactor the plugin to address the issues you mentioned

MonicaOlejniczak (approved):

PR #213, opened 2024-11-14
Update shared hashmap to use RwLock instead of threadlocal

mattcompiles:
## Motivation

While doing some performance testing I found that the threadlocal version of the resolver cache was have a large negative impact on build times. Reverting back to the RwLock version is almost halving the asset graph build times in native builds. I have also validated that V2 builds are also performing better with this change, however not by the same margin. 

## Changes

- Migrate back to RwLock Hash maps for the resolver cache

## Checklist

### DIFF
--- a/crates/atlaspack_filesystem/src/lib.rs
+++ b/crates/atlaspack_filesystem/src/lib.rs
@@ -2,7 +2,7 @@ use std::path::Path;
 use std::path::PathBuf;
 use std::sync::Arc;
 
-use atlaspack_shared_map::ThreadLocalHashMap;
+use atlaspack_shared_map::SharedHashMap;
 
 /// In-memory file-system for testing
 pub mod in_memory_file_system;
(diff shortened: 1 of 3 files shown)


yamadapc (approved):
This doesn't revert the entire PR that moved to thread-local ; I'd rather revert everything or at least understand where this makes sense vs. not at the moment.

> I have also validated that V2 builds are also performing better with this change

How much better are V2 builds? It's weird because this did not cause a regression when rolled-out on our metrics.

It could be that this strategy works for the caches you didn't revert but not the packages/tsconfigs case or that it works for none & something is wrong with benchmarks/measurements (for example the CI agents we're using have a lot more memory than your computer).

mattcompiles:
> This doesn't revert the entire PR that moved to thread-local ; I'd rather revert everything or at least understand where this makes sense vs. not at the moment.

I did originally go to revert but the revert commit needed changes anyway so I decided to just do this file as it was what I had experimented with. 

> How much better are V2 builds? It's weird because this did not cause a regression when rolled-out on our metrics.

I only did one test just to validate it wasn't a regression for V2 and I built the asset graph about 20s quicker,  4:48 for with thread local vs 4:25 with this change.

yamadapc (approved):

PR #216, opened 2024-11-15
Serialise asset graph as json

MonicaOlejniczak:
## Motivation

Serialising the asset graph from Rust to JavaScript can take a significant amount of time for large asset graphs. The current version of Napi is not designed to transport large data structures all at once, which has been alluded to in napi-rs issues in the past.

These changes improve the total time it takes to build the asset graph by serialising with json and parsing the result on the other end, as seen in the following table:

| Benchmark  | Before | After | Improvement
| - | - | - | - |
| internal | 265s | 156s | -109s (41%) |
| three-js | 6.29s  | 4.69s  | -1.6s (25%) |

Whilst serialising to json is not ideal, it can be removed once more of the core has been ported to Rust.

## Changes

* Improve debug asset graph command logs by specifying an origin and the final build time
* Add `edges` and `nodes` functions to the asset graph
* Add `serialize_nodes` to the asset graph, which returns a json structure partitioned so that it never exceeds the max string length in JavaScript. This ensures `Invalid String Length` errors do not occur when parsing the structure back in the `AssetGraphRequestRust` module.
* Update the serialisation to use `serialize_nodes` in addition to default serialisation for edges (this is not expensive)
* Remove old serialisation of the asset graph

## Checklist

### DIFF
--- a/packages/core/core/src/requests/AssetGraphRequestRust.js
+++ b/packages/core/core/src/requests/AssetGraphRequestRust.js
@@ -42,6 +42,9 @@ export function createAssetGraphRequestRust(
       let serializedAssetGraph;
       try {
         serializedAssetGraph = await rustAtlaspack.buildAssetGraph();
+        serializedAssetGraph.nodes = serializedAssetGraph.nodes.flatMap(
+          (node) => JSON.parse(node),
+        );
       } catch (err) {
         throw new ThrowableDiagnostic({
           diagnostic: err,
--- a/crates/atlaspack_core/src/asset_graph.rs
+++ b/crates/atlaspack_core/src/asset_graph.rs
@@ -5,6 +5,7 @@ use petgraph::{
   visit::EdgeRef,
   Direction,
 };
+use serde::Serialize;
 
 use crate::types::{Asset, Dependency};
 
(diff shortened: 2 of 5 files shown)


alshdavid (approved):

packages/core/core/src/requests/AssetGraphRequestRust.js
```diff
@@ -42,6 +42,9 @@ export function createAssetGraphRequestRust(
       let serializedAssetGraph;
       try {
         serializedAssetGraph = await rustAtlaspack.buildAssetGraph();
+        serializedAssetGraph.nodes = serializedAssetGraph.nodes.flatMap(
```
JakeLane, line 45: might be micro optimising but is it worth considering a conventional loop so a new array doesn't need to be allocated?

JakeLane (approved):

packages/core/core/src/requests/AssetGraphRequestRust.js
MonicaOlejniczak: There's only a few entries in the top-level array at most, so it shouldn't make a noticeable difference.

mattcompiles (approved):

PR #219, opened 2024-11-18
stitch dependencies to assets post build

Nikola-3:
Some functionality available to plugins depends on having dependencies directly available on each asset. This PR intends to make them available, after the rust build to avoid duplicating info for serialisation.

### DIFF
--- a/packages/core/core/src/requests/AssetGraphRequestRust.js
+++ b/packages/core/core/src/requests/AssetGraphRequestRust.js
@@ -11,13 +11,14 @@ import type {AtlaspackV3} from '../atlaspack-v3';
 import {ATLASPACK_VERSION} from '../constants';
 import {requestTypes, type StaticRunOpts} from '../RequestTracker';
 import {propagateSymbols} from '../SymbolPropagation';
-import type {Environment} from '../types';
+import type {Environment, Dependency, Asset} from '../types';
 
 import type {
   AssetGraphRequestInput,
   AssetGraphRequestResult,
 } from './AssetGraphRequest';
 import SourceMap from '@parcel/source-map';
+import type {ContentKey} from '@atlaspack/types';
 
 type RunInput = {|
   input: AssetGraphRequestInput,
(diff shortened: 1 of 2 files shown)


packages/core/core/src/requests/AssetGraphRequestRust.js
```diff
@@ -86,6 +87,45 @@ export function createAssetGraphRequestRust(
...
+    ContentKey,
+    | {type: 'dependency', value: Set<Dependency>}
+    | {type: 'asset', value: Asset},
+  >,
+) {
+  const seenResult = seenAssetMap.get(id);
+
+  if (entity.type === 'asset') {
+    const asset = entity.value;
+    asset.dependencies = asset.dependencies ?? new Map();
```
MonicaOlejniczak, line 105: Just set this with the other setters, after `asset.contentKey = id;`

packages/core/core/src/requests/AssetGraphRequestRust.js
```diff
@@ -86,6 +87,45 @@ export function createAssetGraphRequestRust(
   });
 }
 
+export function trackDependenciesPerAsset(
```
MonicaOlejniczak, line 90: I find this code pretty convoluted, might help to split up the logic with separate maps as well

packages/core/core/src/requests/AssetGraphRequestRust.js
```diff
@@ -86,6 +87,45 @@ export function createAssetGraphRequestRust(
...
+        seenResult?.type === 'dependency' ? seenResult.value : new Set();
+      existingDeps.add(dependency);
+      seenAssetMap.set(id, {type: 'dependency', value: existingDeps});
+    } else {
+      seenResult.value.dependencies.set(dependency.id, dependency);
+    }
+  }
+}
+
 function getAssetGraph(serializedGraph, options) {
```
MonicaOlejniczak, line 129: Did you test the performance difference?

mattcompiles:
Closing in favour of #225

PR #220, opened 2024-11-18
V3: Webpack Magic-Comment Transformer

alshdavid:
- Ported Webpack Magic-Comment Transformer
- Integrated the Webpack Magic-Comment transformer into the JS Transformer
  - Not ideal but we lack the APIs to do this another way, it works so it's okay

Enabled with:
```json
"@atlaspack/transformer-js": {
  "magicComments": true
}
```

### DIFF
--- a/packages/transformers/js/core/src/lib.rs
+++ b/packages/transformers/js/core/src/lib.rs
@@ -5,6 +5,7 @@ mod env_replacer;
 mod fs;
 mod global_replacer;
 mod hoist;
+mod magic_comments;
 mod modules;
 mod node_replacer;
 pub mod test_utils;
@@ -39,6 +40,7 @@ pub use hoist::ExportedSymbol;
 use hoist::HoistResult;
 pub use hoist::ImportedSymbol;
 use indexmap::IndexMap;
+use magic_comments::MagicCommentsVisitor;
 use modules::esm2cjs;
 use node_replacer::NodeReplacer;
 use path_slash::PathExt;
@@ -152,6 +154,7 @@ pub struct TransformResult {
   pub has_node_replacements: bool,
   pub is_constant_module: bool,
   pub conditions: HashSet<Condition>,
+  pub magic_comments: HashMap<String, String>,
 }
 
 fn targets_to_versions(targets: &Option<HashMap<String, String>>) -> Option<Versions> {
@@ -266,6 +269,13 @@ pub fn transform(
 
               let global_mark = Mark::fresh(Mark::root());
               let unresolved_mark = Mark::fresh(Mark::root());
+
+              if code.contains("webpackChunkName") {
+                let mut magic_comment_visitor = MagicCommentsVisitor::new(code);
+                module.visit_with(&mut magic_comment_visitor);
+                result.magic_comments = magic_comment_visitor.magic_comments;
+              }
+
               let module = module.fold_with(&mut chain!(
                 resolver(unresolved_mark, global_mark, config.is_type_script),
                 // Decorators can use type information, so must run before the TypeScript pass.
--- a/packages/transformers/js/core/src/lib.rs
+++ b/packages/transformers/js/core/src/lib.rs
@@ -136,6 +136,7 @@ pub struct Config {
   pub standalone: bool,
   pub inline_constants: bool,
   pub conditional_bundling: bool,
+  pub magic_comments: bool,
 }
 
 #[derive(Serialize, Debug, Default)]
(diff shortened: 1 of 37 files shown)


packages/transformers/js/core/src/magic_comments/visitor.rs
```diff
@@ -0,0 +1,72 @@
...
+
+  fn visit_call_expr(&mut self, node: &CallExpr) {
+    if !node.callee.is_import() || node.args.is_empty() || !node.args[0].expr.is_str() {
+      node.visit_children_with(self);
+      return;
+    }
+
+    let (code_start, code_end) = self.fix_callee(&node.span);
+    let code_start_index = (code_start - 1) as usize;
+    let code_end_index = (code_end - 1) as usize;
```
alshdavid, line 51: swc index starts at 1

packages/transformers/js/core/src/lib.rs
```diff
@@ -266,6 +269,13 @@ pub fn transform(
 
               let global_mark = Mark::fresh(Mark::root());
               let unresolved_mark = Mark::fresh(Mark::root());
+
+              if code.contains("webpackChunkName") {
```
MonicaOlejniczak, line 273: I don't love how webpack is in the code here. Could it be a config option we pass in for magic comment names?

packages/transformers/js/core/src/magic_comments/visitor.rs
```diff
@@ -0,0 +1,90 @@
...
+    };
+
+    let Expr::Lit(Lit::Str(specifier)) = &*expr.expr else {
+      return;
+    };
+
+    // Comments are not available in the AST so we have to get the start/end
+    // positions of the code for the call expression and match it with a regular
+    // expression that matches the magic comment keyword within the code slice
+    let code_start = (node.span.lo().0 - self.offset) as usize;
```
MonicaOlejniczak, line 59: We're meant to parse comments per the config of the transformer, did you try to see how it might be used?

packages/transformers/js/core/src/magic_comments/visitor_test.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+#[test]
+fn it_should_set_chunk_name_if_code_contains_magic_comment() -> anyhow::Result<()> {
+  let code = r#"import(/* webpackChunkName: "foo-chunk" */ "./foo")"#;
+
+  let mut visitor = MagicCommentsVisitor::new(code);
+  parse(code)?.visit_with(&mut visitor);
+
+  assert!(
```
MonicaOlejniczak, line 64: `assert_eq` would be better in this case, so you could differentiate if you had `0` vs `2` in your output

packages/transformers/js/core/src/lib.rs
alshdavid: Updated the code to hide the `webpackChunkName` string behind a `MagicCommentsVisitor::has_magic_comments`  static method - but making the string itself configurable means recompiling the regex per invocation. 

I could add caching for the compiled regex and test cases to cover that - but we don't use alternative names for the `webpackChunkName` keyword so I don't think it's worth implementing at this stage.

packages/transformers/js/core/src/magic_comments/visitor_test.rs
```diff
@@ -0,0 +1,253 @@
...
+    visitor.magic_comments.len() == 1,
+    "Expected magic comment to be set"
+  );
+
+  let Some(chunk_name) = visitor.magic_comments.get("./foo") else {
```
MonicaOlejniczak, line 69: Can't this just be?

```rust
assert_eq!(visitor.magic_comments.get("./foo"), Some("foo-chunk"));
```

alshdavid: The only thing with this is I need the inner value of the `Option` so the code would end up

```rust
assert_eq!(visitor.magic_comments.get("./foo"), Some("foo-chunk"));
let chunk_name = visitor.magic_comments.get("./foo").unwrap();
```

I suppose I could just use `.expect` seeing as `assert` will panic anyway
```rust
let chunk_name = visitor.magic_comments.get("./foo").expect("Expected magic comment to be set");
```

MonicaOlejniczak: You're already returning an anyhow result, so you can use `?` and leave the `.expect` out

packages/transformers/js/core/src/magic_comments/visitor.rs
alshdavid: Added:
```json
"@atlaspack/transformer-js": {
  "magicComments": true
}
```

PR #222, opened 2024-11-19
Add package deduplication feature to the resolver

mattcompiles:
## Motivation

Internally we have a feature in our custom resolver that deduplicates package imports when there are multiple package installs on disk that are copies of the same version. This happens due to yarn only being able to hoist/deduplicate a single version of a package but not multiple. It's worth noting that link based packages like pnpm don't suffer this problem.

This PR implements a resolver feature that deduplicates package imports by doing an optional upfront scan of the `node_modules` folder for any duplicates. On a large project this currently take 2-3 seconds, however it also pre-warms the package cache in the resolver which likely makes back a decent portion of the time. 

To be able to do this, I have had to add the `read_dir` method to the Filesystem trait. It's a little tricky to implement this on the custom file systems so for now it is only implemented in the native one, which is fine for our current use cases. However, it can't yet be used in the integration test suite.

You can enable the feature via a new resolver config.
```json
"@atlaspack/resolver-default": {
  "deduplicatePackages": true
}
```

To achieve acceptable performance, I have implemented the scan and parsing using rayon parallel iterators. I originally tried to use async and tokio tasks but it making this function async was quite painful. We could revisit this later though. Another improvement to explore would be to run this in the background while the rest of the init occurs.

## Other changes

- Added `len` and `is_empty` methods to the  `SharedHashMap`

## Checklist


packages/utils/node-resolver-rs/Cargo.toml
```diff
@@ -13,11 +13,13 @@ atlaspack_core = { path = "../../../crates/atlaspack_core" }
 atlaspack_filesystem = { path = "../../../crates/atlaspack_filesystem" }
 atlaspack_shared_map = { path = "../../../crates/atlaspack_shared_map" }
 
+anyhow = { workspace = true }
 bitflags = "1.3.2"                                            # TODO Update dependency
 glob-match = { workspace = true }
 indexmap = { workspace = true, features = ["serde"] }
 itertools = { workspace = true }
 json_comments = { path = "../../../crates/json-comments-rs" }
+rayon = "1.10.0"
```
alshdavid, line 22: Is it possible to add this to the workspace root and use `workspace = true` for the version?

crates/atlaspack_filesystem/src/in_memory_file_system.rs
```diff
@@ -145,6 +145,10 @@ impl FileSystem for InMemoryFileSystem {
     let file = files.get(&path);
     matches!(file, Some(InMemoryFileSystemEntry::Directory { .. }))
   }
+
+  fn read_dir(&self, _path: &Path) -> std::io::Result<std::fs::ReadDir> {
```
MonicaOlejniczak, line 149: /nit can this be below `read`

crates/atlaspack_shared_map/src/lib.rs
```diff
@@ -51,6 +51,18 @@ where
     let mut map = map_cell;
     map.insert(key, value);
   }
+
+  pub fn len(&self) -> usize {
+    let map_cell = self.inner.read().unwrap();
```
MonicaOlejniczak, line 56: Seems a little inefficient, consider storing the len in the struct and updating it on insertion.

packages/utils/node-resolver-core/.gitignore
```diff
@@ -1,5 +1,6 @@
 !/test/fixture/node_modules
 !/test/fixture/node_modules/.pnpm/*/node_modules
+!/test/fixture/node_modules/duplicate-tester/node_modules
```
MonicaOlejniczak, line 3: Can we just unignore everything under `!/test/fixture/node_modules` so it's a single line?

alshdavid:
Can you add `read_dir` to the `PreloadingFileSystem` in `packages/utils/node-resolver-rs/benches/node_resolver_bench.rs`?

packages/utils/node-resolver-rs/src/lib.rs
```diff
@@ -724,14 +724,16 @@ impl<'a> ResolveRequest<'a> {
         package_path: package.path.clone(),
       })
     } else if !subpath.is_empty() {
-      package_dir.push(subpath);
-      if let Some(res) = self.load_path(&package_dir, Some(&*package))? {
+      let mut package_sub_path = package.path.parent().unwrap().to_owned();
```
MonicaOlejniczak, line 727: You should default instead of unwrapping naively imo

packages/utils/node-resolver-rs/src/cache.rs
```diff
@@ -123,49 +126,81 @@ impl Cache {
...
+    // This ensures deterministic results and also favors duplicates that are
+    // less nested
+    package_json_files.sort_by(|a, b| {
+      let a_len = a.to_string_lossy().len();
+      let b_len = b.to_string_lossy().len();
+
+      if a_len == b_len {
+        a.cmp(b)
+      } else {
+        a_len.cmp(&b_len)
```
yamadapc, line 150: maybe path depth is a better criteria?

yamadapc: I guess it is going to be the same (?)

crates/atlaspack_shared_map/src/lib.rs
mattcompiles: Hmm I hear you, however I'm only calling it once for logging anyways. I could track the length in this struct, but I'm worried it'll get missed if a remove/clear was ever added. I think we can optimize this later if it's ever a problem.

packages/utils/node-resolver-rs/src/lib.rs
mattcompiles: I considered this but this is actually done in a bunch of places in the resolver. I'm not sure defaulting is a good option as it may lead to unexpected behaviour. At least this would crash the process.

packages/utils/node-resolver-rs/src/cache.rs
mattcompiles: Yeah, I decided to just leave it as it as I don't think there's much difference.

MonicaOlejniczak (approved):

PR #223, opened 2024-11-20
RFC: Using Tokio rather than Rayon for resolver dedupe

alshdavid:
This is a thought experiment on what it would look like to use Tokio rather than Rayon in Matt's [resolver dedupe branch](https://github.com/atlassian-labs/atlaspack/pull/222). Thought it might be a fun code challenge/learning experience and would love your thoughts & feedback on the approach.

Not fussed if this is incorporated as the original Rayon code is fine and has less code - this just serves as an interesting lesson to examine the usage of Rayon vs Tokio and I enjoy talking about programming - haha.

Pros:
- No recursion / better stack traces
- Is a little more parallel - though I doubt we'd see a difference in performance
- Using the existing Tokio runtime/thread pool means we don't need to spawn another one for Rayon
  - Slightly less information to digest in `samply` profiles 

Cons:
- Slightly more code
- Need to reuse or spawn a default Tokio runtime if the caller isn't using one 
  - Rayon hides this "get or init" logic but it's explicit with Tokio

Neutral:
- Code is procedural rather than functional
  - Personal unpopular opinion, I'm too dumb to read functional programming chains so I prefer old school procedural code, even if they are more verbose


packages/utils/node-resolver-rs/src/cache.rs
```diff
@@ -134,58 +134,66 @@ impl Cache {
...
-    // less nested
-    package_json_files.sort_by(|a, b| {
-      let a_len = a.to_string_lossy().len();
-      let b_len = b.to_string_lossy().len();
-
-      if a_len == b_len {
-        a.cmp(b)
-      } else {
-        a_len.cmp(&b_len)
+    super::runtime::block_on(async move {
```
alshdavid, line 137: This will get-or-init a tokio runtime, spawn a Future and block until its completion. Allowing for futures to be use in a non-async context and avoiding the need for propagating `async`/`await` up the call stack.

The `block_on` function is a small wrapper that will either get the current Tokio runtime used by the caller or create a new Tokio thread pool if the caller isn't using one.

packages/utils/node-resolver-rs/src/cache.rs
```diff
@@ -134,58 +134,66 @@ impl Cache {
...
+        let a_len = a.to_string_lossy().len();
+        let b_len = b.to_string_lossy().len();
+
+        if a_len == b_len {
+          a.cmp(b)
+        } else {
+          a_len.cmp(&b_len)
+        }
+      });
+
```
alshdavid, line 154: The above was unchanged, probably included in the diff because of formatting or something 🤷

packages/utils/node-resolver-rs/src/cache.rs
```diff
@@ -134,58 +134,66 @@ impl Cache {
...
-          let dedupe_key = format!("{}@{}", package_json.name, version);
-
-          if let Some(existing) = packages_by_version.get(&dedupe_key) {
-            self
-              .package_duplicates
-              .insert(package_json.path.clone(), existing.clone());
-          } else {
-            packages_by_version.insert(dedupe_key.clone(), package_json.clone());
+
+      let packages = jobs.join_all().await;
```
alshdavid, line 171: -- change ends here --

packages/utils/node-resolver-rs/src/cache.rs
```diff
@@ -283,41 +291,63 @@ fn read_and_parse_package<'a>(
...
+
+    // Collect results
+    let mut package_json_files = Vec::new();
+
+    while let Ok(path) = rx_result.try_recv() {
+      package_json_files.push(path);
+    }
 
-  Ok(package_json_files)
+    Ok(package_json_files)
```
alshdavid, line 349: There's probably a way to do something like `rx_result.collect()` and avoid the for-loop for collection

packages/utils/node-resolver-rs/src/cache.rs
```diff
@@ -283,41 +291,63 @@ fn read_and_parse_package<'a>(
 }
 
 fn find_package_json_files(fs: FileSystemRef, base_path: &Path) -> anyhow::Result<Vec<PathBuf>> {
```
MonicaOlejniczak, line 293: Can't you just make this one async

alshdavid: Good point, didn't realize the caller was async too

PR #225, opened 2024-11-21
V3: Assign deps to assets

alshdavid:
This PR adds dependencies to assets by relating them during the Rust AssetGraph conversion

### DIFF
--- a/packages/core/core/src/requests/AssetGraphRequestRust.js
+++ b/packages/core/core/src/requests/AssetGraphRequestRust.js
@@ -56,28 +56,6 @@ export function createAssetGraphRequestRust(
         options,
       );
 
-      function findDependencyFor(assetId: string) {
-        const dependencies = new Map();
-
-        for (const node of assetGraph.nodes) {
-          if (
-            node &&
-            node.type === 'dependency' &&
-            node.value.sourceAssetId === assetId
-          ) {
-            dependencies.set(node.id, node.value);
-          }
-        }
-
-        return dependencies;
-      }
-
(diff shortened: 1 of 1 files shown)


packages/core/core/src/requests/AssetGraphRequestRust.js
alshdavid, line 1: If the `Asset` is seen first then create a `Map` to contain the `Dependencies` and assign a reference to it to the `Asset.dependencies`.

If a `Dependency` is seen first, create a `Map` to contain `Dependencies` and assign a reference to the `Dependency` there. The `Asset` will collect the reference later.

packages/core/core/src/requests/AssetGraphRequestRust.js
```diff
@@ -234,6 +250,8 @@ function getAssetGraph(serializedGraph, options) {
     }
   }
 
+  dependencyMap.clear();
```
mattcompiles, line 253: Why the clear?

alshdavid: Not really needed and I can remove it if it's ambiguous/not actually valuable.

My thinking was that we should probably save memory where we can and that map will be filled with the string values of every asset from the Asset Graph - clearing it to give the GC a head start.

Though the dependencyMap falls out of scope a second later so it's probably not valuable 😆

alshdavid: Removed

mattcompiles (approved):

PR #230, opened 2024-11-26
Add support for outputting chrome profiles from tracing

yamadapc:
Setting `ATLASPACK_TRACING_MODE=chrome` should produce a JSON file that can be loaded onto perfetto or `chrome://tracing` with our tracing instrumentation. At the moment by default there will be a lot of events internal to the resolver and SWC resulting in quite a large file.


crates/atlaspack_monitoring/src/tracer.rs
```diff
@@ -62,9 +71,16 @@ impl TracerMode {
   }
 }
 
+enum TracerGuard {
```
MonicaOlejniczak, line 74: Why are these unused?

yamadapc: They are just guards, they'll flush some data to disk when dropped

MonicaOlejniczak: Yeah but why does rust think `WorkerGuard` / ChromeGuard` are unused? Specifically asking about the `#[allow(unused)]` annotation

yamadapc: These libraries use RAII pattern -> https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization

We need to keep the guards alive for the duration of the process and the files will be flushed on exit, or if we destroyed the tracing instance at some point.

yamadapc: Rust thinks they are unused because they are set but not read

MonicaOlejniczak (approved):

PR #233, opened 2024-11-26
Add evaluateRootConditionalBundles option so servers can fallback to safe behaviour

JakeLane:
## Motivation

Currently, in Jira, we have to find every way we serve HTML and make sure we send the conditional bundles for the initial page load. This is difficult as there are many legacy server routes we need to support. To avoid explicitly doing this, we can load all the required bundles in the page load as a fallback and remove these explicit inclusions in some cases.

## Changes

- Rename symmetric set difference and add set difference function
- Add conditional bundle edges to the bundle graph
- In HTMLPackager, when enabled, insert conditional bundle modules

## Checklist

### DIFF
--- a/packages/bundlers/default/src/idealGraph.js
+++ b/packages/bundlers/default/src/idealGraph.js
@@ -373,7 +373,10 @@ export function createIdealGraph(
                 dependencyPriorityEdges[dependency.priority],
               );
 
-              if (dependency.priority === 'conditional') {
+              if (
+                getFeatureFlag('conditionalBundlingApi') &&
+                dependency.priority === 'conditional'
+              ) {
                 let [referencingBundleRoot, bundleGroupNodeId] = nullthrows(
                   stack[stack.length - 1],
                 );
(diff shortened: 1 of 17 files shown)


packages/reporters/conditional-manifest/src/ConditionalManifestReporter.js
```diff
@@ -52,9 +52,8 @@ async function report({
 
     for (const target of targets) {
       const conditionalManifestFilename = join(target.distDir, filename);
-
       const conditionalManifest = JSON.stringify(
-        manifest[target.name],
+        manifest[target.name] ?? {},
```
MonicaOlejniczak, line 56: you don't want it to fail?

JakeLane: Yeah, I'll leave a comment here, but it'll be better for rollout if the manifest always exists (so the request 404ing can be actioned in monitoring).

packages/core/core/src/public/MutableBundleGraph.js
```diff
@@ -280,6 +280,13 @@ export default class MutableBundleGraph
     );
   }
 
+  createBundleConditionalReference(from: IBundle, to: IBundle): void {
```
MonicaOlejniczak, line 283: Naming is a bit weird here for me with the `Reference` suffix, documentation would probably help and for any other APIs you introduce

JakeLane: gotcha, I'll add some comments here. It's based on the `createBundleReference` API which is undocumented. Took me a while to untangle what it's actually for.

JakeLane: tl;dr it's for tracking which bundles have references to others, meaning we'd expect it to be loaded. I added a new type of edge because we only want to load it if configured (otherwise we could rely on the existing parallel priority behaviour).

packages/bundlers/default/src/idealGraph.js
```diff
@@ -377,8 +382,24 @@ export function createIdealGraph(
                   bundleRoots.get(referencingBundleRoot),
                 )[0];
 
-                bundleRoots.set(childAsset, [bundleId, bundleGroupNodeId]);
-                bundleGraph.addEdge(referencingBundleId, bundleId);
+                if (
+                  getFeatureFlag('conditionalBundlingApi') &&
```
MonicaOlejniczak, line 386: How come this isn't in your initial `dependency.priority === 'conditional'` check? Isn't it already implied when the dependency is conditional anyway (though having it is safer)

packages/core/core/src/BundleGraph.js
```diff
@@ -2272,4 +2283,30 @@ export default class BundleGraph {
...
+  getReferencedConditionalBundles(bundle: Bundle): Array<Bundle> {
+    let referencedBundles = new Set();
+    this._graph.dfs({
+      visit: (nodeId) => {
+        let node = nullthrows(this._graph.getNode(nodeId));
+        if (node.type !== 'bundle') {
+          return;
+        }
+
+        if (node.value.id === bundle.id) {
```
MonicaOlejniczak, line 2296: / nit you can combine the ifs

packages/bundlers/default/src/idealGraph.js
JakeLane: good catch, I think this is a symptom of copy/paste code. Will fix

MonicaOlejniczak (approved):

PR #235, opened 2024-11-27
[V3] Ensure Asset ids are in sync with V2 / Fix changing filetype pipelines

mattcompiles:
## Motivation

This PR makes a couple of changes to ensure that Asset ids are the same as in V2 builds. The primary reason for the difference being that V2 updates the Asset id whenever the file type is updated.   

This PR also fixes an issue where pipelines would be executed correctly on assets that changed file type. 

## Changes

- Add  `update_id` method to Asset
- Add `set_meta_id` method to Asset
- Update the Asset id whenever a file_type change is detected in the Asset request
- Set the `Asset.meta.id` before the id updates as the packager expects
- Return an undefined query from the node transformers when the query is empty
- Update the default browserslist to match V2
- Ensure Assets run through a new pipeline if they change file type
- Only read the .browserslistrc when it is present. This removes a bunch of logging noise from the tests.
- Update lightningcss and fix CSS nesting test to use an older browser so the native CSS transformer is in sync with the JS one

## Checklist

### DIFF
--- a/crates/atlaspack/src/requests/asset_request.rs
+++ b/crates/atlaspack/src/requests/asset_request.rs
@@ -17,6 +17,7 @@ use atlaspack_sourcemap::find_sourcemap_url;
 use atlaspack_sourcemap::load_sourcemap_url;
 use std::collections::VecDeque;
 use std::hash::Hash;
+use std::path::Path;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Instant;
@@ -111,8 +112,13 @@ impl Request for AssetRequest {
     });
 
     let transform_context = TransformContext::new(config_loader, self.env.clone());
-    let mut result =
-      run_pipelines(transform_context, asset, request_context.plugins().clone()).await?;
+    let mut result = run_pipelines(
+      transform_context,
+      asset,
+      request_context.plugins().clone(),
+      &request_context.project_root,
+    )
+    .await?;
 
     // TODO: Commit the asset with a project path now, as transformers rely on an absolute path
     // result.asset.file_path = to_project_path(&self.project_root, &result.asset.file_path);
@@ -143,6 +149,7 @@ async fn run_pipelines(
   transform_context: TransformContext,
   input: Asset,
   plugins: PluginsRef,
+  project_root: &Path,
 ) -> anyhow::Result<TransformResult> {
   let mut invalidations = vec![];
   let mut asset_queue = VecDeque::from([(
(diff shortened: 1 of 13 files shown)


crates/atlaspack/src/requests/asset_request.rs
```diff
@@ -211,20 +226,23 @@ async fn run_pipelines(
...
-    // We assume the first asset to complete the pipeline is the initial asset
-    if initial_asset.is_none() {
-      initial_asset = Some(current_asset);
-      initial_asset_dependencies = Some(current_dependencies);
-    } else {
-      processed_assets.push(AssetWithDependencies {
-        asset: current_asset,
-        dependencies: current_dependencies,
-      });
+    if !pipeline_incomplete {
```
MonicaOlejniczak, line 235: This is hard to read with the negation, I would rename it to `pipeline_completed` so you would have

```rust
pipeline_completed = false

if pipeline_completed {}
```

crates/atlaspack_core/src/types/asset.rs
```diff
@@ -368,10 +368,26 @@ impl Asset {
     }
   }
 
+  pub fn update_id(&mut self, project_root: &Path) {
+    self.id = create_asset_id(CreateAssetIdParams {
+      code: None,
```
MonicaOlejniczak, line 373: Code is passed in for discovered assets, should that be maintained?

crates/atlaspack/src/requests/asset_request.rs
```diff
@@ -198,8 +206,15 @@ async fn run_pipelines(
 
       // If the Asset has changed type then we may need to trigger a different pipeline
       if current_asset.file_type != original_asset_type {
-        let next_pipeline =
-          plugins.transformers(&current_asset.file_path, current_asset.pipeline.clone())?;
+        // When the Asset changes file_type we need to regenerate it's id
```
MonicaOlejniczak, line 209:
```suggestion
        // When the Asset changes file_type we need to regenerate its id
```

crates/atlaspack/src/requests/asset_request.rs
mattcompiles: Haha I went back and forth on this one. Guess I chose poorly 😂

crates/atlaspack_core/src/types/asset.rs
mattcompiles: Yeah I also considered this, however it's tricky because we store the code on the Asset directly right now and can't tell between a discovered asset or a regular one. We should address this at some point, however I felt that `file_path + unique_key` would suffice for uniqueness.

MonicaOlejniczak (approved):

PR #237, opened 2024-11-28
Add sourcemaps tests

MonicaOlejniczak:
## Motivation

There are some mismatches between v3 sourcemaps on the branch I have been working on vs v2. These changes introduce some tests I have written along the way to try and identify issues.

## Changes

* Add some public apis to the rust sourcemap type and add some tests for the conversions
* Remove unnecessary asset code handling
* Add sourcemaps tests using mozilla sourcemaps, possibly identifying an issue with how TS enums are mapped

## Checklist


packages/core/integration-tests/test/sourcemaps.js
```diff
@@ -1476,4 +1478,180 @@ describe('sourcemaps', function () {
       });
     },
   );
+
+  // TODO: Enable this for v3 once all the code has been integrated
+  it.v2('creates valid source maps for js transformers', async () => {
```
mattcompiles, line 1483: What's currently missing in V3 to enable this?

mattcompiles (approved):

packages/core/integration-tests/test/sourcemaps.js
MonicaOlejniczak: JS plugin support in the compat layer

PR #240, opened 2024-11-28
V3: Support sourcemaps in JavaScript plugins

MonicaOlejniczak:
## Motivation

Currently v3 does not support sourcemaps in JavaScript plugins, these changes update the workflow to support this feature. It has been found that one soucemap matches v2 internally, but the remaining are difficult to discern with all the bundle output differences in v3.

## Changes

* Skip serialising the map to JavaScript and commit it to LMDB instead
* Set map key so that the map can be retrieved via LMDB
* Pass map into JavaScript plugin worker and store the result back in Rust
* Add implementations for the JavaScript plugin `getMap` and `setMap`
* Ensure `generate` sets the map if provided in the result (this is what babel relies on)
* Update `CommittedAsset` map buffer handling to account for differences with v3 maps
* Enable v3 sourcemap tests

## Checklist

### DIFF
--- a/crates/atlaspack_plugin_rpc/src/nodejs/plugins/nodejs_rpc_transformer.rs
+++ b/crates/atlaspack_plugin_rpc/src/nodejs/plugins/nodejs_rpc_transformer.rs
@@ -3,6 +3,7 @@ use atlaspack_core::plugin::PluginOptions;
 use napi::bindgen_prelude::FromNapiValue;
 use napi::JsBuffer;
 use napi::JsObject;
+use napi::JsString;
 use napi::JsUnknown;
 use std::fmt;
 use std::fmt::Debug;
@@ -107,6 +108,13 @@ impl TransformerPlugin for NodejsRpcTransformerPlugin {
     let asset_env = asset.env.clone();
     let stats = asset.stats.clone();
 
+    let original_source_map = asset.map.clone();
+    let source_map = if let Some(map) = asset.map.as_ref() {
+      Some(map.to_json()?)
+    } else {
+      None
+    };
+
     let run_transformer_opts = RpcTransformerOpts {
       key: self.plugin_node.package_name.clone(),
       options: state.rpc_plugin_options.clone(),
@@ -117,7 +125,7 @@ impl TransformerPlugin for NodejsRpcTransformerPlugin {
       },
     };
 
-    let (result, contents) = self
+    let (result, contents, map) = self
       .nodejs_workers
       .next_worker()
       .transformer_register_fn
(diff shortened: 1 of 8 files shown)


crates/atlaspack_plugin_rpc/src/nodejs/plugins/nodejs_rpc_transformer.rs
```diff
@@ -117,18 +124,24 @@ impl TransformerPlugin for NodejsRpcTransformerPlugin {
...
         move |env| {
           let run_transformer_opts = env.to_js_value(&run_transformer_opts)?;
 
-          let mut result = env.create_buffer(asset.code.len())?;
-          result.copy_from_slice(&asset.code);
+          let mut contents = env.create_buffer(asset.code.len())?;
+          contents.copy_from_slice(&asset.code);
+
+          let map = if let Some(map) = source_map {
+            env.create_string(&map)?.into_unknown()
```
mattcompiles, line 139: Wonder if we should use a buffer here? Strings of code were causing big problems with perf before.

packages/core/core/src/atlaspack-v3/worker/worker.js
```diff
@@ -269,19 +275,24 @@ export class AtlaspackWorker {
...
           symbols: mutableAsset.symbols.intoNapi(),
+          type: mutableAsset.type,
           uniqueKey: mutableAsset.uniqueKey,
-          sideEffects: mutableAsset.sideEffects,
-          isBundleSplittable: mutableAsset.isBundleSplittable,
-          isSource: mutableAsset.isSource,
         },
         await mutableAsset.getBuffer(),
+        // Send back the original map if it wasn't set
+        mutableAsset.isMapDirty
```
mattcompiles, line 292: Wonder if it'd be cheaper to just send nothing when it's not changed?

packages/core/core/src/requests/AssetGraphRequestRust.js
```diff
@@ -181,21 +175,15 @@ function getAssetGraph(serializedGraph, options) {
       asset.committed = true;
       asset.contentKey = id;
       asset.env.id = getEnvId(asset.env);
+      // We need to add this property for source map handling, as some assets like runtimes
+      // are processed after the Rust transformation and take the v2 code path
```
mattcompiles, line 179: Side note: I've been wondering if we should use the new asset graph for runtimes too. There's a small perf improvement we could make there too.

crates/atlaspack_plugin_rpc/src/nodejs/plugins/nodejs_rpc_transformer.rs
MonicaOlejniczak: I had a buffer initially, but you just end up doing more conversions on the JS side. Strings are what we use for serialisation of the whole asset graph nodes, so I don't think it's a big issue.

packages/core/core/src/atlaspack-v3/worker/worker.js
MonicaOlejniczak: Yes, it's just a bit tedious on the napi side but I can revisit it.

mattcompiles (approved):

PR #243, opened 2024-12-02
[V3] Make V3 output closer to V2

mattcompiles:
## Motivation

This PR attempts to make the V3 output more closely match the V2 one. I have added some tests that build apps with both versions and assert their output is identical. I have also made a few changes to ensure the tests are passing. 

## Changes

- Add tests for ensuring V2 and V3 output are identical
- Change CSS module quote style to match V2
- Only use virtual code in ID generation when it comes from the resolver (matches V2 behaviour)
- Replicate V2 environment for `esmodule-helpers` import

## Checklist

### DIFF
--- a/crates/atlaspack_core/src/types/asset.rs
+++ b/crates/atlaspack_core/src/types/asset.rs
@@ -221,7 +221,7 @@ pub struct Asset {
   /// True if the Asset's code was returned from a resolver rather than being
   /// read from disk.
   #[serde(skip_serializing)]
-  pub is_virtual_code: bool,
+  pub is_virtual: bool,
 
   /// True if the asset has CommonJS exports
   pub has_cjs_exports: bool,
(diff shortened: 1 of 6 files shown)


crates/atlaspack_core/src/types/asset.rs
```diff
@@ -218,6 +218,11 @@ pub struct Asset {
   ///
   pub is_source: bool,
 
+  /// True if the Asset's code was returned from a resolver rather than being
+  /// read from disk.
+  #[serde(skip_serializing)]
+  pub is_virtual_code: bool,
```
MonicaOlejniczak, line 224: The entire asset is virtual, not just the code. So it should be `is_virtual` imo

mattcompiles: Agreed, dunno what I was thinking

MonicaOlejniczak (approved):

PR #246, opened 2024-12-03
Expose rust tracing instrumentation to javascript

yamadapc:
This commit adds a way to instrument javascript code with the rust tracing implementation.

This allows unified chrome profiles to be generated that include both rust and javascript spans.

Since span names need to be static, currently javascript spans will need to be labeled with a `label` arg/attribute.

### DIFF
--- a/packages/core/integration-tests/test/bundler.js
+++ b/packages/core/integration-tests/test/bundler.js
@@ -14,7 +14,7 @@ import {
 import {hashString} from '@atlaspack/rust';
 import {normalizePath} from '@atlaspack/utils';
 
-describe('bundler', function () {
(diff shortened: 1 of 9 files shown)


packages/core/integration-tests/test/bundler.js
```diff
@@ -14,7 +14,7 @@ import {
 import {hashString} from '@atlaspack/rust';
 import {normalizePath} from '@atlaspack/utils';
 
-describe('bundler', function () {
+describe.only('bundler', function () {
```
MonicaOlejniczak, line 17: Remove only

MonicaOlejniczak (approved):

PR #247, opened 2024-12-03
WIP - Warn if transformer does not update source-map

yamadapc:
This adds a warning if a transformer changes the asset source but does not update its source-map.

- - -

WIP because I'll update this to remove non JS/TS warnings and to log only a reasonable amount.

### DIFF
--- a/packages/core/core/src/Transformation.js
+++ b/packages/core/core/src/Transformation.js
@@ -649,17 +649,25 @@ export default class Transformation {
     }
 
     // Transform.
-    let transfomerResult: Array<TransformerResult | MutableAsset> =
+    const mutableAsset = new MutableAsset(asset);
+    const transfomerResult: Array<TransformerResult | MutableAsset> =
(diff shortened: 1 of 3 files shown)


packages/core/core/src/Transformation.js
```diff
@@ -649,17 +649,25 @@ export default class Transformation {
...
         config,
         options: pipeline.pluginOptions,
         resolve,
         logger,
         tracer,
       });
-    let results = await normalizeAssets(this.options, transfomerResult);
+    const results = await normalizeAssets(this.options, transfomerResult);
+
+    if (mutableAsset.hasUpdatedSource && !mutableAsset.hasUpdatedSourceMap) {
```
MonicaOlejniczak, line 665: I think you should consider modifying the check to also verify if a sourcemap exists yet. For example, some transformers rely on babel / swc to add the source map later and this will have correct mappings. You'd probably need to also add a check at the end to make sure a sourcemap exists after all pipelines run.

PR #248, opened 2024-12-03
Clean-up fast inline requires flag

yamadapc:
Clean-up inline requires rust migration feature-flag.

The final improvement is a 11.5% statistically significant improvement on certain production builds.

Screenshot 2024-11-04 at 4 57 34 PM


packages/optimizers/inline-requires/src/InlineRequires.js
```diff
@@ -45,11 +42,10 @@ module.exports = new Optimizer<empty, BundleConfig>({
     return {assetPublicIdsWithSideEffects};
   },
 
-  async optimize({
+  optimize({
```
JakeLane, line 45: why was this dropped?

JakeLane (approved):

packages/optimizers/inline-requires/src/InlineRequires.js
yamadapc: There's no await anymore

PR #261, opened 2024-12-03
Modify BundleGraph::getAssetWithDependency to use graph iteration methods

yamadapc:
Prevents creating temporary arrays when finding the asset linked to a dependency. This is a relatively hot path in the bundler and can improve its performance.

I will perform a dev release from this branch and do some performance testing.

### DIFF
--- a/packages/core/graph/src/Graph.js
+++ b/packages/core/graph/src/Graph.js
@@ -163,7 +163,7 @@ export default class Graph<TNode, TEdgeType: number = 1> {
     return this.adjacencyList.hasEdge(from, to, type);
   }
 
-  forEachNodeIdConnectedTo(to: NodeId, fn: (nodeId: NodeId) => void) {
+  forEachNodeIdConnectedTo(to: NodeId, fn: (nodeId: NodeId) => boolean | void) {
     this._assertHasNodeId(to);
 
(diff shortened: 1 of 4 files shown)


JakeLane (approved):

packages/core/graph/src/Graph.js
```diff
@@ -163,16 +163,20 @@ export default class Graph<TNode, TEdgeType: number = 1> {
     return this.adjacencyList.hasEdge(from, to, type);
   }
 
-  forEachNodeIdConnectedTo(to: NodeId, fn: (nodeId: NodeId) => void) {
+  forEachNodeIdConnectedTo(
+    to: NodeId,
+    fn: (nodeId: NodeId) => boolean | void,
+    type: AllEdgeTypes | TEdgeType | NullEdgeType = 1,
```
MonicaOlejniczak, line 169: Any reason you're using `1` over the actual type?

yamadapc: We just don't have the `NULL_EDGE_TYPES` constant declared anywhere, but I can declare and use that

yamadapc: https://github.com/atlassian-labs/atlaspack/pull/261/commits/fcce72b126c502edec86e4b8345ec0ba33bbab0c

PR #263, opened 2024-12-05
V3: Passing diagnostics from Rust

alshdavid:
Errors are currently sent to the JavaScript layer as strings. This is a problem because the JavaScript layer currently handles formatting the error output from a `Diagnostic` object - which cannot be done from a string leading to poor error messages.

This PR allows Rust to return `Diagnostic` objects to JavaScript to handle formatting and printing. This works by having Rust check if the internal `anyhow::Error` is a `Diagnostic`, if it is then it converts it to JavaScript and returns that otherwise it returns the error in string form.

On the Napi side, to serialize the `Diagnostic` value we need access to `Env` which isn't available in `deferred.reject()` so instead the error value is passed via `deferred.resolve()`. 

This then necessitated `AtlaspackNapi.buildAssetGraph()` returning a tuple and rethrowing is handled in the JavaScript layer.

```typescript
type Ok<T> = [T, null]
type Err<E> = [null, E]
type Result<T, E> = Ok<T> | Err<E>
```

## Preview

By comparison to V2 error messages, V3 error messages are pretty low on detail at this stage

#### V2 Error Message

image

#### V3 Error Message

image

## Todo

We need to add additional error information (`codeFrames`, `hints`, etc) to the error handling of our internals (resolver, transformer, etc). 

For the time being this PR allows Diagnostic errors to propage to JavaScript, following PRs should look at addressing the contents of those errors

### DIFF
--- a/crates/atlaspack/src/error.rs
+++ b/crates/atlaspack/src/error.rs
@@ -4,7 +4,7 @@ use serde::Serialize;
 
 pub enum AtlaspackError {
   Diagnostic(Diagnostic),
-  String(String),
+  Unknown(String),
 }
 
 impl Serialize for AtlaspackError {
@@ -14,7 +14,7 @@ impl Serialize for AtlaspackError {
   {
     match self {
       AtlaspackError::Diagnostic(diagnostic) => diagnostic.serialize(serializer),
-      AtlaspackError::String(message) => message.serialize(serializer),
+      AtlaspackError::Unknown(message) => message.serialize(serializer),
     }
   }
 }
@@ -24,9 +24,9 @@ impl From<&anyhow::Error> for AtlaspackError {
     if let Some(diagnostic) = error.downcast_ref::<Diagnostic>() {
       Self::Diagnostic(diagnostic.clone())
     } else if let Some(message) = error.downcast_ref::<String>() {
-      Self::String(message.clone())
+      Self::Unknown(message.clone())
     } else {
-      Self::String(error.to_string())
+      Self::Unknown(error.to_string())
     }
   }
 }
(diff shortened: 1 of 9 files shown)


crates/atlaspack/src/request_tracker/request_tracker.rs
```diff
@@ -224,7 +224,7 @@ impl RequestTracker {
 
     *request_node = match result {
       Ok(result) => RequestNode::Valid(result.clone()),
-      Err(error) => RequestNode::Error(anyhow!(error.to_string())),
+      Err(error) => RequestNode::Error(AtlaspackError::from(error).into()),
     };
```
alshdavid, line 228: Rather then returning the error as a string, unpack the error, clone the internals and return a new error.

crates/atlaspack/src/request_tracker/request_tracker.rs
```diff
@@ -249,7 +249,7 @@ impl RequestTracker {
     match request_node {
       RequestNode::Root => Err(diagnostic_error!("Impossible")),
       RequestNode::Incomplete => Err(diagnostic_error!("Impossible")),
-      RequestNode::Error(error) => Err(anyhow!(error.to_string())),
+      RequestNode::Error(error) => Err(AtlaspackError::from(error).into()),
```
alshdavid, line 252: And again, clone/propagate the error here

crates/atlaspack/src/error.rs
```diff
@@ -0,0 +1,41 @@
+use anyhow::anyhow;
+use atlaspack_core::types::Diagnostic;
+use serde::Serialize;
+
+pub enum AtlaspackError {
+  Diagnostic(Diagnostic),
+  String(String),
```
mattcompiles, line 7: I wonder if `UnknownError` or similar would be more accurate?

mattcompiles (approved):
Nice one mate. Just a small question about the naming.

crates/atlaspack/src/error.rs
MonicaOlejniczak: Agree this would be better

MonicaOlejniczak (approved):

PR #265, opened 2024-12-06
Fix inline requires source map generation

yamadapc:
The wrong source-map was getting returned from inline requires.

An integration test is added to verify the mapping is correct (and was failing previously with the inline requires optimizer, but not with only the transformer).


JakeLane (approved):

yamadapc:
v3 is broken here ; I might skip the test on v3

PR #266, opened 2024-12-06
V3: Error messages for @atlaspack/transformer-js

alshdavid:
This PR maps the `atlaspack_js_swc_core::Diagnostic` type to the new `atlaspack_core::types::Diagnostic` and propagates that to the JavaScript layer.

I had to add a `Diagnostics` type that can be contained within `anyhow`, the downcasting logic and serialization logic for it to work in the JavaScript layer.

### V2

image

### Old V3

image

### New V3

image


crates/atlaspack_plugin_transformer_js/src/map_diagnostics.rs
```diff
@@ -0,0 +1,90 @@
...
+
+  if let Some(code_highlights) = input.code_highlights {
+    let mut code_frame = CodeFrame {
+      code_highlights: vec![],
+      code: source_code.clone(),
+      language: file_type.map(|v| v.into()),
+      file_path: path.clone(),
+    };
+
+    println!("{:?}", path);
```
mattcompiles, line 71: println

alshdavid: Oft, good catch. We should probably enable a lint rule to ban `println!`

crates/atlaspack/src/error.rs
```diff
@@ -1,9 +1,11 @@
 use anyhow::anyhow;
 use atlaspack_core::types::Diagnostic;
+use atlaspack_core::types::Diagnostics;
 use serde::Serialize;
 
 pub enum AtlaspackError {
   Diagnostic(Diagnostic),
+  Diagnostics(Diagnostics),
```
mattcompiles, line 8: Why not just make this `Diagnostics(Vec<Diagnostic>)`?

packages/transformers/js/core/src/lib.rs
```diff
@@ -10,7 +10,7 @@ mod modules;
 mod node_replacer;
 pub mod test_utils;
 mod typeof_replacer;
-mod utils;
+pub mod utils;
```
MonicaOlejniczak, line 13: This shouldn't need to be public?

crates/atlaspack/src/error.rs
alshdavid: `Display` is needed for `anyhow` and you can't add `Display` to `Vec` because the type is external to the current crate.  

So you have to create a wrapper to make the value usable as an `Error`.

packages/transformers/js/core/src/lib.rs
alshdavid: Yep, updated

MonicaOlejniczak (approved):
Nice job

PR #268, opened 2024-12-09
Add domain sharding to js runtime

benjervis:
In order to support domain sharding for HTTP1.1 users, we need to apply shards in two places. Firstly, in the generated HTML, which is done at request time by the SSR server (as we need to know whether this is a 1.1 request or not). Secondly, when requesting bundles from within other bundles we need to shard those requests as well, and not just rely on the parent shard.

To support this, I've made some changes to the JS Runtime, adding in a sharded version of `helpers/bundle-url` which is activated if the domain sharding config is provided in a build.

```json
"@atlaspack/runtime-js": {
  "domainSharding": {
    "maxShards": 5,
    "cookieName": "should.shard.domains"
  }
}
```

`bundle-url-shards` started out as a copy of `bundle-url`, but has been refactored a bit to make the more complicated logic easier to work with. That number of functions also became a little tricky to track, so the file is now a TypeScript file which will be compiled in the client build.

There was a todo in here to migrate `getOrigin` to just using the URL constructor once we didn't need to support IE 11 anymore, which we don't, so I've made that change and brought it across to `bundle-url` also.

Lastly there are a bunch of tests to make sure the sharding logic does what it's meant to, and that the runtime generates the call with the right arguments.

### DIFF
--- a/packages/runtimes/js/src/helpers/bundle-url-shards.ts
+++ b/packages/runtimes/js/src/helpers/bundle-url-shards.ts
@@ -1,4 +1,8 @@
-import {getBaseURL} from './bundle-url-common';
+import {
+  getBaseURL,
+  stackTraceUrlRegexp,
+  stackTraceUrlRegexp,
+} from './bundle-url-common';
 
 const bundleURL: Record<string, string> = {};
 
@@ -18,9 +22,7 @@ export function getShardedBundleURL(
   try {
     throw inputError ?? new Error();
   } catch (err) {
-    var matches = ('' + err.stack).match(
-      /(https?|file|ftp|(chrome|moz|safari-web)-extension):\/\/[^)\n]+/g,
-    );
+    var matches = ('' + err.stack).match(stackTraceUrlRegexp);
 
     if (!matches) {
       return '/';
--- a/packages/runtimes/js/src/helpers/bundle-url-shards.ts
+++ b/packages/runtimes/js/src/helpers/bundle-url-shards.ts
@@ -1,8 +1,8 @@
-import {getBaseURL, stackTraceUrlRegexp} from './bundle-url-common';
+const {getBaseURL, stackTraceUrlRegexp} = require('./bundle-url-common');
 
 const bundleURL: Record<string, string> = {};
 
-export function getShardedBundleURL(
+function getShardedBundleURL(
   bundleName: string,
   cookieName: string,
   cookieString: string,
@@ -91,3 +91,5 @@ function removeTrailingShard(subdomain: string) {
   const shardIdx = subdomain.lastIndexOf('-');
   return subdomain.slice(0, shardIdx);
 }
+
+exports.getShardedBundleURL = getShardedBundleURL;
--- a/packages/runtimes/js/src/helpers/bundle-url-shards.ts
+++ b/packages/runtimes/js/src/helpers/bundle-url-shards.ts
@@ -1,3 +1,5 @@
+import {getBaseURL} from './bundle-url-common';
+
 const bundleURL: Record<string, string> = {};
 
 export function getShardedBundleURL(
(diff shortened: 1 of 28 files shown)


packages/runtimes/js/src/helpers/bundle-url-shards.ts
```diff
@@ -0,0 +1,102 @@
...
+  if (!trailingShardRegex.test(subdomain)) {
+    return subdomain;
+  }
+
+  const shardIdx = subdomain.lastIndexOf('-');
+  return subdomain.slice(0, shardIdx);
+}
+
+// Get the URL without the filename (last / segment)
+export function getBaseURL(url: string) {
```
mattcompiles, line 96: Let's share this version of `getBaseURL` with the normal bundle-url asset.

packages/runtimes/js/src/helpers/bundle-url-shards.ts
```diff
@@ -0,0 +1,102 @@
...
+  inputError?: string,
+): string {
+  let value = bundleURL[bundleId];
+
+  if (value) {
+    return value;
+  }
+
+  try {
+    throw inputError ?? new Error();
```
marcins, line 17: Does the original implementation of this in `bundle-url.js` cache the result of this Error stack based thing? It's the same result for the whole runtime isn't it? It's not dependent on which bundle it's called from.

It might be worthwhile as calculating an Error stack trace is relatively expensive.

benjervis: No, it throws a new Error each time. And no I believe the result here will be the URL of the bundle that's requesting this one, so it will be different depending on when/where it's getting called.

packages/runtimes/js/src/helpers/bundle-url-shards.ts
```diff
@@ -0,0 +1,102 @@
...
+    // Use the 2nd one, which will be a runtime in the original bundle.
+    const stackUrl = matches[1];
+    const baseUrl = getBaseURL(stackUrl);
+
+    // If the cookie doesn't exist then we don't need to shard
+    if (cookieString.indexOf(cookieName) === -1) {
+      return baseUrl;
+    }
+
+    const shardNumber = getDomainShardIndex(bundleId, maxShards);
```
marcins, line 37: The server-side code for this presumably doesn't use the `bundleId` since it only has access to the bundle URL at that point - for some async bundles they get preloaded (e.g. `async-issue-app`) and then later actually loaded by the frontend code - this means that the index we calculate here for that shared bundle might not match what we calculate on the server-side, and so we end up with a preload "miss".

benjervis: Oh that's true. I think I need name here instead

packages/runtimes/js/src/helpers/bundle-url-shards.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    value = url.toString();
+
+    bundleURL[bundleId] = value;
+    return value;
+  }
+}
+
+function getDomainShardIndex(str: string, maxShards: number) {
+  let shard = str.split('').reduce((a, b) => {
```
JakeLane, line 51: nit: a conventional for loop would be faster

JakeLane (approved):

packages/runtimes/js/src/helpers/bundle-url-shards.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  if (value) {
+    return value;
+  }
+
+  try {
+    throw inputError ?? new Error();
+  } catch (err) {
+    var matches = ('' + err.stack).match(
+      /(https?|file|ftp|(chrome|moz|safari-web)-extension):\/\/[^)\n]+/g,
```
yamadapc, line 22: Might want to hoist the regex

mattcompiles (approved):
Nice one bud!

PR #276, opened 2024-12-12

alshdavid:
Working towards fixing the flakey integration tests. This PR adds type checking to some tests that are missing types.

**Notes:** 

Type checking only occurs on test files that are annotated with `// @flow` however this is not enforced and files not annotated can still be written with flow and will still have their types stripped by babel register. 

This leads to misleading usage of APIs and related problems.

Currently, many of the tests that are not type checked are actually using incorrect type signatures to call Atlaspack APIs making it difficult to refactor the tests to improve their reliability.

In future I will enable type checking enforcement on tests, this will help with tracing the cause of brittle tests and also help with the eventual migration to TypeScript

### DIFF
--- a/packages/core/cache/test/LMDBLiteCache.test.js
+++ b/packages/core/cache/test/LMDBLiteCache.test.js
@@ -1,8 +1,9 @@
-import * as path from 'node:path';
+// @flow
+import * as path from 'path';
 import {tmpdir} from 'os';
 import {LMDBLiteCache} from '../src/index';
-import {deserialize, serialize} from 'node:v8';
-import assert from 'node:assert';
+import {deserialize, serialize} from 'v8';
+import assert from 'assert';
 
 const cacheDir = path.join(tmpdir(), 'lmdb-lite-cache-tests');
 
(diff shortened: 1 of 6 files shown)


packages/core/cache/test/LMDBLiteCache.test.js
alshdavid, line 1: flow does not support the `node:*` prefix for stdlib imports

packages/core/core/test/requests/AssetGraphRequestRust.test.js
alshdavid, line 1: Flow doesn't see `assert` as an exit so I added if statements to exit early to avoid type checks failing on nullable values

packages/core/integration-tests/test/bundle-text.js
```diff
@@ -201,7 +203,7 @@ describe('bundle-text:', function () {
             assets: [
               'index.js',
               'main.js',
-              !scopeHoist && 'esmodule-helpers.js',
+              ...(!scopeHoist ? ['esmodule-helpers.js'] : []),
             ].filter(Boolean),
```
mattcompiles, line 207: filter shouldn't be required anymore

alshdavid: Updated

yamadapc (approved):

PR #277, opened 2024-12-12
Reliability: Flow types for tests (part 2)

alshdavid:
[Continuing from here](https://github.com/atlassian-labs/atlaspack/pull/276)

Breaking this down into smaller PRs because it will be easier to review and also I can't run the tests locally so smaller changes help with narrowing down issues in the tests

### DIFF
--- a/flow-typed/npm/rimraf_v5.x.x.js
+++ b/flow-typed/npm/rimraf_v5.x.x.js
@@ -9,6 +9,7 @@ declare module 'rimraf' {
 
   declare module.exports: {
     (f: string, opts?: Options): Promise<boolean>,
+    rimraf(f: string, opts?: Options): Promise<boolean>,
     sync(path: string, opts?: Options): boolean,
     ...
   };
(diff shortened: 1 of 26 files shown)


packages/core/cache/test/LMDBLiteCache.test.js
alshdavid, line 1: flow does not support the node:* prefix for stdlib imports

flow-typed/npm/rimraf_v5.x.x.js
alshdavid, line 1: For `packages/core/integration-tests/test/feature-flags.js`

packages/core/integration-tests/test/bundler.js
```diff
@@ -885,10 +886,16 @@ describe('bundler', function () {
       },
     ]);
 
+    let assetBar = findAsset(b, 'bar.js');
+    if (!assetBar) return assert(false);
```
yamadapc, line 890: btw if you import assert as invariant then it can be used as type refinement

```
import invariant from 'assert';

const stuff: null | string = 'something';
invariant(stuff != null);
// stuff is refined to `string`
```

yamadapc (approved):

PR #278, opened 2024-12-12
Reliability: Flow types for tests (part 3)

alshdavid:
[Continuing from here](https://github.com/atlassian-labs/atlaspack/pull/277)

### DIFF
--- a/flow-typed/node/assert.js
+++ b/flow-typed/node/assert.js
@@ -1,0 +1,29 @@
+// Missing flow types for node:assert
+declare module "assert/strict" {
+  declare module.exports: {
+    (value: any, message?: string): void;
+    ok(value: any, message?: string): void;
+    fail(message?: string | Error): void;
+    // deprecated since v10.15
+    fail(actual: any, expected: any, message: string, operator: string): void;
+    equal(actual: any, expected: any, message?: string): void;
+    strictEqual(actual: any, expected: any, message?: string): void,
+    notEqual(actual: any, expected: any, message?: string): void;
+    deepEqual(actual: any, expected: any, message?: string): void;
+    notDeepEqual(actual: any, expected: any, message?: string): void;
+    throws(
+      block: Function,
+      error?: Function | RegExp | (err: any) => boolean,
+      message?: string
+    ): void;
+    rejects(
+      block: Function,
+      error?: Function | RegExp | (err: any) => boolean,
+      message?: string
+    ): void;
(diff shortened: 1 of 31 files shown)


packages/core/integration-tests/test/globals.js
```diff
@@ -58,7 +59,6 @@ describe('globals', function () {
     it('when scope hoisting is disabled', async function () {
       let bundleGraph = await bundle(path.join(dir, 'index.js'), {
         defaultTargetOptions: {
-          context: 'browser',
```
alshdavid, line 61: This is not on `InitialOptions` and does not appear to be used within core

packages/core/integration-tests/test/globals.js
alshdavid, line 1: `defaultTargetOptions.context` is not on `InitialOptions` and does not appear to be used within core

packages/core/integration-tests/test/html.js
```diff
@@ -1,4 +1,5 @@
-import assert from 'assert';
+// @flow
+import assert from 'assert/strict';
```
alshdavid, line 2: Using local assert definition to add in `assert.rejects`

packages/core/integration-tests/test/html.js
```diff
@@ -2647,7 +2674,10 @@ describe('html', function () {
 
       for (let htmlBundle of b.getBundles().filter((b) => b.type === 'html')) {
         let htmlSiblings = b
-          .getReferencedBundles(htmlBundle, true)
+          .getReferencedBundles(htmlBundle, {
+            recursive: true,
+            includeInline: true,
+          })
```
alshdavid, line 2680: I think this is what it was intending to do?

flow-typed/node/assert.js
alshdavid, line 1: flow types missing for `assert.rejects`, using override to pass type check

packages/core/integration-tests/test/html.js
```diff
@@ -2317,7 +2333,7 @@ describe('html', function () {
       });
 
       // could run in either order.
-      assert(output.sort(), ['a', 'b', 'c']);
```
alshdavid, line 2320:
```javascript
assert(['a', 'b', 'c', 'c'], ['a', 'b', 'c'])
```
Was erroneously passing. I have updated the assertion to include the missing asset. From the `assertBundles` above, my change looks correct but I'm not 100% sure. Please confirm that it's the expected output

yamadapc (approved):

PR #279, opened 2024-12-12
Reliability: Flow types for tests (part 4)

alshdavid:
Continue from here [link](https://github.com/atlassian-labs/atlaspack/pull/278)


packages/core/test-utils/src/stubs.js
alshdavid, line 1: We have many tests that are marked as skipped which depend on these functions. The tests were no updated when these functions were removed and they were not detected because type checking was disabled on those files.

I added these stubs to pass the type checker, we will need to correct the skipped tests in the future

packages/core/integration-tests/test/worklets.js
```diff
@@ -69,6 +70,7 @@ describe('atlaspack', function () {
         },
       },
     });
+    if (!url) return assert.fail();
     assert(/^http:\/\/localhost\/worklet\.[0-9a-f]+\.js$/.test(url));
```
MonicaOlejniczak, line 74: /nit you could have `.test(url ?? '')` to avoid the assertion failure. Same for some others in this file

packages/core/test-utils/src/stubs.js
MonicaOlejniczak: Do we want to keep those skipped tests, or should the be deleted?

alshdavid: I'm in favour of deleting them but I'd like to leave that for another PR as there are quite a lot of tests to remove. This PR is mainly focused on fixing typing errors

packages/core/integration-tests/test/worklets.js
alshdavid: Updated

MonicaOlejniczak (approved):

PR #281, opened 2024-12-13

alshdavid:
Adding flow types to tests


packages/core/integration-tests/test/rust.js
alshdavid, line 1: This test file uses deprecated APIs but is skipped, will remove this in future PR

JakeLane (approved):

PR #284, opened 2024-12-13
Switch to using a global property for domain sharding

benjervis:
Originally domain sharding was set up to work by looking for the presence of some specified cookie to decide whether to shard or not, as this is what the CFE SSR server does already.

However that's a little less easy to use, and other products would prefer to use a key on the `window` object instead.

Now, the runtime looks for a boolean value in `window.__ATLASPACK_ENABLE_DOMAIN_SHARDS` as its switch.

SSR servers can inject an inline script tag at the top of the document if sharding is required.

```html
<script>
  window.__ATLASPACK_ENABLE_DOMAIN_SHARDS = true;
</script>
```

Or, if cookie based logic is still preferred
```html
<script>
  window.__ATLASPACK_ENABLE_DOMAIN_SHARDS = document.cookie.indexOf('existing.cookie.name') !== -1;
</script>
```

### DIFF
--- a/packages/runtimes/js/test/bundle-url-shards.test.js
+++ b/packages/runtimes/js/test/bundle-url-shards.test.js
@@ -4,8 +4,6 @@ import {fsFixture, overlayFS, bundle} from '@atlaspack/test-utils';
 
 import {getShardedBundleURL} from '../src/helpers/bundle-url-shards';
 
-const testingCookieName = 'DOMAIN_SHARDING_TEST';
-
 const createErrorStack = (url) => {
   // This error stack is copied from a local dev, with a bunch
   // of lines trimmed off the end so it's not unnecessarily long
@@ -34,13 +32,7 @@ describe('bundle-url-shards helper', () => {
         'https://bundle-shard.assets.example.com/assets/ParentBundle.cba321.js',
       );
 
-      const result = getShardedBundleURL(
-        testBundle,
-        testingCookieName,
-        `${testingCookieName}=1`,
-        5,
-        err,
-      );
+      const result = getShardedBundleURL(testBundle, true, 5, err);
 
       assert.equal(result, 'https://bundle-shard-0.assets.example.com/assets/');
     });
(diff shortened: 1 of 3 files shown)


mattcompiles (approved):

packages/runtimes/js/test/bundle-url-shards.test.js
```diff
@@ -130,7 +109,7 @@ describe('bundle-url-shards helper', () => {
       const code = await overlayFS.readFile(mainBundle.filePath, 'utf-8');
       assert.ok(
         code.includes(
-          `require("449af90f11ccd363").getShardedBundleURL('b.8575baaf.js', '${testingCookieName}', document.cookie, ${maxShards}) + "b.8575baaf.js"`,
+          `require("e73f1ec3075dec82").getShardedBundleURL('b.8575baaf.js', Boolean(window.__ATLASPACK_ENABLE_DOMAIN_SHARDS), ${maxShards}) + "b.8575baaf.js"`,
```
JakeLane, line 112: nit: If you inline `window.__ATLASPACK_ENABLE_DOMAIN_SHARDS` into the helper, bundle size will be smaller (the global var also can't be minified).

JakeLane (approved):

packages/runtimes/js/test/bundle-url-shards.test.js
benjervis: As in do the check inside the function? I wanted to avoid having to pass the whole window object through, mostly for testing purposes

JakeLane: Yeah inside the function you could check `globalThis. __ATLASPACK_ENABLE_DOMAIN_SHARDS` (rather than passing through the window object). Then mock the globalThis object in test.

benjervis: Ok I've pushed a change trying that out

JakeLane (approved):

PR #285, opened 2024-12-16

alshdavid:
Adding flow types to tests


packages/core/integration-tests/test/api.js
alshdavid, line 1: This fixture was being transpiled as flow implicitly. Converted it to commonjs to avoid flow transpilation

yamadapc (approved):

packages/core/integration-tests/test/api.js
MonicaOlejniczak: /nit you could use modern `mjs` instead

MonicaOlejniczak (approved):

PR #286, opened 2024-12-16
WIP Reliability: Enforce type checking on tests

alshdavid:
This PR verifies that tests are using the correct transpiler/type checker for the file type and adds support for tests written in TypeScript

Rules:
|Condition|Engine|
|-|-|
|`.js` `.cjs` `.mjs` | node native |
|`.js` `.jsx` `startsWith('// @flow')` | babel register |
|`.ts` `.mts` `.cts`  | esbuild register |

## Motivation

There were many test files that were being type stripped as Flow files implicitly without being type checked because the files were not annotated with `// @flow`. Some TypeScript files were actually being stripped by Flow and were working because they only used syntax that overlapped.

This allowed tests to be written with invalid JavaScript syntax and invalid Flow types causing subtle drift from core APIs. This resulted in the accumulation of skipped tests that could no longer be run but did not break CI and was also an obstacle to auto-typescript conversion of the codebase. 

This PR aims to to ensure that all tests will have their appropriate type checker run and fail CI if invalid.

### Notes

It would be wise to force explicit file extensions in import & require paths. This is the default in Nodejs with `type: module` which will make the eventual transition easier, is required by Nodejs's built-in TypeScript runner and also makes it a bit easier to deal with the multiple register systems we have.

It might also be a good idea to name Flow files with `*.flow.js` in addition to `// @flow` to help further discriminate the files for the register systems and make it easier to know how many we have

### DIFF
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -3,14 +3,14 @@ const args = process.argv.slice(2);
 
 function isFileArg(file) {
   return (
-    /\.(j|t)s$/.test(file) ||
+    /\.(j|t|cj|mj|ct|mt)s$/.test(file) ||
     (fs.existsSync(file) && fs.statSync(file).isDirectory())
   );
 }
 
 const spec = args.some(isFileArg)
   ? args.filter(isFileArg)
-  : 'packages/*/!(integration-tests)/test/{*.{js,ts},**/*.{test,spec}.{js,ts}}';
+  : 'packages/*/!(integration-tests)/test/{*.{js,ts,cts,mts,cjs,mjs},**/*.{test,spec}.{js,ts,mts,cts,cjs,mjs}}';
 
 module.exports = {
   spec,
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "test:js": "yarn test:js:unit && yarn test:integration",
     "test:js:coverage": "yarn nyc yarn test:js:unit",
     "test:js:coverage:report": "yarn nyc report --reporter=html-spa",
-    "test:js:unit": "cross-env NODE_ENV=test mocha --conditions=development --timeout 5000",
+    "test:js:unit": "cross-env NODE_ENV=test mocha --conditions=\"@atlaspack::sources\" --timeout 5000",
     "test:unit": "yarn test:js:unit && cargo test",
     "dev:release": "SKIP_PLUGIN_COMPATIBILITY_CHECK=true lerna publish -y --canary --preid dev --dist-tag=dev --exact --force-publish=* --no-git-tag-version --no-push",
     "canary:release": "SKIP_PLUGIN_COMPATIBILITY_CHECK=true lerna publish -y --canary --preid canary --dist-tag=canary --exact --force-publish=* --no-git-tag-version --no-push",
--- a/packages/core/integration-tests/test/library-bundler.js
+++ b/packages/core/integration-tests/test/library-bundler.js
@@ -141,7 +141,7 @@ describe.v2('library bundler', function () {
         export {foo, bar} from './foo';
 
       foo.js:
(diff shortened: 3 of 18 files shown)


.mocharc.js
```diff
@@ -3,14 +3,14 @@ const args = process.argv.slice(2);
 
 function isFileArg(file) {
   return (
-    /\.(j|t)s$/.test(file) ||
+    /\.(j|t|cj|mj)s$/.test(file) ||
```
MonicaOlejniczak, line 6: There is also `cts` and `mts` now, might be better to have `\.[cm][jt]s$`

package.json
alshdavid, line 1: We can use `--condition=@atlaspack::sources` to tell node to consume the `src` files rather than `lib` files. The caller needs to use the flow register system we have in Atlaspack - but in future, when the repo is migrated to TypeScript, we can either use

```
NODE_OPTIONS="--conditions=@atlaspack::sources --import tsx" atlaspack build
```

And when Nodejs completes shipping native support for TypeScript, omit the `--import tsx`

alshdavid: This also needs to be namespaced to `@atlaspack::sources` to avoid Nodejs resolving `development` from packages in `node_modules`

packages/core/workers/test/integration/workerfarm/console.js
alshdavid, line 1: This will use `package.json#exports` to determine which import to use

packages/core/workers/test/integration/workerfarm/console.js
```diff
@@ -1,4 +1,4 @@
-const WorkerFarm = require('../../../src/WorkerFarm').default;
+const WorkerFarm = require('#atlaspack/workers').default;
```
alshdavid, line 1: This will use package.json#imports to determine which import to use

packages/core/workers/package.json
```diff
@@ -16,6 +16,20 @@
   "engines": {
     "node": ">= 16.0.0"
   },
+  "exports": {
+    ".": {
+      "types": "./index.d.ts",
```
MonicaOlejniczak, line 21: We shouldn't need this? You can just import at top-level

packages/core/workers/test/integration/workerfarm/console.js
MonicaOlejniczak: Can't you just use `@atlaspack/workers` without the `imports`

alshdavid: I can but the lint rules complain. I guess I could just ignore the linting error lol

packages/core/workers/package.json
alshdavid: It's good to keep things consistent with types given the top level properties are being phased out in place of `exports`. For now I am leaving the top level properties for legacy support but they can/will be removed.

MonicaOlejniczak: Oh right, I misread this a bit too and thought you were exporting `@atlaspack/workers/types` which I wouldn't want.

packages/core/workers/test/integration/workerfarm/console.js
MonicaOlejniczak: Yeah, I'd prefer this

alshdavid: Updated

packages/core/integration-tests/test/library-bundler.js
alshdavid, line 1: Macros are evaluated in Nodejs. Previously the babel register hid failures by transpiling the `import` syntax to cjs. If we decide to keep macros, they should probably be run in a vm or something

MonicaOlejniczak (approved):

JakeLane (approved):

PR #288, opened 2024-12-17
Refactor domain sharding and support bundle-manifest requests

benjervis:
Immediately after landing the domain sharding PR, I changed my mind about how it should work, partially driven by a desire to have the behaviour centralised into a package that can be imported by consuming services.

In this PR:
- The sharding logic has been mostly moved to a new package, `@atlaspack/domain-sharding`
- The sharding function now takes an absolute URL and applies the shards if necessary
- The logic inside the JS Runtime has been reworked to use the new "just give me the whole URL" format
- Sharding has been added to all usages of `getBundleURL` (after I found out that it wasn't being added to scope hoisting builds)


packages/utils/domain-sharding/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "@atlaspack/domain-sharding",
+  "version": "2.12.0",
+  "source": "src/index.js"
```
MonicaOlejniczak, line 4: Is this going to work without a `main` field?

mattcompiles (approved):

PR #296, opened 2024-12-24
Fix unused symbols for dependencies without symbols

JakeLane:
## Motivation

From the issue found by @alshdavid. simple fix to avoid errors and deoptimise if the bundle doesn't support symbols.

## Changes

When a dependency doesn't have symbols, don't error but just deoptimise instead.

## Checklist

### DIFF
--- a/packages/packagers/js/src/ScopeHoistingPackager.js
+++ b/packages/packagers/js/src/ScopeHoistingPackager.js
@@ -1331,7 +1331,7 @@ ${code}
         let unused = incomingDeps.every((d) => {
           let symbols = this.bundleGraph.getUsedSymbols(d);
 
-          // No used symbols avaiable for the asset, make sure we keep all of them
+          // No used symbols available for the asset, make sure we keep all of them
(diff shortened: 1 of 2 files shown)


alshdavid (approved):

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -1329,7 +1329,11 @@ ${code}
         }
 
         let unused = incomingDeps.every((d) => {
-          let symbols = nullthrows(this.bundleGraph.getUsedSymbols(d));
+          let symbols = this.bundleGraph.getUsedSymbols(d);
+
+          // No used symbols avaiable for the asset, make sure we keep all of them
```
MonicaOlejniczak, line 1334: `available`

JakeLane: My spell checker stopped working and I'm instantly pushing typos 😅

PR #298, opened 2025-01-02
V3: Removed indirection from asset graph

alshdavid:
- `AssetNode` and `DependencyNode` stored directly in the `AssetGraph` avoiding the need for external `Vec<_>` and tracking both the `node_index` and `vec_index`
- `AssetGraph` only handle data storage and operation on the data structure are externalized and done at the call site.
- Using `**_idx` naming convention for node indexes in the graph


crates/atlaspack_core/src/asset_graph/propagate_requested_symbols.rs
alshdavid, line 1: Separated this into a dedicated function, removed recursion and removed the closure. I removed the closure because it's annoying to deal with `move` semantics in the closure - particularly as we refactor the request tracker approach in the future.

I noticed this is always called after `graph.add_depenedency`, should this be hidden as part of `add_dependency()`?

crates/atlaspack_core/src/asset_graph/serialize_asset_graph.rs
alshdavid, line 1: This is more a concern for the binding layer rather than of the `AssetGraph` itself. Extracted it into a function, should probably move it into `node_bindings`

crates/atlaspack_core/src/asset_graph/asset_graph.rs
alshdavid, line 1: Trying to keep this data structure focused on being a holder of data a minimize operations on that data. It's just a graph, consumers decide what to do with the graph

crates/atlaspack/src/atlaspack.rs
```diff
@@ -139,16 +139,20 @@ impl Atlaspack {
...
   }
 
-  fn commit_assets(&self, assets: &[AssetNode]) -> anyhow::Result<()> {
+  fn commit_assets(&self, assets: Vec<&AssetNode>) -> anyhow::Result<()> {
     let mut txn = self.db.environment().write_txn()?;
 
-    for AssetNode { asset, .. } in assets.iter() {
-      self.db.put(&mut txn, &asset.id, asset.code.bytes())?;
-      if let Some(map) = &asset.map {
+    for asset_node in assets {
```
MonicaOlejniczak, line 145: The previous destructuring was clearer, does it need to be an owned vec?

crates/atlaspack_core/src/asset_graph/serialize_asset_graph.rs
MonicaOlejniczak: We have a lot of serialisation baked into the core types already, but it can go in either location 🤷

crates/atlaspack/src/requests/asset_graph_request.rs
```diff
@@ -137,16 +138,16 @@ impl AssetGraphBuilder {
   }
 
   fn handle_path_result(&mut self, result: PathRequestOutput, request_id: u64) {
-    let node = *self
-      .request_id_to_dep_node_index
+    let dependency_nx = *self
```
MonicaOlejniczak, line 141: This isn't a channel, why use the `x` convention?

crates/atlaspack/src/requests/asset_graph_request.rs
```diff
@@ -377,20 +384,19 @@ impl AssetGraphBuilder {
...
-          &mut self.request_context,
-          &self.sender,
-          dependency_node_index,
-          dependency,
-        );
-      },
-    );
+    for (dependency_node_index, dependency) in
+      propagate_requested_symbols(&mut self.graph, asset_node_index, incoming_dep_node_index)
+        .unwrap()
```
MonicaOlejniczak, line 389: This should be safer, with a let some

crates/atlaspack/src/requests/asset_graph_request.rs
```diff
@@ -377,20 +384,19 @@ impl AssetGraphBuilder {
...
-          &mut self.work_count,
-          &mut self.request_context,
-          &self.sender,
-          dependency_node_index,
-          dependency,
-        );
-      },
-    );
+    for (dependency_node_index, dependency) in
+      propagate_requested_symbols(&mut self.graph, asset_node_index, incoming_dep_node_index)
```
MonicaOlejniczak, line 388: I'd consider measuring performance of this, since propagation is done a bit differently now

crates/atlaspack/src/requests/asset_graph_request.rs
alshdavid: The idea was that `nx` stood for **n**ode inde**x**

What about `nix` for **n**ode **i**nde**x**

I want to ensure it can't be confused for an `id` because we don't use node indexes as identifiers

crates/atlaspack/src/requests/asset_graph_request.rs
alshdavid: Benchmarked this and it looks like it has the same performance but I reverted back to the callback method to minimize the change.

Rust will be merging [this feature](https://github.com/rust-lang/rfcs/pull/3513) into the language (essentially JavaScript generators) on [Feb 20](https://doc.rust-lang.org/edition-guide/rust-2024/index.html). I'll migrate this function to use that approach which will replicate the original behaviour and make it easier to handle `move` semantics.

crates/atlaspack/src/requests/asset_graph_request.rs
MonicaOlejniczak: I don't personally find the `node` part of the variable name essential, up to you. `nix` is a bit clearer but I think `dependency_idx` is fine too 🤷

crates/atlaspack/src/requests/asset_graph_request.rs
MonicaOlejniczak: Here I thought Rust had generators all along

crates/atlaspack/src/requests/asset_graph_request.rs
yamadapc: you can just call it `dependency_index`

alshdavid: Updated to use `**_index`

crates/atlaspack/src/requests/asset_graph_request.rs
```diff
@@ -691,4 +703,24 @@ mod tests {
...
+        Some(AssetGraphNode::Asset(asset_node)) => {
+          first_asset.replace(asset_node);
+          break;
+        }
+        _ => continue,
+      }
+    }
+
+    first_asset
+  }
```
alshdavid, line 725: The AssetGraph looks like:
```
root -> dependency (n) -> asset (n, entry)
```
This will walk the graph and find the first asset which should be the entry asset

crates/atlaspack/src/requests/asset_graph_request.rs
alshdavid: FWIW, [petgraph](https://docs.rs/petgraph/latest/petgraph/visit/struct.Bfs.html) uses `nx` to describe Node indexes

MonicaOlejniczak (approved):

PR #299, opened 2025-01-03
Example: V3 Determinism solution

alshdavid:
It appears that identifiers are generated based on the insertion order of nodes into the `AssetGraph`. I believe the determinism issues in V3 are a result of the `PathRequest` requests occurring concurrently (I think they run on the main thread in V2 making the order deterministic).  

Having a deterministic insertion position on the `AssetGraph` is one solution (`BTreeMap` based graph?), or alternatively identifiers could be be excluded from identifiers.

In the meantime, a simple fix is to send a sorted graph back to JavaScript - ensuring that insertion order does not matter.

The serialized graph looks like:

```typescript
{
  edges: Array<[number, number]>,   // Sorted alphabetically by `${number} ${number}`,
  nodes: Array<AssetGraphNode>,     // Sorted alphabetically by AssetGraphNode.id()
}
```

It works but it's kinda gross. I'm fine with this approach being scrapped or incorporated temporarily as a stop-gap.


alshdavid:
No longer needed

PR #300, opened 2025-01-06
Replicate the `"main": false` patch from Parcel

benjervis:
Unfortunately, this [ecosystem pollution][pol] has spread to our world as well, and we've inherited Parcel's problem of not parsing the main field as a boolean.

This PR just replicates [Devon's patch][p] into Atlaspack, including his test.

[p]: https://github.com/parcel-bundler/parcel/pull/10053/commits/bb97f0e68e113bf31c7d35f95b385fa33b0c4d20
[pol]: https://github.com/es-shims/dunder-proto/issues/2

### DIFF
--- a/packages/utils/node-resolver-rs/src/package_json.rs
+++ b/packages/utils/node-resolver-rs/src/package_json.rs
@@ -38,12 +38,17 @@ pub struct PackageJson {
   pub path: PathBuf,
   #[serde(default, deserialize_with = "ok_or_default")]
   pub name: String,
+  #[serde(default, deserialize_with = "ok_or_default")]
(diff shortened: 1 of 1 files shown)


MonicaOlejniczak (approved):

yamadapc:
I think we should consider treating `main` specifically rather than adding catch-all `ok_or_default` to the fields.

With this change the following invalid package.json will get parsed:
```
    let pkg: PackageJson = serde_json::from_str(r#"{"name":1234}"#).unwrap();
    assert_eq!(pkg.name, "");
```

`name` will be set to `""` which is not a really semantically valid value, a similar problem will happen for the other fields.

PR #301, opened 2025-01-06
V3: Faster AssetGraph serialization

alshdavid:
This brings a small improvement to the AssetGraph serialization time in JFE

This is done by transferring the AssetGraph to Nodejs as a Buffer.

Due to the string size limitations of JavaScript, the `AssetNode`s are transferred as multiple individual JSON documents that share the same Buffer. You can think of this as kind of like a newline separated list of JSON objects. 

JavaScript then loops over the buffer, splits it on each document and parses each document individually. For JavaScript to know where to split the buffer to seperate each document, each JSON body is headered with its byte length (u32 as little endian binary)

So the buffer is encoded sorta like this:
```javascript
const encoded = toBinary([
  assetNode1Bytes.length, ...assetNode1Bytes,
  assetNode2Bytes.length, ...assetNode2Bytes,
  assetNode3Bytes.length, ...assetNode3Bytes,
  // ... and so on
])
```


crates/atlaspack_core/src/asset_graph.rs
```diff
@@ -631,100 +626,4 @@ mod tests {
     // "*" should be marked as requested on the stuff dep
     assert_requested_symbols(&graph, stuff_dep, vec!["*"]);
   }
-
-  #[test]
-  fn serialize_nodes_handles_max_size() -> anyhow::Result<()> {
```
alshdavid, line 636: Max buffer size is 4gb, the serialized asset graph in JFE isn't even close to 500mb. No longer a need to test the max size

MonicaOlejniczak: You should still test the function works

alshdavid:
Will come back to this later, closing to reduce the noise

PR #302, opened 2025-01-07
Chore: Rust 1.83.0 & Lint Fix

alshdavid:
- Bump the version of Rust
- Run lint fixer (`cargo clippy --fix`)
- Updated CI to use new Rust version


MonicaOlejniczak:
@alshdavid Can you update the ci and release workflow to use https://github.com/actions-rust-lang/setup-rust-toolchain? It will read from the toolchain file unlike the current action, which does not intend to support that feature

MonicaOlejniczak (approved):

yamadapc (approved):

PR #305, opened 2025-01-08
Added import retries into packager

alshdavid:
The previous import retry feature gave the consumer the ability to retry failed requests. This required the consumer to handle the actual retry logic:

```javascript
for (let i = 0; i < 5; i++) {
  try {
    await import('./foo')
  } catch(err) {
    await new Promise(res => setTimeout(res, 1000))
  }
}
```

This PR added the retry logic into the JavaScript runtime, enabled with `--feature-flag importRetry=true`

```javascript
await import('./foo') // will auto retry on failure
```

For reporting purposes, it will also emit a custom event on the window to notify when retries were triggered

```javascript
window.addEventListener('atlaspack:import_retry', ({detail}) => console.log(detail));
```

### DIFF
--- a/packages/runtimes/js/src/helpers/browser/esm-js-loader-retry.js
+++ b/packages/runtimes/js/src/helpers/browser/esm-js-loader-retry.js
@@ -1,25 +1,41 @@
 async function load(id) {
-  if (!parcelRequire.retryState) {
-    parcelRequire.retryState = {};
-  }
+  const url = require('../bundle-manifest').resolve(id);
 
-  if (!globalThis.navigator.onLine) {
-    await new Promise((res) =>
-      globalThis.addEventListener('online', res, {once: true}),
-    );
-  }
+  const maxRetries = 6;
 
-  let url = require('../bundle-manifest').resolve(id);
+  for (let i = 1; i <= maxRetries; i++) {
+    // Wait for the user to go online before making a request
+    if (!globalThis.navigator.onLine) {
+      await new Promise((resolve) =>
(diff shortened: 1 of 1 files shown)


packages/runtimes/js/src/helpers/browser/esm-js-loader-retry.js
```diff
@@ -1,25 +1,45 @@
...
+    try {
+      // eslint-disable-next-line no-undef
+      const module = await __parcel__import__(requestUrl);
+      if (i !== 0) {
+        importRetryEvent({result: 'success', target: url, attempts: i});
+      }
+      return module;
+    } catch (error) {
+      if (i === maxRetries) {
+        importRetryEvent({result: 'failure', target: url, attempts: i});
```
yamadapc, line 34: Instead of doing this ; log each retry.
We already have the data for failures / successes (the page-loads and failed to import exceptions).

packages/runtimes/js/src/helpers/browser/esm-js-loader-retry.js
```diff
@@ -1,25 +1,45 @@
...
+        throw error;
+      }
+      const jitter = Math.round(Math.random() * 100);
+      const multiple = i + 1;
+      await new Promise((resolve) =>
+        setTimeout(resolve, (baseTimeoutDuration + jitter) * multiple),
```
yamadapc, line 40: In my opinion it's easier to just use exponential back-off off the bat with some maximum wait time limit.

yamadapc (approved):

PR #307, opened 2025-01-13
Dev: Enable the use of "exports" when resolving bundler plugins

alshdavid:
This PR allows the bundler to select plugins and internals using Nodejs's [conditional exports](https://nodejs.org/api/packages.html#conditional-exports)

This will help with the migration path to TypeScript as it allows us to select sources using built-in Nodejs conditional resolution (and builtin nodejs TypeScript support)

```bash
NODE_OPTIONS="--condition=@atlaspack::sources" atlaspack build

node --condition=@atlaspack::sources ./node_modules/.bin/atlaspack build
```

Where Atlaspack packages will be ported to

```json
{
  "name": "@atlaspack/foo",
  "exports": {
    "@atlaspack::sources": "./src/index.ts",
    "types": "./src/index.ts",
    "default": "./lib/index.js",
   }
}
```

### DIFF
--- a/packages/core/package-manager/src/nodejsConditions.js
+++ b/packages/core/package-manager/src/nodejsConditions.js
@@ -2,10 +2,11 @@
 import process from 'process';
 
 // https://nodejs.org/api/packages.html#conditional-exports
+// TODO We don't support { "type": "module" }
 export const defaultNodejsConditions: Array<string> = Object.freeze([
   'node-addons',
   'node',
-  'import',
+  // 'import',
   'require',
   'module-sync',
   'default',
--- a/packages/core/package-manager/src/nodejsConditions.js
+++ b/packages/core/package-manager/src/nodejsConditions.js
@@ -1,0 +1,34 @@
+// @flow
+import process from 'process';
+
+// https://nodejs.org/api/packages.html#conditional-exports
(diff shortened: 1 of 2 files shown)


packages/core/package-manager/src/nodejsConditions.js
```diff
@@ -0,0 +1,35 @@
...
+let envConditions: void | Array<string> = undefined;
+
+/** @description Gets the export conditions from NODE_OPTIONS and node arguments */
+export function getConditionsFromEnv(): Array<string> {
+  if (!envConditions) {
+    const conditions = [];
+
+    for (const arg of [
+      ...process.execArgv,
+      ...(process.env.NODE_OPTIONS || '').split(' '),
```
MonicaOlejniczak, line 24: node options isn't valid for setting conditions, I'd remove it

alshdavid: Where did you find that? It seems to work when I test it:

image

MonicaOlejniczak: Oh I was thinking of something else, all good! Nice

MonicaOlejniczak (approved):

PR #308, opened 2025-01-17
Fix: Import retries - simultaneous imports

alshdavid:
Expanded the import retries logic to include a global cache to avoid issues when multiple imports for the same asset occur simultaneously. Example:
```javascript
Promise.all([
  import('./foo'),
  import('./foo'),
])
```

### DIFF
--- a/packages/runtimes/js/src/helpers/browser/esm-js-loader-retry.js
+++ b/packages/runtimes/js/src/helpers/browser/esm-js-loader-retry.js
@@ -1,42 +1,61 @@
 async function load(id) {
-  const url = require('../bundle-manifest').resolve(id);
+  // Global state maps the initial url to the completed task.
+  // This ensures the same URL is used in subsequent imports
+  if (!parcelRequire._retryState) parcelRequire._retryState = {};
+  /** @type {Record<string, Promise<void>>} */
+  const retryState = parcelRequire._retryState;
 
+  // The number of retries before rethrowing the error
   const maxRetries = 6;
 
-  for (let i = 1; i <= maxRetries; i++) {
-    // Wait for the user to go online before making a request
-    if (!globalThis.navigator.onLine) {
-      await new Promise((resolve) =>
-        globalThis.addEventListener('online', resolve, {once: true}),
-      );
-    }
+  // Resolve the request URL from the bundle ID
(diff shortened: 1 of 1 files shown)


packages/runtimes/js/src/helpers/browser/esm-js-loader-retry.js
```diff
@@ -1,42 +1,61 @@
...
+  // then start the import retry task. Otherwise reuse the
+  // existing result or wait for the current task to complete
+  if (!retryState[url]) {
+    retryState[url] = (async () => {
+      // Try first request with normal import circuit
+      try {
+        // eslint-disable-next-line no-undef
+        return await __parcel__import__(url);
+      } catch {
+        /**/
```
mattcompiles, line 31: Doesn't this mean it'll always retry as we ignore the error?

mattcompiles: Ohh ok, I think I get what you're going for. How come you're not using the catch handler here? Reads a bit confusing.

mattcompiles (approved):
It would be great to have some tests for this soon.

packages/runtimes/js/src/helpers/browser/esm-js-loader-retry.js
alshdavid: This logic will only run once the first time `load` is called as the promise is cached in `retryState[url]` which is reused on subsequent calls to `load`

The flow goes:

- `retryState[url]` is immediately set to a `Promise<void>` defined as the import task
  - subsequent requests will get `retryState[url]` which will be the pending import
  
- The first import attempt uses the normal import logic
  - If the first import fails it begins retrying
    - if a retry attempt succeeds then `retryState[url]` resolves
    - if all retry attempts fail then ``retryState[url]` rejects

alshdavid: > How come you're not using the catch handler here? Reads a bit confusing.

Being a bit defensive about when retries start to occur. I wanted the original retry logic to exist as it did before before the retry circuit is kicked off.

I had it as part of the loop before which is fine too - just wanted to seperate it to be a little more explicit. Can refactor this later after I add some tests

PR #312, opened 2025-01-21
DRAFT: Using link tag for import retry

alshdavid:
This is an alternative approach to import retries. May be required due to SSR servers having issues with querystring parameters


alshdavid:
No longer needed as the original implementation works

PR #313, opened 2025-01-22
V3: Using JsBuffer to transfer AssetGraph to Nodejs

alshdavid:
To help with the determinism issue I am experimenting with changes to the serialization of the AssetGraph. This PR simplifies the transferral of the AssetGraph from `JsString` values to `JsBuffer` to make it easier to work with when experimenting with changes to the serialization approach.

It performs pretty much identically (slightly faster but not meaningfully) however does not have the same string size limitations and avoids the need to manually create a JSON string from Rust.


crates/node-bindings/src/atlaspack/serialize_asset_graph.rs
```diff
@@ -0,0 +1,66 @@
...
+        value: SerializedDependency {
+          id: dependency_node.dependency.id(),
+          dependency: dependency_node.dependency.as_ref().clone(),
+        },
+        has_deferred: dependency_node.state == DependencyState::Deferred,
+      },
+    };
+
+    let node_bytes = serde_json::to_vec(&serialized_node)?;
+    let js_buffer = env.create_buffer_with_data(node_bytes)?;
```
yamadapc, line 55: why do we do this instead of just returning an object?

if this is better why not serialise the entire thing as JSON?

crates/atlaspack_core/src/asset_graph/serialize_asset_graph.rs
```diff
@@ -1,172 +0,0 @@
...
-  // Add the current node if it did not overflow in size
-  if curr_node.len() < (max_str_len - 3) {
-    nodes.push(format!("[{curr_node}]"));
-  }
-
-  Ok(nodes)
-}
-
-#[cfg(test)]
-mod tests {
```
yamadapc, line 73: Deleted tests

alshdavid: These tests were checking the max string length limitation of Nodejs was respected by the manual creation of the original `JsString` in Rust.

Because the new function uses `Env` directly it cannot be unit tested without a Nodejs environment. This circuit is covered by the Nodejs integration tests in v3 because failure to transfer the AssetGraph would result in a critical failure

crates/node-bindings/src/atlaspack/serialize_asset_graph.rs
alshdavid: > why do we do this instead of just returning an object?

If you're referring to a `napi_derive::napi(object)` it's because the macros are broken in the IDE (though the build will pass). We can either move to napi_v3 (prerelease) or manually cast the properties to an object.

> if this is better why not serialise the entire thing as JSON?

Returning a single JSON exceeds the max string length limit in Nodejs. Having each node be its own json allows for the total JSON size to be broken up into smaller pieces. Using buffers isn't necessary but allows the values to be moved rather than copied

yamadapc (approved):

PR #315, opened 2025-01-24
Analytics events helper in JS Runtime

alshdavid:
This allows forwarding runtime events in JavaScript runtimes


packages/runtimes/js/src/helpers/browser/esm-js-loader-retry.js
```diff
@@ -58,4 +59,16 @@ async function load(id) {
   return retryState[url];
 }
 
+function sendAnalyticsEvent(action, targetUrl, attempt) {
+  require('./analytics/analytics.js').sendAnalyticsEvent({
+    kind: 'operational',
+    action: 'importRetry',
+    attributes: {
+      action,
```
yamadapc, line 67: `action` is duplicated?

packages/runtimes/js/src/helpers/browser/analytics/analytics.d.ts
```diff
@@ -0,0 +1,8 @@
+export type AtlaspackAnalyticsEvent = {
+  kind?: string;
```
yamadapc, line 2: Remove `kind` / `tags`

packages/runtimes/js/src/helpers/browser/esm-js-loader-retry.js
yamadapc: Use the `action` field instead of another one

packages/runtimes/js/src/helpers/browser/analytics/analytics.d.ts
alshdavid: Updated

yamadapc (approved):

PR #326, opened 2025-02-05
Add dominance algorithm implementation

yamadapc:
Implements "A simple, fast dominance algorithm", to find the immediate
dominators of all nodes in a graph.

- https://www.cs.tufts.edu/comp/150FP/archive/keith-cooper/dom14.pdf

This will be used for a new experimental bundler package to be added.
This adds the empty package.

Test Plan: Run `yarn test` to execute unit-tests for the implementation


packages/bundlers/bundler-experimental/test/DominatorBundler/findAssetDominators/simpleFastDominance.test.js
```diff
@@ -0,0 +1,126 @@
+// @flow strict-local
+
+import assert from 'assert';
+import {ContentGraph} from '@atlaspack/graph';
+import {simpleFastDominance} from '../../../src/DominatorBundler/findAssetDominators/simpleFastDominance';
```
MonicaOlejniczak, line 5: Any reason you're using a test folder instead of placing it with the source?

packages/bundlers/bundler-experimental/test/DominatorBundler/findAssetDominators/simpleFastDominance.test.js
```diff
@@ -0,0 +1,126 @@
...
+
+const baseGraph = () => {
+  const inputGraph = new ContentGraph();
+  const root = inputGraph.addNodeByContentKey('root', 'root');
```
MonicaOlejniczak, line 9: These tests would be better if you had different values for the content key and the node, to ensure what you get back is what you expect

packages/bundlers/bundler-experimental/test/DominatorBundler/findAssetDominators/simpleFastDominance.test.js
yamadapc: What do you mean? 

Do you mean using

* /src
* /test

?

That's just how the repository is setup. I'd rather colocate the tests.

MonicaOlejniczak (approved):

packages/bundlers/bundler-experimental/test/DominatorBundler/findAssetDominators/simpleFastDominance.test.js
MonicaOlejniczak: Yeah, colocation. Be the change...

packages/bundlers/bundler-experimental/test/DominatorBundler/findAssetDominators/simpleFastDominance.test.js
yamadapc: I've changed to use a normal graph; content key is not relevant on these

MonicaOlejniczak: Perfect

packages/bundlers/bundler-experimental/test/DominatorBundler/findAssetDominators/simpleFastDominance.test.js
yamadapc: I think we could do that for the entire repository separately (?)

PR #328, opened 2025-02-05
additional fixes after testing in product

pancaspe87:
Following previous [PR](https://github.com/atlassian-labs/atlaspack/pull/321), we are now extending option type to also include the build event. This will allow us to track other data such as the duration of the build


yamadapc (changes requested):
There's no need for this change

It doesn't make sense to pass in a value into a construct then read it back like this, also it shouldn't be optional

PR #329, opened 2025-02-05
Add graph that can store edge weights

yamadapc:
Add a content graph class that can store edge weights, rather than just node weights

Test Plan: yarn test


packages/bundlers/bundler-experimental/test/DominatorBundler/EdgeContentGraph.test.js
```diff
@@ -0,0 +1,58 @@
...
+      const weight2 = 20;
+
+      graph.addWeightedEdge(a, b, edgeType, weight1);
+      graph.addWeightedEdge(b, c, edgeType, weight2);
+
+      const clonedGraph = graph.clone();
+      assert.deepEqual(clonedGraph.nodes, graph.nodes);
+      assert.deepEqual(clonedGraph.getAllEdges(), graph.getAllEdges());
+      assert.equal(clonedGraph.getEdgeWeight(a, b), weight1);
+      assert.equal(clonedGraph.getEdgeWeight(b, c), weight2);
```
MonicaOlejniczak, line 55: maybe assert that `clonedGraph` is not the same instance as `graph` and that adding to `clonedGraph` does not impact `graph`

MonicaOlejniczak (approved):

PR #330, opened 2025-02-05
Make 'removeOrphans' optional on replaceNodeIdsConnectedTo

yamadapc:
The graph clean-up is a traversal, so there's a large linear performance hit if
we clean-up on every modification to the graph. This makes it configurable to
disable clean-up on replacement.

Test Plan: `yarn mocha src/packages/core/graph`


packages/core/graph/test/Graph.test.js
```diff
@@ -274,26 +267,63 @@ describe('Graph', () => {
...
 
-    assert(graph.hasNode(nodeA));
-    assert(graph.hasNode(nodeB));
-    assert(!graph.hasNode(nodeC));
-    assert(graph.hasNode(nodeD));
-    assert.deepEqual(Array.from(graph.getAllEdges()), [
-      {from: nodeA, to: nodeB, type: 1},
-      {from: nodeA, to: nodeD, type: 1},
-    ]);
+    it('replaceNodeIdsConnectedTo does not remove orphan nodes if flag is fals', () => {
```
MonicaOlejniczak, line 300: false

MonicaOlejniczak (approved):

PR #332, opened 2025-02-06
Add Tarjan SCC algorithm implementation

yamadapc:
Adds an implementation of Tarjan's algorithm to find strongly connected
components.

This will be used on the experimental bundler to simplify cycle handling and
can be used on other parts of the codebase to simplify the symbol propagation
implementation.

On the bundler, we will use this to convert the graph to an acyclic graph
(replace the strongly connected components with a single 'group' node) before
doing other work.

Test Plan: yarn test

### DIFF
--- a/packages/bundlers/bundler-experimental/src/DominatorBundler/cycleBreaker/findStronglyConnectedComponents.js
+++ b/packages/bundlers/bundler-experimental/src/DominatorBundler/cycleBreaker/findStronglyConnectedComponents.js
@@ -1,0 +1,79 @@
+// @flow strict-local
+
+import type {Graph, NodeId} from '@atlaspack/graph';
+
+export type StronglyConnectedComponent = NodeId[];
+
+/**
+ * Robert Tarjan's algorithm to find strongly connected components in a graph.
+ *
+ * Time complexity: O(V + E)
+ * Space complexity (worst case): O(V)
+ *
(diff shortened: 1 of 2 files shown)


JakeLane (approved):

packages/bundlers/bundler-experimental/src/DominatorBundler/cycleBreaker/findStronglyConnectedComponents.js
```diff
@@ -0,0 +1,79 @@
...
+): StronglyConnectedComponent[] {
+  type State = {|
+    index: null | NodeId,
+    lowlink: null | NodeId,
+    onStack: boolean,
+  |};
+  const result = [];
+  let index = 0;
+  const stack = [];
+  const state: State[] = new Array(graph.nodes.length).fill(null).map(() => ({
```
MonicaOlejniczak, line 27: fill isn't necessary?

yamadapc: It's necessary because the array returned from `new Array(length)` will be sparse https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Indexed_collections#sparse_arrays

MonicaOlejniczak: Ah right, I was mixing it up with this syntax which is a bit more succinct

```javascript
Array.from({ length: 10 }, () => ({
    index: null,
    lowlink: null,
    onStack: false,
  }))
```

PR #333, opened 2025-02-06
Add 'IdentifierRegistry' mechanism for debugging IDs

yamadapc:
Whenever an ID is generated, we will optionally log it to a file with the data
used to generate it.

This will help us debug generated identifiers on artefacts or runtime errors.

Test Plan: N/A


packages/core/core/src/IdentifierRegistry.js
```diff
@@ -0,0 +1,20 @@
...
+
+import fs from 'fs';
+
+export class IdentifierRegistry {
+  constructor() {}
+
+  addIdentifier(type: string, identifier: string, data: mixed) {
+    if (process.env.ATLASPACK_IDENTIFIER_DEBUG === 'true') {
+      fs.appendFile(
+        './atlaspack-identifiers.txt',
```
MonicaOlejniczak, line 11: Can we have a better location for parcel related things, the top-level `.parcel_supervisor_minidumper` files are really noisy for instance. I would prefer a new folder like `.parcel-cache` but maybe `.atlaspack` or something

yamadapc: Hey, those files aren't generated in this repository.

Given this is just a debugging file, it should be okay to write it to CWD.

If you want to store them in a new folder, then you want me to create `.atlaspack` and write just this file into that?

MonicaOlejniczak: Yes

yamadapc: I've changed it to write to `.atlaspack`.

MonicaOlejniczak (approved):

PR #336, opened 2025-02-11
Add (tracing) metrics around VCS calls

yamadapc:
Test Plan: N/A


packages/core/core/src/requests/AssetGraphRequestRust.js
```diff
@@ -4,6 +4,7 @@ import invariant from 'assert';
 
 import ThrowableDiagnostic from '@atlaspack/diagnostic';
 import type {Async} from '@atlaspack/types';
+import {instrument} from '@atlaspack/logger';
```
MonicaOlejniczak, line 7: How come you opted for re-using the logger package as opposed to a tracing package? They seem discrete enough to me, but let me know what you think.

yamadapc: Hey, I thought because `tracing::...` is just one back-end (generally), and we will likely want to integrate JS and Rust logging/tracing back-ends, we could use the same package for now.

pancaspe87 (approved):

PR #339, opened 2025-02-13
Update cache invalidation to report build type

yamadapc:
We would want to track the initial cache invalidation differently from other
cache invalidation events.

Test Plan: N/A


.changeset/blue-jeans-double.md
```diff
@@ -0,0 +1,133 @@
...
+'@atlaspack/swc-plugin-contextual-imports': minor
+'@atlaspack/create-react-app': minor
+'@atlaspack/domain-sharding': minor
+'@atlaspack/events': minor
+'@atlaspack/macros': minor
+'@atlaspack/node-resolver-core': minor
+'@atlaspack/service-worker': minor
+'@atlaspack/ts-utils': minor
+'@atlaspack/validator-eslint': minor
+'@atlaspack/validator-typescript': minor
```
mattcompiles, line 130: How come we're bumping every package? Can't we just release core?

yamadapc: I mentioned this earlier, I'm not sure but I reckon also changes to core do impact everything

MonicaOlejniczak: Changesets should take care of it if core is listed as an explicit dependency

yamadapc: Hey, I believe changesets will bump all packages because they have explicit dependencies with one another and core is depended on by everything.

JakeLane (approved):

PR #340, opened 2025-02-14
Fix error masking when errors occur in the bundle graph request

mattcompiles:
## Motivation

Prior to this change, if an error occurred during parts of the bundle graph request, the error would be hidden due to the `finally` block throwing an error attempting to create a graphviz file for the graph. This is optional behaviour, exclusively used for internal debugging so it's safe to opt out of. 

## Changes

- Ensure bundle graph request errors show up correctly
- Add skipped test for discovered broken `BundleGraph.fromAssetGraph` edge case

## Checklist


packages/core/integration-tests/test/scope-hoisting.js
```diff
@@ -6455,4 +6455,40 @@ describe('scope hoisting', function () {
       },
     ]);
   });
+
+  // This is currently failing due to dependency.sourceAssetId becoming stale
```
yamadapc, line 6459: it started failing with this change?

yamadapc (approved):

packages/core/integration-tests/test/scope-hoisting.js
mattcompiles: No, I discovered it from a support request. I'm making a ticket to track the fix separately but thought it was worth committing the repro

PR #343, opened 2025-02-17
Improve performance of VCS diff listing

yamadapc:
Improves performance of listing yarn lock files and VCS changes by:

* changing exclude patterns
* using rayon

Test Plan: cargo test

### DIFF
--- a/crates/atlaspack_vcs/src/lib.rs
+++ b/crates/atlaspack_vcs/src/lib.rs
@@ -114,6 +114,7 @@
 
 use anyhow::anyhow;
 use git2::{DiffOptions, Repository};
+use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use std::{
   hash::{Hash, Hasher},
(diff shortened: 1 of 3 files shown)


crates/atlaspack_vcs/src/lib.rs
```diff
@@ -148,11 +149,15 @@ impl VCSState {
     exclude_patterns: &[String],
     failure_mode: FailureMode,
   ) -> anyhow::Result<VCSState> {
+    tracing::info!("Reading VCS state");
```
pancaspe87, line 152: for my understanding, why do we refer the "full path" rather than doing `use tracing::{debug, info, error};` so that we can avoid repeating `tracing::` ?

crates/atlaspack_vcs/src/lib.rs
```diff
@@ -324,15 +332,18 @@ pub fn get_changed_files(
   Ok(changed_files)
 }
 
-pub fn vcs_list_files(dir: &Path) -> Result<Vec<String>, anyhow::Error> {
+pub fn vcs_list_yarn_lock_files(dir: &Path) -> Result<Vec<String>, anyhow::Error> {
   let mut command = Command::new("git");
   command
     .arg("ls-files")
+    .arg("--cached")
```
pancaspe87, line 339: Is it how we are going to report also staged files?

pancaspe87 (approved):

crates/atlaspack_vcs/src/lib.rs
yamadapc: No, this is just to list the yarn.lock files committed

crates/atlaspack_vcs/src/lib.rs
yamadapc: I haven't thought too much about it but I generally have used the full path because `error`, `info`, etc. are common names that may be shadowed.

crates/atlaspack_vcs/src/lib.rs
pancaspe87: Are we dealing with staged files in a follow up PR?

The command in line 366 will report only unstaged

PR #344, opened 2025-02-18
Add tracing logs around git dirty hashes listing

pancaspe87:
Tested locally by running `cargo run --example vcs_file_diff snapshot --repository-root="../../"`

Screenshot 2025-02-18 at 11 15 53 AM

### DIFF
--- a/crates/atlaspack_vcs/src/lib.rs
+++ b/crates/atlaspack_vcs/src/lib.rs
@@ -139,7 +139,7 @@ pub mod yarn_integration;
 pub struct VCSState {
   pub git_hash: String,
   pub dirty_files: Vec<VCSFile>,
-  pub dirty_files_execution_time: u64,
+  pub dirty_files_execution_time: u32,
   pub yarn_states: Vec<YarnSnapshot>,
 }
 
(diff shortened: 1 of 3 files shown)


crates/atlaspack_vcs/src/lib.rs
```diff
@@ -138,30 +139,46 @@ pub mod yarn_integration;
...
   /// Read the VCS state from a repository root. Ignore dirty files matching
   /// the exclude patterns.
+  #[tracing::instrument(level = "info", skip_all)]
   pub fn read_from_repository(
     path: &Path,
     exclude_patterns: &[String],
     failure_mode: FailureMode,
   ) -> anyhow::Result<VCSState> {
     tracing::info!("Reading VCS state");
+    let span = tracing::span!(tracing::Level::INFO, "vcs_dirty_files_duration");
```
yamadapc, line 156: You could just `#[tracing::instrument]` the `vcs_list_dirty_files` function instead, which would do pretty much this code (but also end the span when the function is over.

yamadapc (approved):

PR #345, opened 2025-02-18
Support specific event types from vcs fs

yamadapc:
Adds support for VCS change lists to differentiate between create/delete/update
events.

Test Plan: cargo test

### DIFF
--- a/crates/atlaspack_vcs/src/lib.rs
+++ b/crates/atlaspack_vcs/src/lib.rs
@@ -266,6 +266,14 @@ impl FileChangeEvent {
   pub fn change_type(&self) -> &FileChangeType {
     &self.change_type
   }
+
+  pub fn change_type_str(&self) -> &str {
+    match self.change_type() {
+      atlaspack_vcs::FileChangeType::Create => "create",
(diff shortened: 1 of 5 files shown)


crates/atlaspack_vcs/src/lib.rs
```diff
@@ -274,8 +299,42 @@ pub fn get_changed_files(
...
+          changed_files.push(FileChangeEvent {
+            path: new_file_path,
+            change_type: FileChangeType::Create,
+          });
+          return true;
+        }
+
+        changed_files.push(FileChangeEvent {
+          path: new_file_path,
+          change_type: match status {
```
pancaspe87, line 324: Do we need all of these types for our solution? Do they affect how we bundle things?

pancaspe87 (approved):

crates/atlaspack_vcs/src/lib.rs
yamadapc: They do at the moment, but we might tweak the invalidation routine a bit to handle full node_modules sub-directory invalidation

PR #347, opened 2025-02-19
Helper dev scripts

alshdavid:
A script with a few commands I run frequently

```
./scripts/dev build
./scripts/dev watch
```


yamadapc:
Shouldn't we put in package.json though?tttttt

alshdavid:
> Shouldn't we put these in package.json though?

I find the `package.json#scripts` not to be the best place to house multi-step, multi-line or cross platform build tasks. Personally I'd prefer it if we used a task runner like [just](https://github.com/casey/just) or make. Could also just use bash but that has cross-platform problems

PR #350, opened 2025-02-20
Fix case where lockfiles are added into a repo

yamadapc:
If a lock-file is added into the repository then we will consider all
dependency paths changed.

Test Plan: cargo build

### DIFF
--- a/crates/atlaspack_vcs/src/yarn_integration/mod.rs
+++ b/crates/atlaspack_vcs/src/yarn_integration/mod.rs
@@ -39,6 +39,10 @@ impl YarnResolution {
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(transparent)]
 pub struct YarnLock {
+  /// Map of 'requirement' to 'resolution'
+  ///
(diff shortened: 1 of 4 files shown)


crates/atlaspack_vcs/src/yarn_integration/mod.rs
```diff
@@ -39,6 +39,10 @@ impl YarnResolution {
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(transparent)]
 pub struct YarnLock {
+  /// Map of 'requirement' to 'resolution'
```
pancaspe87, line 42: Should we remove this? I can do it in a follow up PR

pancaspe87 (approved):

crates/atlaspack_vcs/src/yarn_integration/mod.rs
yamadapc: It can serve as documentation but we can remove it too or write a clearer description

PR #351, opened 2025-02-20
Fix initial change listing with statuses

yamadapc:
Fix VCS events since listing to include currently untracked files.

Test Plan: cargo test

### DIFF
--- a/crates/atlaspack_vcs/src/lib.rs
+++ b/crates/atlaspack_vcs/src/lib.rs
@@ -309,15 +309,41 @@ pub fn get_changed_files_from_git(
   old_commit: &git2::Commit<'_>,
   new_commit: &git2::Commit<'_>,
 ) -> anyhow::Result<Vec<FileChangeEvent>> {
+  let mut changed_files = Vec::new();
+
+  // list current dirty files
+  tracing::debug!("Listing dirty files");
+  let mut status_options = git2::StatusOptions::new();
+
+  status_options.include_ignored(false);
+  status_options.include_untracked(true);
+  status_options.include_unmodified(false);
+
+  let statuses = repo.statuses(Some(&mut status_options))?;
+  statuses.iter().for_each(|entry| {
+    let path = entry.path().unwrap();
+    let status = entry.status();
+    let mut change_type = FileChangeType::Update;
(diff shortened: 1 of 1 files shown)


crates/atlaspack_vcs/src/lib.rs
```diff
@@ -309,15 +309,41 @@ pub fn get_changed_files_from_git(
...
+  status_options.include_untracked(true);
+  status_options.include_unmodified(false);
+
+  let statuses = repo.statuses(Some(&mut status_options))?;
+  statuses.iter().for_each(|entry| {
+    let path = entry.path().unwrap();
+    let status = entry.status();
+    let mut change_type = FileChangeType::Update;
+    if status.is_wt_deleted() {
+      change_type = FileChangeType::Delete;
```
pancaspe87, line 328: Is there a difference in the way we handle this status type in cache invalidation?
Also, if the files is untracked, do they have all the different statuses?

crates/atlaspack_vcs/src/lib.rs
```diff
@@ -309,15 +309,41 @@ pub fn get_changed_files_from_git(
   old_commit: &git2::Commit<'_>,
   new_commit: &git2::Commit<'_>,
 ) -> anyhow::Result<Vec<FileChangeEvent>> {
+  let mut changed_files = Vec::new();
+
+  // list current dirty files
```
pancaspe87, line 314: what's the difference with `vcs_list_dirty_files` and why do we need this?

pancaspe87 (approved):

JakeLane (approved):

crates/atlaspack_vcs/src/lib.rs
yamadapc: There is, the invalidations are registered against specific file change events (update/create/delete)

crates/atlaspack_vcs/src/lib.rs
yamadapc: We will unify but this is doing slightly different work ;

There's a snapshot side and a start-up side. We actually needed to reconcile the snapshot with the dirty files here and re-hash the contents.

PR #352, opened 2025-02-21
Issue/vcs toggle vcs modes

pancaspe87:
We want to implement a consistency check pattern to enable VCS modes under a feature flag. There are 4 different scenarios that will be controlled by combining a 'killswitch', already implemented in product,  and a flag to toggle the relevant mode, as explains below:

Screenshot 2025-02-21 at 2 13 32 PM

The first 2 modes are already available in the products and are switched on/off by an infra ff. This PR introduces the ability to return different values for the last 2 modes through another infra ff.

### DIFF
--- a/packages/core/feature-flags/src/index.js
+++ b/packages/core/feature-flags/src/index.js
@@ -14,6 +14,7 @@ export const DEFAULT_FEATURE_FLAGS: FeatureFlags = {
   fixQuadraticCacheInvalidation: 'OLD',
   useLmdbJsLite: false,
   conditionalBundlingApi: false,
+  vcsMode: 'OLD',
 };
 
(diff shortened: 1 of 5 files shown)


packages/core/feature-flags/src/index.js
```diff
@@ -27,6 +28,12 @@ export function getFeatureFlag(flagName: $Keys<FeatureFlags>): boolean {
   return value === true || value === 'NEW';
 }
 
+export function getFeatureFlagValue(
```
pancaspe87, line 31: @yamadapc we can't use the existing `getFeatureFlag` as we need to differentiate between `NEW_AND_CHECK` and `NEW`

yamadapc (approved):

PR #353, opened 2025-02-22
Experiment: @atlassian/atlaspack package unifying everything in one package

alshdavid:
This is just a little weekend experiment. I created the `@atlassian/atlaspack` package which combines all of the packages in the repo and distributes them under a single package `@atlassian/atlaspack`.

_One package to rule them all_

# Usage

```typescript
import { Atlaspack } from '@atlassian/atlaspack'
import { Resolver } from '@atlassian/atlaspack/plugin'
import { FileSystem } from '@atlassian/atlaspack/fs'
import { Asset } from '@atlassian/atlaspack/types'

// Currently only @atlaspack/core is re-exported on the top level
// I could re-export other packages on the top level too as there are no collisions

import { Atlaspack, Resolver, FileSystem } from '@atlassian/atlaspack'
```

# Try it out

```
npm install @alshdavid/atlaspack
npx atlaspack build
```

## Motivation

- I was curious to see if there was a better way of managing the upgrade path for consumers
- Perhaps this could lead to a better approach to compiling Atlaspack 
- Distribute Atlaspack as a single file executable?

## Changes

- Added support to the `PackageManager` for resolving `package.json#import` statements [docs](https://nodejs.org/api/packages.html#imports)
- Added the `@atlassian/atlaspack` package
  - Includes `core`
  - Includes all configs/plugins/runtimes/bundlers/etc

## Building

Requires Node 22+ to build and Node 14+ to run

```bash
yarn build-native-release
yarn build
yarn build-ts
yarn build-unified
```


packages/atlaspack/package.json
```diff
@@ -0,0 +1,113 @@
...
+    "#atlaspack/reporter/*": "./reporters/*/index.js",
+    "#atlaspack/resolver/*": "./resolvers/*/index.js",
+    "#atlaspack/runtime/*": "./runtimes/*/index.js",
+    "#atlaspack/transformer/*": "./transformers/*/index.js",
+    "#atlaspack/validator/*": "./validators/*/index.js"
+  },
+  "bin": {
+    "atlaspack": "./cli/index.js"
+  },
+  "dependencies": {
```
alshdavid, line 50: `@atlaspack/core` and related non-plugin packages have fairly few dependencies however the plugins have many.

I tried to bundle the plugins with rspack so they bring their dependencies with them. That worked well but simply copying them is more direct.

PR #358, opened 2025-02-26
Add a step for scanning for side effects caused by loadable package

benjervis:
Confluence is having an issue adopting the inline requires optimizer because the way Loadable is set up puts a lot of side effects throughout the codebase, and inlining the requires messes that up.

This can be made to work by setting a large number of `package.json`s to have `"sideEffects": true`, but that can easily be ruined by someone incorrectly configuring another package.

A more robust approach is to find any imports of `@confluence/loadable`, and follow the asset graph up to mark all ancestor as having side effects. This ensures that transitive deps will be marked also.

Functionality is locked behind a feature flag.

### DIFF
--- a/.changeset/many-pears-tan.md
+++ b/.changeset/many-pears-tan.md
@@ -1,0 +1,6 @@
+---
+'@atlaspack/core': minor
+---
+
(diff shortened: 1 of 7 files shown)


mattcompiles (approved):
Nice

JakeLane (approved):

PR #364, opened 2025-03-03
V3: Plumbing for respondToFsEvents()

alshdavid:
Rust Core
- Adds `Atlaspack.respond_to_fs_events()`
- Adds `WatchEvent`

Rust Napi
- Adds `AtlaspackV3.respond_to_fs_events()`
- Lazily initializing `Atlaspack`
- Initializing workers in parallel on `AtlaspackV3.create()`

Javascript:
- Adds `AtlaspackV3.respondToFsEvents()`
- Initializing workers in parallel on `AtlaspackV3.create()`
- Increased maxListerners to `100`


crates/node-bindings/src/atlaspack/atlaspack.rs
```diff
@@ -103,44 +93,48 @@ impl AtlaspackNapi {
...
 
-    Ok(Self {
+    // Initialize Nodejs worker threads
+    for _ in 0..node_worker_count {
+      let transferable = JsTransferable::new(tx_worker.clone());
+
+      napi_options
+        .register_worker
+        .call1::<JsTransferable<Sender<NodejsWorker>>, JsUnknown>(transferable)?;
+    }
```
alshdavid, line 105: Initially, workers were set up and torn down at the start/end of an `AtlaspackV3.build()`. Given the state lives for the life of the `Atlaspack` instance, workers should too. 

Previously we would call build immediately after `AtlaspackV3.create()` so there is no change in behaviour. Also this doesn't change the performance characteristics because the workers initialize in parallel (as they were before).

crates/node-bindings/src/atlaspack/atlaspack_cell.rs
```diff
@@ -0,0 +1,38 @@
...
+use atlaspack::Atlaspack;
+use atlaspack::AtlaspackInitOptions;
+use once_cell::sync::OnceCell;
+
+/// Initialize Atlaspack in another thread and store the result.
+/// the result is available on the first access of Atlaspack
+#[derive(Clone)]
+pub struct AtlaspackCell(Arc<OnceCell<anyhow::Result<Arc<Atlaspack>>>>);
+
+impl AtlaspackCell {
```
mattcompiles, line 13: What do we gain from this? Feels like we're adding complexity but I'm not clear why. Also I don't think it's really related to this PR?

alshdavid: To preserve the state within `Atlaspack`, we need to new up the `Atlaspack` instance once and reuse that every time we call methods on it.

Currently we new up and throw away `Atlaspack` when we call `AtlaspackNapi.build_asset_graph` losing the state.

`Atlaspack` needs to be newed up on a seperate thread because Nodejs, so it cannot be initialized in the constructor. This is similar to:

```javascript
class AtlaspackNapi {
  #inner
  
  constructor() {
    this.#inner = new Promise(res => setTimeout(() => res(new Atlaspack), 0))
  }
  
  async buildAssetGraph() {
    let atlaspack = await this.#inner
    return await atlaspack.buildAssetGraph()
  }
}
```

crates/atlaspack/src/request_tracker/test.rs
```diff
@@ -49,41 +49,47 @@ async fn should_reuse_previously_run_request() {
...
 
-  assert_eq!(result, "A");
-  assert_eq!(request_a.run_count(), 1);
+  // assert_eq!(result, "A");
+  // assert_eq!(request_a.run_count(), 1);
 
-  let result = run_sub_request(&mut rt, &request_a).await;
-  assert_eq!(result, "A");
-  assert_eq!(request_a.run_count(), 1);
+  // let result = run_sub_request(&mut rt, &request_a).await;
```
MonicaOlejniczak, line 66: Should this pr go back in draft? Not sure why these are commented out

alshdavid: These tests relate to [this change](https://github.com/atlassian-labs/atlaspack/pull/364/files/ec806f7de55865947aec62e93d12e128c9ddc7cf). The change makes all requestTracker requests run which fails these tests.

Matt is working on a change that re-enables RT caching, these tests will be re-enabled then

PR #366, opened 2025-03-06
V3: NapWorkerPool to manage worker pool for AtlaspackV3

alshdavid:
Initialize workers once and reuse them between AtlaspackV3 instances

```typescript
// Initialize workers once
const napiWorkerPool = NapiWorkerPool()

const ap1 = new AtlaspackV3({
  napiWorkerPool
})

const ap2 = new AtlaspackV3({
  napiWorkerPool
})

// If not supplied it will default to spawning a new worker pool internally
const ap3 = new AtlaspackV3() // internally `new NapiWorkerPool()`
```

## Motivation
- Enable sharing worker instances between calls to `AtlaspackV3.build_asset_graph()` to facilitate dev mode 
- Workers become managed by their own self contained unit, simplifying the WorkerPool implementation
- Helps simplifying the logic used to share the WorkerPool between Atlaspack instances in the integration tests


crates/atlaspack_napi_helpers/src/transferable.rs
alshdavid, line 1: Added `get(): Arc<T>` that returns a reference to the original value

crates/atlaspack_plugin_rpc/src/nodejs/rpc/nodejs_rpc_factory.rs
alshdavid, line 1: This used to be responsible for initializing the workers however now that workers are initialized externally, this is simply passing the `Vec<NodejsWorker>` directly through `start()`.

mattcompiles (approved):

PR #367, opened 2025-03-07
V3: Get basic incremental builds working in V3

mattcompiles:
## Motivation

This PR adds the remaining features to get the first watcher test running against V3. 

## Changes

- Add `Invalid` and `FileInvalidation` request graph node types
- Assign request invalidations to the graph during `store_request`
- Add new `respond_to_fs_events` API to the native request tracker which invalidates relevant request results from prior builds
- Add  `asset.file_path` as an invalidation to all asset requests
- Add `asset.output_hash` and set it after finishing the transformation pipeline
  - This fixes false positive cache responses in the packaging phase
- Force parcel build request and bundle graph requests when native is set 

## Checklist

### DIFF
--- a/crates/atlaspack/src/request_tracker/request_tracker.rs
+++ b/crates/atlaspack/src/request_tracker/request_tracker.rs
@@ -17,6 +17,8 @@ use petgraph::visit::Reversed;
 use crate::plugins::PluginsRef;
 use crate::requests::RequestResult;
 use crate::AtlaspackError;
+use crate::WatchEvent;
+use crate::WatchEvents;
 
 use super::Request;
 use super::RequestEdgeType;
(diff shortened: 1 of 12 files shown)


crates/atlaspack/src/request_tracker/request_tracker.rs
```diff
@@ -281,6 +310,45 @@ impl RequestTracker {
...
+    for invalidation in watch_events.iter() {
+      match invalidation {
+        WatchEvent::Delete(file_path) | WatchEvent::Update(file_path) => {
+          if let Some(invalidation_node) = self.invalidations.get(file_path) {
+            let mut invalid_nodes = Vec::new();
+            {
+              let reverse_graph = Reversed(&self.graph);
+              let mut dfs = Dfs::new(reverse_graph, *invalidation_node);
+
+              while let Some(node_index) = dfs.next(reverse_graph) {
```
yamadapc, line 328: you could terminate early if the node has been visited already

mattcompiles: > you could terminate early if the node has been visited already by a previous event
I could terminate early but it likely wouldn't do much as the graph will only even be like 3/4 edges deep. 

> I'm not super sure if you need to traverse at all though, it might be enough just to taint this node?
You definitely need to traverse to ensure the parents are marked as invalid too.

MonicaOlejniczak (approved):

PR #368, opened 2025-03-10
Fix finding project root based on .git directory

yamadapc:
Test Plan: cargo test

### DIFF
--- a/crates/node-bindings/src/fs_search.rs
+++ b/crates/node-bindings/src/fs_search.rs
@@ -17,7 +17,7 @@ pub fn find_ancestor_file(filenames: Vec<String>, from: String, root: String) ->
 
     for name in &filenames {
       let fullpath = dir.join(name);
-      if fullpath.is_file() {
+      if fullpath.exists() {
         return Some(fullpath.to_string_lossy().into_owned());
       }
     }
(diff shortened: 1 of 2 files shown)


alshdavid (approved):

MonicaOlejniczak (approved):

pancaspe87 (approved):

crates/node-bindings/src/fs_search.rs
```diff
@@ -17,7 +17,7 @@ pub fn find_ancestor_file(filenames: Vec<String>, from: String, root: String) ->
 
     for name in &filenames {
       let fullpath = dir.join(name);
-      if fullpath.is_file() {
+      if fullpath.exists() {
```
dddlr, line 20: what is the difference between `is_file` and `exists`, and why does it matter here?

yamadapc: `is_file` will fail if the path is a directory or symlink

we should be testing `.git`/`.hg` and so on

dddlr: ah makes sense, i got confused because the function name is `find_ancestor_file` which i suppose should be something like `find_ancestor_path` instead?

PR #369, opened 2025-03-10
Use new server syntax that works in jest env

JakeLane:
## Motivation

For supporting importCond with jest APIs, it's much more trivial to lean on the server style proxying of the import object. This means we can set the feature gate at runtime, allowing for feature gate test suites to work seamlessly in the monorepo. It may make more sense to just support loading fresh modules (more like the browser environment), but it's a big rabbit hole to go down for a small benefit.

## Changes

This change moves to update each identifier in the module, referencing importCond assignment, to the lazy getter.

## Checklist

### DIFF
--- a/packages/utils/babel-plugin-transform-contextual-imports/src/index.ts
+++ b/packages/utils/babel-plugin-transform-contextual-imports/src/index.ts
@@ -1,6 +1,5 @@
-import type {PluginObj} from '@babel/core';
+import type {PluginObj, NodePath, types as BabelTypes} from '@babel/core';
 import {declare} from '@babel/helper-plugin-utils';
-import type {StringLiteral} from '@babel/types';
 
 interface Opts {
   // Use node safe import cond syntax
@@ -9,7 +8,8 @@ interface Opts {
 
 interface State {
   opts: Opts;
-  importMap?: Map<string, string>;
+  conditionalImportIdentifiers?: Set<string>;
+  visitedIdentifiers?: Set<BabelTypes.Identifier>;
 }
 
 const isNode = (opts: Opts): boolean => !!('node' in opts && opts.node);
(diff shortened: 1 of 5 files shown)


packages/utils/babel-plugin-transform-contextual-imports/src/index.ts
MonicaOlejniczak, line 1: I'm finding this file difficult to read, comments would be helpful to explain the logic

packages/utils/babel-plugin-transform-contextual-imports/src/index.ts
```diff
@@ -166,14 +147,68 @@ export default declare((api): PluginObj<State> => {
...
+                  call.callee.type === 'Identifier' &&
+                  call.callee.name === 'importCond'
+                ) {
+                  if (
+                    call.arguments.length === 3 &&
+                    call.arguments.every((arg) => arg.type === 'StringLiteral')
+                  ) {
+                    const [cond, ifTrue, ifFalse] = call.arguments;
+
+                    // Make module pass lazy in ssr
```
MonicaOlejniczak, line 172: Elaborate on this.

```suggestion
                    // Make conditional import evaluate lazily in a node environment so that conditions can be injected at runtime
```

packages/utils/babel-plugin-transform-contextual-imports/src/index.ts
```diff
@@ -166,14 +147,68 @@ export default declare((api): PluginObj<State> => {
...
+                    const identUid = path.scope.generateUid(
```
MonicaOlejniczak, line 173: Why can't you re-use the original identifier exactly?

packages/utils/babel-plugin-transform-contextual-imports/src/index.ts
```diff
@@ -1,19 +1,18 @@
...
 interface Opts {
-  server?: boolean;
+  // Use node safe import cond syntax
+  node?: boolean;
 }
 
 interface State {
   opts: Opts;
-  importNodes?: any[]; // Statement types didn't work so using any
+  importMap?: Map<string, string>;
```
MonicaOlejniczak, line 12: `importMap` is confusing to me. I automatically presume an import refers to an import declaration when operating on an AST, but really this is a map from the conditional import identifier → a renamed identifier. Either rename it (though I can't think of something clean and succinct) or document it.

packages/utils/babel-plugin-transform-contextual-imports/src/index.ts
JakeLane: This was initially required because I was creating a new variable, this is no longer needed so I'll remove

MonicaOlejniczak (approved):

PR #370, opened 2025-03-10
Add tests to VCS fs and untracked handling

yamadapc:
Fixes handling of untracked state with the VCS watcher implementation. Adds
integration tests for the VCS watcher listing.

Add instrumentation and way to turn on VCS by default without having to provide
it programmatically.


packages/core/fs/src/NodeVCSAwareFS.js
```diff
@@ -13,33 +13,74 @@ import {getFeatureFlagValue} from '@atlaspack/feature-flags';
 import packageJSON from '../package.json';
 
 export interface NodeVCSAwareFSOptions {
-  gitRepoPath: FilePath;
+  gitRepoPath: null | FilePath;
```
pancaspe87, line 16: If this is null -> we default to the old watcher events.
In what instance we expect this to not be provided?

pancaspe87 (approved):

packages/core/core/src/resolveOptions.js
```diff
@@ -45,11 +45,39 @@ function compileGlobs(globs: string[]): RegExp[] {
   return globs.map((glob) => globToRegex(glob));
 }
 
+function getDefaultFS(): FileSystem {
+  if (true) {
```
yamadapc, line 49: ooooops

packages/core/fs/src/NodeVCSAwareFS.js
yamadapc: This would be null if the project isn't a git repository.

crates/atlaspack_vcs/examples/vcs_file_diff.rs
```diff
@@ -56,7 +56,13 @@ fn main() {
     } => {
       let changes = get_changed_files(
         &repository_root,
-        &start_rev,
+        &VCSState {
```
MonicaOlejniczak, line 59: nit: Having a convenience constructor would be nice, like `new` or `from_git_hash` etc

yamadapc: I can do that yeah

packages/core/fs/src/NodeVCSAwareFS.js
```diff
@@ -13,33 +13,74 @@ import {getFeatureFlagValue} from '@atlaspack/feature-flags';
...
   constructor(options: NodeVCSAwareFSOptions) {
     super();
-    this.#options = options;
+    this.#excludePatterns = options.excludePatterns;
+    this.#logEventDiff = options.logEventDiff;
+    this.#gitRepoPath = options.gitRepoPath;
+  }
+
+  // $FlowFixMe[unclear-type] this is the serialization API
+  static deserialize(data: any): NodeVCSAwareFS {
```
MonicaOlejniczak, line 34: You should extract it to a type for ser / de

MonicaOlejniczak (approved):

PR #376, opened 2025-03-14
Optimize resolver performance by reducing calls to Specifier::to_string

yamadapc:
At the moment, for every import we will loop over the list of merged
`tsconfig.json` paths and convert the `Specifier`s on those paths into strings.

For all of the "package" specifiers, this involves a `format!` call, which is
taking quite a considerable amount of time to execute.

We can improve this by:

* caching the `Specifier::to_string` results for a tsconfig object, through the
  duration of the build

Test Plan: cargo test

### DIFF
--- a/packages/utils/node-resolver-rs/src/tsconfig.rs
+++ b/packages/utils/node-resolver-rs/src/tsconfig.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::cell::OnceCell;
 use std::collections::HashMap;
 use std::path::Path;
 use std::path::PathBuf;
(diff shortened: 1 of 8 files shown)


packages/utils/node-resolver-rs/src/tsconfig.rs
```diff
@@ -140,8 +158,7 @@ impl TsConfig {
       let mut best_key = None;
       let full_specifier = specifier.to_string();
 
-      for key in paths.keys() {
-        let path = key.to_string();
+      for (key, path) in &self.paths_specifier_strings {
```
yamadapc, line 161: we can make this less flaky with `OnceCell`

JakeLane (approved):
Missing a changeset

MonicaOlejniczak (approved):

PR #379, opened 2025-03-19
Support nested importCond usage in async bundles

JakeLane:
### Motivation

Async bundles were sometimes not loading all the correct bundles. Turns out this is because dependent `importCond` usages need to be loaded based on the condition. This was omitted by mistake in the implementation and should have been the initial behaviour.

### Changes

- Make `loadConditionalBundlesInParallel` only opt-in, rather than defaulting to enabled in dev mode
- Add new feature flag `conditionalBundlingNestedRuntime`
- When feature flag is enabled, load nested bundles in async runtime edges


.vscode/launch.json
```diff
@@ -15,6 +15,7 @@
       "args": [
         "--inspect-brk",
         "${workspaceFolder}/node_modules/.bin/_mocha",
+        "--timeout=1000000",
```
dddlr, line 18: are the tests timing out?

packages/bundlers/default/src/idealGraph.js
```diff
@@ -385,11 +385,8 @@ export function createIdealGraph(
                   bundleRoots.get(referencingBundleRoot),
                 )[0];
 
-                if (
-                  config.loadConditionalBundlesInParallel ??
-                  !bundle.env.shouldScopeHoist
```
dddlr, line 390: sorry, i forgot what the context behind removing `shouldScopeHoist` was 😅 could you please remind me

dddlr (approved):

packages/runtimes/js/src/JSRuntime.js
dddlr, line 1: i don't think i have enough context on this file to give it a thorough review, but the changes and the new test both look reasonable

packages/bundlers/default/src/idealGraph.js
JakeLane: yeah, it was becoming too confusing to talk about with the team. Given we've already configured this to false in Jira, this has no change.

liamqma (approved):

.changeset/great-turtles-hide.md
```diff
@@ -0,0 +1,9 @@
+---
+'@atlaspack/integration-tests': patch
+'@atlaspack/feature-flags': patch
+'@atlaspack/bundler-default': patch
+'@atlaspack/transformer-js': patch
```
mattcompiles, line 5: The transformer isn't changing here right?

mattcompiles (approved):
This seems fine and your tests look good. I'm not really clear on why `conditionalBundlingNestedRuntime` is needed though. Is the intention to replace the original behaviour at some point?

.changeset/great-turtles-hide.md
JakeLane: yeah, this was automated by changesets. I'll double check it's correct

JakeLane:
> This seems fine and your tests look good. I'm not really clear on why `conditionalBundlingNestedRuntime` is needed though. Is the intention to replace the original behaviour at some point?

It's required to rollout safely, as we're already using conditional bundling in production

PR #381, opened 2025-03-21
Fix re-used bundles for async imports with circular dependencies

mattcompiles:
## Motivation

This PR resolves an issue with re-used bundles when they have circular dependencies on each other. Previous to this fix, Assets that were required by two async bundles that had a circular dependency on each other would not be placed in either bundle, causing `Cannot find module` errors at runtime.  

## Changes

The change here ensures that Assets are not removed from reused bundles when there isn't anywhere else safe to put them. 

## Checklist

### DIFF
--- a/packages/bundlers/default/src/idealGraph.js
+++ b/packages/bundlers/default/src/idealGraph.js
@@ -961,7 +961,7 @@ export function createIdealGraph(
           );
 
           reachableIntersection.forEach((otherCandidateId) => {
-            // In the case of a ciruclar dependency, you may end up with two
+            // In the case of a circular dependency, you may end up with two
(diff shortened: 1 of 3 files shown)


packages/bundlers/default/src/idealGraph.js
```diff
@@ -961,6 +961,12 @@ export function createIdealGraph(
           );
 
           reachableIntersection.forEach((otherCandidateId) => {
+            // In the case of a ciruclar dependency, you may end up with two
```
JakeLane, line 964:
```suggestion
            // In the case of a circular dependency, you may end up with two
```

MonicaOlejniczak (approved):

PR #388, opened 2025-03-25
Make lmdb-js-lite the default cache back-end

yamadapc:
* Enable `useLmdbJsLite` by default.
* Clean-up cache directories after every test


packages/core/feature-flags/src/index.js
```diff
@@ -12,7 +12,7 @@ export const DEFAULT_FEATURE_FLAGS: FeatureFlags = {
   useWatchmanWatcher: false,
   importRetry: false,
   fixQuadraticCacheInvalidation: 'OLD',
-  useLmdbJsLite: false,
+  useLmdbJsLite: true,
```
MonicaOlejniczak, line 15: Why not cleanup the flag completely?

yamadapc: so we could turn it off still

packages/core/fs/src/NodeFS.js
```diff
@@ -134,7 +134,11 @@ export class NodeFS implements FileSystem {
 
     writeStream.once('error', () => {
       failed = true;
-      fs.unlinkSync(tmpFilePath);
+      try {
```
mattcompiles, line 137: Any context on this?

mattcompiles (approved):
Looks good mate 👍

packages/core/fs/src/NodeFS.js
yamadapc: I had code added which removed the temporary files before this ran.

If this fails (due to the `tmpFilePath` not existing), then there's an unhandled exception on this block. We need to handle the error here.

Since this is clean-up it should be fine to ignore it.

PR #390, opened 2025-03-26
Modified AtlaspackLazy to avoid concurrent access

alshdavid:
The previous implementation of `AtlaspackLazy` would allow methods to be accessed concurrently which would cause issues.

Also added `.take()` method to facilitate manual shutdown of Atlaspack.

```javascript
const atlaspack = new Atlaspack({ /* ... */})

await Promise.all([
  atlaspack.buildAssetGraph(),
  atlaspack.buildAssetGraph(),
  atlaspack.buildAssetGraph(),
])
```

### DIFF
--- a/crates/node-bindings/src/atlaspack/atlaspack_lazy.rs
+++ b/crates/node-bindings/src/atlaspack/atlaspack_lazy.rs
@@ -2,31 +2,56 @@ use std::sync::Arc;
 
 use atlaspack::Atlaspack;
 use atlaspack::AtlaspackInitOptions;
-use once_cell::sync::Lazy;
+use parking_lot::Mutex;
 
-type LazyAtlaspackResult = Result<Arc<Atlaspack>, anyhow::Error>;
-type LazyAtlaspackInner =
-  Arc<Lazy<LazyAtlaspackResult, Box<dyn FnOnce() -> LazyAtlaspackResult + Send + 'static>>>;
+// type LazyAtlaspackResult = Result<Arc<Atlaspack>, anyhow::Error>;
+type LazyAtlaspackInner = Arc<Mutex<Option<anyhow::Result<Arc<Atlaspack>>>>>;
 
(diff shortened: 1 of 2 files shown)


crates/node-bindings/src/atlaspack/atlaspack_lazy.rs
```diff
@@ -2,31 +2,56 @@ use std::sync::Arc;
 
 use atlaspack::Atlaspack;
 use atlaspack::AtlaspackInitOptions;
-use once_cell::sync::Lazy;
+use parking_lot::Mutex;
 
-type LazyAtlaspackResult = Result<Arc<Atlaspack>, anyhow::Error>;
-type LazyAtlaspackInner =
-  Arc<Lazy<LazyAtlaspackResult, Box<dyn FnOnce() -> LazyAtlaspackResult + Send + 'static>>>;
+// type LazyAtlaspackResult = Result<Arc<Atlaspack>, anyhow::Error>;
```
MonicaOlejniczak, line 7: Remove this, and add a changeset

JakeLane:
> The previous implementation of AtlaspackLazy would allow methods to be accessed concurrently which would cause issues.

Which issues? Is this resolving a known error or something else?

crates/node-bindings/src/atlaspack/atlaspack_lazy.rs
```diff
@@ -2,31 +2,56 @@ use std::sync::Arc;
...
+pub struct AtlaspackLazy {
+  atlaspack_container: LazyAtlaspackInner,
+  options: AtlaspackInitOptions,
+}
 
 impl AtlaspackLazy {
   pub fn new(options: AtlaspackInitOptions) -> Self {
-    Self(Arc::new(Lazy::new(Box::new(
-      move || match Atlaspack::new(options) {
+    Self {
```
JakeLane, line 20: I think this needs to be feature flagged to roll out safely

PR #395, opened 2025-03-27
CI: Cache Rust "target" in CI

alshdavid:
Added caching for `target` directory in CI to speed up builds.

`main` will create the cache, branches will fork that for their own build

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
     name: Build native
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -67,9 +92,14 @@ jobs:
           cache: yarn
           node-version: ${{ matrix.node }}
       - uses: ./.github/actions/rust-toolchain
-      - uses: Swatinem/rust-cache@v2
+      - name: Cache Rust
```
MonicaOlejniczak, line 95: Remove these since they're in `build_native`

alshdavid:
Can't seem to stop Rust from rebuilding the workspace packages 🤷

PR #396, opened 2025-03-27
Bring back server option to transform-contextual-imports

JakeLane:
## Motivation

The recent breaking change shipped to the babel plugin was not within our new process, so if we bring back the original config option - we can bump without config changes

## Changes

Bring back the original code but behind a function, so it can be easily deleted in a future PR.

## Checklist

### DIFF
--- a/.changeset/good-toes-worry.md
+++ b/.changeset/good-toes-worry.md
@@ -2,4 +2,4 @@
 '@atlaspack/babel-plugin-transform-contextual-imports': patch
 ---
 
-Bring back server config option to @atlaspack/babel-plugin-transform-contextual-imports
+Bring back server config option to babel plugin
--- a/.changeset/good-toes-worry.md
+++ b/.changeset/good-toes-worry.md
@@ -1,0 +1,5 @@
+---
+'@atlaspack/babel-plugin-transform-contextual-imports': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/good-toes-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atlaspack/babel-plugin-transform-contextual-imports': patch
+---
+
+Bring back server config option to @atlaspack/babel-plugin-transform-contextual-imports
```
MonicaOlejniczak, line 5: nit: You shouldn't need to specify the package name as the changeset will group it for you

packages/utils/babel-plugin-transform-contextual-imports/src/index.ts
```diff
@@ -2,18 +2,25 @@ import type {PluginObj, NodePath, types as BabelTypes} from '@babel/core';
 import {declare} from '@babel/helper-plugin-utils';
 
 interface Opts {
+  // Deprecated syntax, use "node" instead
```
MonicaOlejniczak, line 5: Use jsdoc for this, as it should come up in IDEs https://jsdoc.app/tags-deprecated

MonicaOlejniczak (approved):

PR #399, opened 2025-03-28
Fix out-of-sync default config with dependencies and make test match

JakeLane:
## Motivation

https://github.com/atlassian-labs/atlaspack/pull/383/files removed parcelDependencies meaning `index.json` now is missing the correct dependencies.

## Changes

Simple fix is to specify all the deps. We may want to reconsider this in the future if we want to reduce the amount of deps required for basic builds.

## Checklist


JakeLane:
@benjervis fixed this in a different PR

PR #402, opened 2025-03-28
Native: Initialize AtlaspackNative async

alshdavid:
## Motivation

Lazy initialization of AtlaspackNapi was hiding errors occuring asynchronously. This fixes the watcher hanging, but also Matt's PR bumping the watcher version fixes it too. Por que no los dos? 

## Changes

- Updated AtlaspackNapi to have an async constructor
- Updated AtlaspackNapi to use functions rather than napi methods
- Re-enabled napi macros (cannot use napi classes until we migrate to napi v3)

## Checklist


crates/atlaspack_plugin_rpc/src/nodejs/rpc/nodejs_rpc_factory.rs
alshdavid, line 1: This can be initialized synchronously now, no need for any channel shenanigans

packages/core/core/src/Atlaspack.js
```diff
@@ -161,7 +161,7 @@ export default class Atlaspack {
       const version = require('../package.json').version;
       await lmdb.put('current_session_version', Buffer.from(version));
 
-      rustAtlaspack = new AtlaspackV3({
+      rustAtlaspack = await AtlaspackV3.new({
```
mattcompiles, line 164: Haha call it `init` or `create` or something. `new` feels magic haha

crates/node-bindings/src/atlaspack/atlaspack_lazy.rs
alshdavid, line 1: 🪦

packages/core/core/src/Atlaspack.js
alshdavid: Soo picky 🤣  - Updated PR

packages/core/core/src/atlaspack-v3/AtlaspackV3.js
```diff
@@ -66,6 +82,6 @@ export class AtlaspackV3 {
   }
 
   respondToFsEvents(events: Array<Event>): boolean {
-    return this._internal.respondToFsEvents(events);
+    return atlaspackNapiRespondToFsEvents(this._atlaspack_napi, events);
```
MonicaOlejniczak, line 85: Why did you opt for functions instead of adding to the `_atlaspack_napi` instance like before?

alshdavid: napi-rs V2 has broken macros for napi classes requring us to disable macro expansion in our LSPs. This results in incorrect intelisense and limits the napi features we can use. For instance; the boilerplate in [this section of code](https://github.com/atlassian-labs/atlaspack/blob/main/crates/node-bindings/src/atlaspack/serialize_asset_graph.rs#L42) could all be handled by the macro but without macro expansion we cannot use it.

I'd like to turn macro expansion back on and to do that we either need to:

- Migrate to [napi-rs v3](https://crates.io/crates/napi/3.0.0-alpha.33) (which is still in alpha)
- Avoid the use of napi-rs classes (recreating them in the JS layer)

Personally, I'm okay with migrating to v3, even if it's in alpha, but I am not sure how much work is needed for the migration and if the team is comfortable with pre-release software.

mattcompiles (approved):

PR #405, opened 2025-04-02
Improve monolith bundler support

benjervis:
## Motivation

When the single file output option was first defined, it was a very quick and naive implementation that only added JS assets to a single bundle, primarily to support SSR runtimes that only allow a single file.

This falls apart when attempting to server render something like an SVG, because the existing implementation would ignore them entirely.

## Changes

This is fixed by adding support for "isolated" bundles, of which SVGs are one but not the only use case. The content itself is not included in the primary bundle, but the references between them (like an SVG URL) will now be inserted.

If the bundling algorithm finds a dependency whose asset is isolated, it now creates a separate bundle, links it to the main bundle, and preserves its downstream dependencies.

## Checklist


mattcompiles (approved):
Nice

packages/bundlers/default/src/MonolithicBundler.js
```diff
@@ -16,17 +17,69 @@ export function addJSMonolithBundle(
...
+      let assets = bundleGraph.getDependencyAssets(dependency);
+      invariant(
+        assets.length === 1,
+        'Expected dependency to have exactly one asset',
+      );
+
+      let asset = assets[0];
+
+      // For assets marked as isolated, we create new bundles and let other
+      // plugins like optimizers include them in the primary bundle
```
mgfalzon, line 49: Why can't we add these types of assets directly to the bundle like we do for the js assets and for the async dependencies?

PR #406, opened 2025-04-02
Upgrade punycode to 2.3.1

yamadapc:
Should fix errors related to incompatible version ranges on projects that install the newer version.

Since we are bundling the code, the changes in v2 shouldn't be an issue.


alshdavid (approved):

alshdavid:
Nice! 👍

PR #410, opened 2025-04-03
Support sparse checkouts

yamadapc:
Query git status from the CLI rather than libgit2 so we can support sparse checkouts.

### DIFF
--- a/crates/atlaspack_vcs/src/lib.rs
+++ b/crates/atlaspack_vcs/src/lib.rs
@@ -447,16 +447,13 @@ pub fn get_changed_files(
   new_rev: Option<&str>,
   failure_mode: FailureMode,
 ) -> anyhow::Result<Vec<FileChangeEvent>> {
-  tracing::info!("Repository::open");
   let repo = Repository::open(repo_path)?;
   let old_rev = &vcs_state.git_hash;
-  tracing::info!("Repository::open done");
   let old_commit = repo.revparse_single(old_rev)?.peel_to_commit()?;
   let new_commit = repo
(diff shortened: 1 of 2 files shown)


pancaspe87 (approved):

crates/atlaspack_vcs/src/lib.rs
```diff
@@ -423,29 +403,64 @@ pub fn get_changed_files_from_git(
...
+  command.arg("status").arg("--porcelain").arg("--no-ignored");
+  command.current_dir(repo_path);
+  let output = command.output()?;
+  if !output.status.success() {
+    return Err(anyhow::anyhow!("Git status failed"));
+  }
+  let output = String::from_utf8(output.stdout)?;
+  let lines = output.split_terminator('\n');
+  for line in lines {
+    let status = line.chars().nth(0).unwrap();
```
MonicaOlejniczak, line 423: Add some comments on the nth / skipping logic. Also does this work if you get this spam I usually get in AFM?

```
warning: could not open directory '..': No such file or directory
```

yamadapc: I think anything like that would be stderr ;

There are integration tests, but not for the sparse checkout specifically

MonicaOlejniczak (approved):

crates/atlaspack_vcs/src/lib.rs
MonicaOlejniczak: The warnings don't appear for me when running the command in rust for some reason, and appears to work as expected

PR #412, opened 2025-04-04
Fix unit-tests on node v22

yamadapc:
Also update workflow to run them on v22.

This was failing because globalThis.navigator does not seem to be being mocked
correctly on node v22.

Test Plan: yarn test:unit

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
     name: Unit tests (${{ matrix.os }}, Node ${{ matrix.node }})
     strategy:
       matrix:
-        node: [18, 20]
+        node: [18, 20, 22]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     name: Unit tests (${{ matrix.os }}, Node ${{ matrix.node }})
     strategy:
       matrix:
-        node: [18, 20]
+        node: [18, 20, 22]
```
mattcompiles, line 60: We often struggle with mac agents limits already. How would you feel about not bothering with the unit tests on Mac?

yamadapc: Sure ; done

mattcompiles (approved):

PR #418, opened 2025-04-07
Run VCS queries on a background thread

yamadapc:
Prevent blocking the main-thread with git commands and yarn diffing.


JakeLane (approved):

crates/node-bindings/src/vcs.rs
```diff
@@ -1,34 +1,49 @@
...
-  let path = Path::new(&path);
-  let vcs_state = VCSState::read_from_repository(
-    path,
-    &exclude_patterns,
-    FailureMode::IgnoreMissingNodeModules,
-  )
-  .map_err(|err| napi::Error::new(napi::Status::GenericFailure, format!("[napi] {}", err)))?;
-  let vcs_state = env.to_js_value(&vcs_state)?;
+) -> napi::Result<JsObject> {
+  env.execute_tokio_future(
```
alshdavid, line 14: Probably best not to include the napi tokio adapter as that spawns another worker pool alongside the existing worker pools. Perhaps use `thread::spawn` here?

yamadapc: Hm... Do you mean we aren't currently using the napi tokio runtime // there's a different runtime we are instantiating?

alshdavid: We had this chat in DMs but for everyone else's context. We have a different runtime that's seperately initialized so initializing tokio again with a general runtime would create a new runtime that lives alongside our existing one.

crates/node-bindings/Cargo.toml
```diff
@@ -33,6 +33,7 @@ num_cpus = { workspace = true }
 parking_lot = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
+tokio = { workspace = true }
```
alshdavid, line 36: Can remove this now 👍

alshdavid (approved):

PR #419, opened 2025-04-08
fix vcsMode feature flagging

pancaspe87:
ensure `vcsMode` are only checked if VCS is enabled


yamadapc (changes requested):
We don't need a separate flag for this.

We can ensure that the vcsMode flag is set correctly instead.

PR #420, opened 2025-04-09
Support async runtime in conditional bundling

JakeLane:
## Motivation

One thing that was missed when implementing conditional bundling, was the unstable async runtime in use in Jira. If we get a race condition, async bundles may load too late.

## Changes

This change adds support for conditional bundles into the async runtime, by making sure the runtime is added for additional bundles.

## Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,6 +14,7 @@
       "name": "Run Integration Tests (lldb)",
       "args": [
         "--inspect-brk",
+        "--experimental-vm-modules",
         "${workspaceFolder}/node_modules/.bin/_mocha",
         "--timeout=1000000",
         "${input:integration-tests-args}"
--- a/packages/core/core/src/public/BundleGraph.js
+++ b/packages/core/core/src/public/BundleGraph.js
@@ -444,10 +444,11 @@ export default class BundleGraph<TBundle: IBundle>
   // be used by a webserver to understand which conditions are used by which bundles,
   // and which bundles those conditions require depending on what they evaluate to.
   getConditionalBundleMapping(): Map<
-    TBundle,
+    string,
     Map<
       string,
       {|
+        bundle: TBundle,
         ifTrueBundles: Array<TBundle>,
         ifFalseBundles: Array<TBundle>,
       |},
@@ -486,14 +487,15 @@ export default class BundleGraph<TBundle: IBundle>
       }
 
       for (let bundle of bundles) {
-        const conditions = bundleConditions.get(bundle) ?? new Map();
+        const conditions = bundleConditions.get(bundle.id) ?? new Map();
 
         conditions.set(cond.key, {
+          bundle,
           ifTrueBundles,
           ifFalseBundles,
         });
 
-        bundleConditions.set(bundle, conditions);
+        bundleConditions.set(bundle.id, conditions);
       }
     }
 
--- a/packages/core/integration-tests/test/conditional-bundling.js
+++ b/packages/core/integration-tests/test/conditional-bundling.js
@@ -921,7 +921,10 @@ describe('conditional bundling', function () {
           shouldOptimize: false,
           outputFormat: 'esmodule',
         },
-        featureFlags: {conditionalBundlingApi: true},
+        featureFlags: {
+          conditionalBundlingApi: true,
+          conditionalBundlingAsyncRuntime: true,
+        },
         inputFS: overlayFS,
       },
     );
--- a/packages/core/integration-tests/test/conditional-bundling.js
+++ b/packages/core/integration-tests/test/conditional-bundling.js
@@ -15,6 +15,7 @@ import {
   distDir,
 } from '@atlaspack/test-utils';
 import sinon from 'sinon';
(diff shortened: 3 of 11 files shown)


packages/core/core/src/public/BundleGraph.js
```diff
@@ -486,14 +487,15 @@ export default class BundleGraph<TBundle: IBundle>
       }
 
       for (let bundle of bundles) {
-        const conditions = bundleConditions.get(bundle) ?? new Map();
+        const conditions = bundleConditions.get(bundle.id) ?? new Map();
```
mattcompiles, line 490: How come you changed to string keys? Not a biggy, but string keys are much slower than reference or number keys if you're doing a lot of lookups.

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -296,15 +309,46 @@ export class ScopeHoistingPackager {
...
 
     let params = [
       JSON.stringify(this.bundle.publicId),
       fnExpr(this.bundle.env, [], [codeToRun]),
-      JSON.stringify(deps),
+      `${JSON.stringify(deps)}${
+        conditions.length > 0
+          ? `.concat([${conditions
+              .map((conditions) => `...${conditions}`)
+              .join(',')}])`
```
mattcompiles, line 349: Why do this work at runtime? Can't we just add the conditions to the `deps`?

JakeLane: We need to because each dep is only required if the condition passes/fails

packages/core/core/src/public/BundleGraph.js
JakeLane: Yeah the bundle references weren't matching for some reason, so I had to refactor this to use ID

.vscode/launch.json
```diff
@@ -14,6 +14,7 @@
       "name": "Run Integration Tests (lldb)",
       "args": [
         "--inspect-brk",
+        "--experimental-vm-modules",
```
dddlr, line 17: what is the purpose of `--experimental-vm-modules`? :eyes:

JakeLane: This is because some ESM modules are imported during the test, which requires this flag for Node to understand. Before this, it was transpiled from ESM to CJS.

packages/core/integration-tests/test/conditional-bundling.js
```diff
@@ -870,4 +871,82 @@ describe('conditional bundling', function () {
...
+    await overlayFS.mkdirp(dir);
+
+    await fsFixture(overlayFS, dir)`
+      cond-a.js:
+        export default 'cond-a';
+
+      cond-b.js:
+        export default 'cond-b';
+
+      cond-a.html:
```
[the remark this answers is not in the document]
JakeLane: unfortunately not, because there's no HTML loading in integration tests as it runs in a node env. This test tries to emulate that with loading order.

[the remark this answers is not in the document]
JakeLane: If we had E2E browser tests we could though

packages/packagers/js/src/ScopeHoistingPackager.js
dddlr: this is my first time seeing `$parcel$global.rwr` so forgive me for the basic question 😅 

does `$parcel$global.rwr` run at runtime? and on a related note, does this change add the entries from `ifTrueBundle` and `ifFalseBundle` as runtime dependencies for `$parcel$global.rwr` depending on the runtime value of `globalThis.__MCOND('${key}')` ?

dddlr (approved):

packages/packagers/js/src/ScopeHoistingPackager.js
JakeLane: `rwr` means "Run When Ready", which is required because scripts load async (`<script async src="..." />`). This `rwr` means that the code will execute when the required modules have registered themselves as loaded.

Given conditional bundling could conditionally load a script - we need to make sure we only wait for it if it's needed.

packages/core/integration-tests/test/conditional-bundling.js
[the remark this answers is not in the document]
mattcompiles: FYI We do actually support loading HTML in the integration tests via simulated code in [runBundle](https://github.com/atlassian-labs/atlaspack/blob/570493beaf754e7985aebc7daaaf6dfcfa8fe56b/packages/core/test-utils/src/utils.js#L451). There are also [E2E playwright tests](https://github.com/atlassian-labs/atlaspack/blob/main/packages/core/e2e-tests/test/e2e-tests.test.js) if that would be better.

mattcompiles (approved):

packages/core/integration-tests/test/conditional-bundling.js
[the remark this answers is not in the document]
JakeLane: Oof I wish I knew about the E2E tests, would have added that

PR #423, opened 2025-04-09
bumping @parcel/watcher

alshdavid:
## Motivation

Bumping version of `@parcel/watcher` dependency

## Checklist

- [x ] Existing or new tests cover this change
- [x ] There is a changeset for this change, or one is not required


alshdavid:
Breaking this up

PR #425, opened 2025-04-09
Move Rust lint config to workspace rather than per-crate

alshdavid:
[no-changeset]: No consumer facing changes

## Motivation

Centralize Rust lint configuration

## Changes

- Added global clippy config
- Moved clippy settings to the workspace root

## Checklist


Cargo.toml
```diff
@@ -11,6 +11,13 @@ members = [
 inherits = "release"
 debug = true
 
+[workspace.lints.rust]
+# TODO enforce
+# rust_2018_idioms = "warn"
```
alshdavid, line 16: Enabled in a follow up PR

.changeset/kind-birds-teach.md
```diff
@@ -0,0 +1,6 @@
+---
+'@atlaspack/integration-tests': patch
+'@atlaspack/transformer-js': patch
+---
+
+Added global clippy settings
```
mattcompiles, line 6: I wouldn't add a changeset for this as there's no user facing changes

alshdavid: Updated

PR #426, opened 2025-04-10
Alsh/bump rust 1 86 0

alshdavid:
## Changes

Bump rust version to `1.86.0`

## Checklist


alshdavid:
Breaking this up

PR #427, opened 2025-04-10
Enforce rust_2018_idioms lints

alshdavid:
## Changes

Enabled `rust_2018_idioms` lints and updated code to fix errors

## Checklist


JakeLane (approved):

crates/json-comments-rs/src/lib.rs
```diff
@@ -423,17 +423,9 @@ mod tests {
...
-            "a": 4,
-            # another
-        }"#;
-
-    let expected = "{
-                           \n            \"a\": 4,
-                     \n        }";
-
-    assert_eq!(strip_string(json), expected);
+    let json = include_str!("../fixtures/line_comments.txt").to_string();
```
mattcompiles, line 426: absolutely not

alshdavid: Added a comment explaining why these were externalized

PR #432, opened 2025-04-10
Chore: Rust update v1.86.0

alshdavid:
## Changes

- Update Rust to 1.86.0
- Correct new lint rules

## Checklist


alshdavid:
Will come back to this

PR #456, opened 2025-04-23
Chore: Native build script refactor

alshdavid:
## Motivation

Centralize native build script logic so that it's contained within the `build-native.js` file

## Changes

Split the native build into two phases:
- Build rust projects
- Generate napi types and glue code per project

A package in the repo hooks into the native build by specifying a `package.json#napi.name` key.
A package can skip napi bindings by specifying `package.json#napi.skip`

## Checklist

[no-changeset]: Update build script


scripts/build-native.js
```diff
@@ -2,70 +2,80 @@
...
-        shell: true,
+        cwd: pkgDir,
       });
-
-      yarn.on('close', (code) => (code === 0 ? resolve() : reject()));
-    }).catch(() => process.exit(1));
-  }
-}
-
-// This setup is necessary for cross-compilation for Apple Silicon in GitHub Actions.
```
alshdavid, line 54: We are no longer compiling macos-arm64 from macos-amd64 so this is no longer needed

packages/core/rust/package.json
```diff
@@ -30,12 +30,7 @@
     "napi-wasm": "^1.0.1"
   },
   "scripts": {
-    "build": "napi build --platform --cargo-cwd ../../../crates/node-bindings",
-    "build-canary": "napi build --platform --profile canary --features canary --cargo-cwd ../../../crates/node-bindings",
-    "build-release": "napi build --platform --release --cargo-cwd ../../../crates/node-bindings",
-    "test": "mocha",
-    "wasm:build": "cargo build -p atlaspack-node-bindings --target wasm32-unknown-unknown && cp ../../../target/wasm32-unknown-unknown/debug/atlaspack_node_bindings.wasm .",
-    "wasm:build-release": "CARGO_PROFILE_RELEASE_LTO=true cargo build -p atlaspack-node-bindings --target wasm32-unknown-unknown --release && wasm-opt --strip-debug -O ../../../target/wasm32-unknown-unknown/release/atlaspack_node_bindings.wasm -o atlaspack_node_bindings.wasm"
```
alshdavid, line 38: Remove per-package config for napi builds, instead rely on the build script to copy artifacts and generate types

crates/lmdb-js-lite/package.json
```diff
@@ -10,6 +10,7 @@
     "win32"
   ],
   "napi": {
+    "skip": true,
```
mattcompiles, line 13: Why are we skipping this?

alshdavid: It was excluded in the original script and has appears to have hand-written bindings to facilitate napi stuff. So skipping this to ensure it has the same behaviour as before

mattcompiles (approved):

PR #460, opened 2025-04-23
Updated pipeline to upload/download/merge ./packages/**/* prior to publish

alshdavid:
## Motivation

Need to update the pipeline so the folders in `/packages` are distributed "as is" to avoid native packages requiring manually copying things in/out per package

Looks good: 
https://github.com/atlassian-labs/atlaspack/actions/runs/14635502646
https://www.npmjs.com/package/@atlaspack/rust/v/3.0.1-dev.16?activeTab=code

image

## Changes

- Update pipelines to build once and reuse build artifacts
  - Multi-platform builds run in parallel, upload their `packages/**/*` "as is"
  - Downstream pipeline steps download & merge `packages/**/*`
- Using native linux-arm64 runner for arm64 build (no cross compilation)
- Removed 32 bit arm build (can re-add it later)
- Added script under `scripts` to handle extracting/stripping debug symbols
  - Will remove this once `@atlaspack/rust` is split into platform specific packages because the file size will be smaller, even with the debug symbols, than it is now with all 5 platform/arch binaries included

## Checklist

[no-changeset]: Updating pipeline


scripts/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@workspace/scripts",
+  "private": true,
+  "version": "0.0.1-local",
+  "devDependencies": {
+    "glob": "^7.1.6",
+    "zx": "^8.1.9"
```
yamadapc, line 7: left-over? if we have `zx` might as well use it
makes the script neater / safer

alshdavid: One of the scripts uses `zx` so I included it in the package.json to reflect. Strictly speaking I don't need to include a package.json but I feel it's warranted for correctness - haha.

I can remove it if that's contentious

mattcompiles (approved):

scripts/package.json
mattcompiles: I think the point is more that zx would be way cleaner than stuff like this. 
```
function cmd(command, options) {
  try {
    const [arg0, ...args] = command.split(' ').filter(v => v !== '').map(v => v.trim())
    child_process.execFileSync(arg0, args, { stdio: 'inherit', shell: true, ...options })
  } catch (error) {
    console.error(`Failed: ${command}`)
    console.error(error)
    process.exit(1)
  }
}
```

PR #463, opened 2025-04-23
Update build script to avoid double builds for napi projects

alshdavid:
## Motivation

Build script will do a double build for napi projects because `@napi-rs/ci` does some stuff internally that triggers it

## Changes

- Update native build script to run the `napi build` command from the project root 
- Update native build script so it only runs if the package declares `package.json#napi.targets`
  - Will use this to split napi builds into separate packages

## Checklist

[no-changeset]: updating build script

### DIFF
--- a/packages/core/rust/package.json
+++ b/packages/core/rust/package.json
@@ -12,7 +12,8 @@
   "main": "index.js",
   "browser": "browser.js",
   "napi": {
-    "name": "atlaspack-node-bindings"
+    "name": "atlaspack-node-bindings",
+    "permittedTargets": "*"
   },
   "engines": {
     "node": ">= 16.0.0"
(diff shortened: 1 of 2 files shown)


packages/core/rust/package.json
```diff
@@ -12,7 +12,17 @@
   "main": "index.js",
   "browser": "browser.js",
   "napi": {
-    "name": "atlaspack-node-bindings"
+    "name": "atlaspack-node-bindings",
+    "targets": [
+      "x86_64-unknown-linux-gnu",
+      "x86_64-unknown-linux-musl",
```
mattcompiles, line 18: Why do we list all these musl and windows targets when we currently don't support them?

alshdavid: Updated

mattcompiles (approved):

PR #464, opened 2025-04-24
DNM; Splitting native into separate @atlaspack/rust-{os}-{arch} packages

alshdavid:
## Motivation

Reduces the download size of the native package and allows us to distribute debug symbols within the package

## Changes

- moved `/package/core/rust` to `/packages/native/rust`
- Added `/packages/native/rust-linux-amd64`
- Added `/packages/native/rust-linux-arm64`
- Added `/packages/native/rust-macos-amd64`
- Added `/packages/native/rust-macos-arm64`
- Added `/packages/native/rust-windows-amd64`

## Checklist

- [] There is a changeset for this change, or one is not required


alshdavid:
Will come back to this

PR #472, opened 2025-04-28
Add worker thread hack

yamadapc:
This adds a work-around to `rust#91979` (https://github.com/rust-lang/rust/issues/91979)

A segmentation fault is currently triggered on worker thread exit if:

* Application runs with `atlaspack` on the main thread
* Plugin on worker-thread imports `parcel/rust`

The work-around is to force parcel native bindings to be loaded eagerly if
available.

Test Plan: test dev release

### DIFF
--- a/packages/core/core/src/rustWorkerThreadDylibHack.js
+++ b/packages/core/core/src/rustWorkerThreadDylibHack.js
@@ -1,0 +1,18 @@
+// @flow strict-local
+
+/**
+ * This is a workaround for https://github.com/rust-lang/rust/issues/91979
+ * when running atlaspack with parcel bindings, it is possible that the parcel
+ * dylib will be loaded from a node worker thread, which causes a crash on exit.
+ *
+ * This is a workaround to ensure that the parcel dylib is loaded in the main
+ * thread, which fixes the crash.
+ */
+export function loadRustWorkerThreadDylibHack() {
+  try {
+    // $FlowFixMe
+    require('@parcel/rust'); // eslint-disable-line
+  } catch (err) {
+    /* ignore */
+  }
+}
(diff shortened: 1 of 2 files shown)


mattcompiles (approved):

packages/core/core/src/rustWorkerThreadDylibHack.js
```diff
@@ -0,0 +1,18 @@
...
+ * when running atlaspack with parcel bindings, it is possible that the parcel
+ * dylib will be loaded from a node worker thread, which causes a crash on exit.
+ *
+ * This is a workaround to ensure that the parcel dylib is loaded in the main
+ * thread, which fixes the crash.
+ */
+export function loadRustWorkerThreadDylibHack() {
+  try {
+    // $FlowFixMe
+    require('@parcel/rust'); // eslint-disable-line
```
JakeLane, line 14: Might be good to have an env variable killswitch just in case this causes a regression

JakeLane (approved):

packages/core/core/src/rustWorkerThreadDylibHack.js
yamadapc: Added a flag ; default is true

packages/core/feature-flags/src/index.js
```diff
@@ -20,7 +20,8 @@ export const DEFAULT_FEATURE_FLAGS: FeatureFlags = {
   inlineBundlesSourceMapFixes: false,
   conditionalBundlingNestedRuntime: false,
   patchProjectPaths: false,
-  inlineStringReplacementPerf: false,
+  enableRustWorkerThreadDylibHack: false,
+  inlineStringReplacementPerf: true,
```
mattcompiles, line 24: I think you enabled the wrong one by default?

PR #480, opened 2025-04-29
Use less Mac agent time

alshdavid:
## Motivation

To use less MacOS agent time to help free up CI resources

## Changes

## Checklist


.github/workflows/ci.yml
```diff
@@ -107,6 +105,27 @@ jobs:
         with: { components: "clippy, rustfmt" }
       - run: cargo fmt --all -- --check
       - run: cargo clippy -- -D warnings
+      - run: cargo test --target x86_64-unknown-linux-gnu
```
mattcompiles, line 108: Why did you add this?

.github/workflows/ci.yml
```diff
@@ -107,6 +105,27 @@ jobs:
...
+        - name: 🍎 Unit Test MacOS
+          os: macos-15
+          target: aarch64-apple-darwin
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust
+        uses: ./.github/actions/rust-toolchain
+      - run: cargo test --target ${{ matrix.target }}
```
mattcompiles, line 128: What's the benefit in supplying the target here?

.github/workflows/ci.yml
```diff
@@ -107,6 +105,27 @@ jobs:
...
+  unit_tests_rust:
+    strategy:
+      matrix:
+        include:
+        # Linux AMD64
+        - name: 🐥 Unit Test Linux
+          os: ubuntu-22.04
+          target: x86_64-unknown-linux-gnu
+        # MacOS ARM64
+        - name: 🍎 Unit Test MacOS
```
mattcompiles, line 119: Can we name theses consistently with the other unit tests and mention that they're rust?

.github/workflows/ci.yml
alshdavid: Initially I thought it would avoid a double build but it's not needed - cargo will build again any way. Updated to remove this

.github/workflows/ci.yml
alshdavid: Updated

mattcompiles (approved):

PR #482, opened 2025-04-30
APVM: Adding sources

alshdavid:
## Changes

Adding in the source code for apvm. I expect there will be stuff in here that may not be required - happy to discuss/discard parts.

```
Usage: apvm [OPTIONS] [APVM_DIR] [APVM_LOCAL] <COMMAND>

Commands:
  default    Set the default version of Atlaspack
  install    Install a version of Atlaspack
  link       Helpers to work with node_modules
  list       List installed versions of Atlaspack
  reinstall  Reinstall a previously installed version of Atlaspack
  status     Print info on the current status
  uninstall  Uninstall a previously installed version of Atlaspack
  version    Version information
  atlaspack  Run command with specified version of atlaspack
  help       Print this message or the help of the given subcommand(s)

Arguments:
  [APVM_DIR]    [default value: "$HOME/.local/apvm"] [env: APVM_DIR=]
  [APVM_LOCAL]  [env: APVM_ATLASPACK_LOCAL=/mnt/data/Development/atlassian-labs/atlaspack]

Options:
  -D, --debug  Log all of the things
  -h, --help   Print help
```

### Usage

```bash
# Install apvm

npm install -g @atlaspack/apvm
npm install --save-dev @atlaspack/apvm
yarn add -D @atlaspack/apvm

# Link a version of Atlaspack (Will autoinstall)
apvm link 2.15.0

# Install a version of Atlaspack without linking it
apvm install 2.16.0

# List the versions installed on the system
apvm list

# Get current status
apvm status
```

### Config

Configure versions with aliases in the `package.json`

```json
// package.json
{
  "name": "my-package",
  // ...
  "atlaspack": {
    "versions": {
      "default": "2.15.0",
      "next": "2.16.0"
    }
  }
}
```

``` bash
apvm link default # Install & Link 2.15.0
apvm link next    # Install & Link 2.16.0
```

### Try it out locally

```bash
cd atlaspack
yarn install
yarn build-native-release

# Via npm
npx apvm help

# Or use the bin directly
export PATH="$PWD/target/x86_64-unknown-linux-gnu/release:$PATH" # Linux
export PATH="$PWD/target/aarch64-apple-darwin/release:$PATH"     # MacOS

apvm help
```

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,9 @@ rust_2018_idioms = { priority = -1, level = "deny" }
 
 [workspace.lints.clippy]
 module_inception = "allow"
+manual_map = "allow"
+useless_format = "allow"
+for_kv_map = "allow"
 # all = { priority = -1, level = "deny" }
 # restriction = { priority = -1, level = "deny" }
 # pedantic = { priority = -1, level = "deny" }
@@ -33,6 +36,7 @@ aho-corasick = "1.1.3"
 anyhow = "1.0.95"
 assert_fs = "1.1.2"
 async-trait = "0.1.85"
+base64 = "0.22.1"
 bitflags = "2.6.0"
 browserslist-rs = "0.17.0"
 cfg-if = "1.0.0"
@@ -46,12 +50,15 @@ data-encoding = "2.6.0"
 derive_builder = "0.20.2"
 dunce = "1.0.5"
 dyn-hash = "0.2.0"
+env_logger = "0.11.8"
+flate2 = "1.1.1"
 getrandom = { version = "0.2.15", default-features = false }
 git2 = { version = "0.20.0", default-features = false }
 glob = "0.3.2"
 glob-match = "0.2.1"
 heed = "0.21.0"
 hex = "0.4.3"
+homedir = "0.3.4"
 html5ever = { git = "https://github.com/servo/html5ever", rev = "a831e82fcee980d80f53699dc14bdfc39a17dc5f" }
 image = "0.25.5"
 indexmap = "2.7.0"
@@ -89,6 +96,7 @@ pretty_assertions = "1.4.1"
 rand = "0.8.5"
 rayon = "1.10.0"
 regex = "1.11.1"
+reqwest = { version = "0.12.15", default-features = false  }
 sentry = { version = "0.35.0", default-features = false, features = [
   "anyhow",
   "backtrace",
@@ -108,10 +116,12 @@ serde_yml = "0.0.12"
 serde-bool = "0.1.3"
 serde-value = "0.7.0"
 sha-1 = "0.10.1"
+sha2 = "0.10.9"
 sourcemap = "9.1.2"
 swc_core = "10.0.0"
 swc_ecma_parser = "6.0.1"
 swc_ecma_transforms_testing = "7.0.0"
+tar = "0.4.44"
 tempfile = "3.17"
 thiserror = "2.0.9"
 thread_local = "1.1.8"
@@ -125,6 +135,7 @@ tracing-subscriber = "0.3.19"
 tracing-test = "0.2.5"
 url = "2.5.4"
 url-search-params = "12.0.0"
+which = "7.0.3"
 whoami = "1.5.2"
 xxhash-rust = "0.8.15"
 
--- a/crates/apvm/src/cmd/debug.rs
+++ b/crates/apvm/src/cmd/debug.rs
@@ -1,0 +1,24 @@
+use clap::Parser;
+use clap::Subcommand;
+
+use crate::context::Context;
+
+#[derive(Debug, Subcommand, Clone)]
+pub enum DebugCommandType {}
+
+#[derive(Debug, Parser)]
+pub struct DebugCommand {
+  #[clap(subcommand)]
+  pub query: Option<DebugCommandType>,
+}
+
+// Hidden utilities for debugging
+pub fn main(ctx: Context, cmd: DebugCommand) -> anyhow::Result<()> {
+  match cmd.query {
+    None => {
+      print!("{}", serde_json::to_string_pretty(&ctx)?);
+    }
+  }
+
+  Ok(())
+}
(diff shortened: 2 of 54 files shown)


crates/apvm/src/cmd/debug.rs
alshdavid, line 1:
```
apvm debug
```
These are hidden utilities I use for debugging

crates/apvm/src/cmd/default.rs
alshdavid, line 1:
```
apvm default <version>
```
Given our flow is mostly based off of linking into `node_modules`, this command probably isn't needed. Though having a system default/global version set makes sense to me

crates/apvm/src/cmd/install.rs
alshdavid, line 1:
```
apvm install
apvm install <version>
```
This will install an Atlaspack version (but not link it). If the version is omitted then it will try to get it from the `package.json#atlaspack.versions` key

crates/apvm/src/cmd/list.rs
alshdavid, line 1:
```
apvm list
```
This give some output on the currently installed versions of Atlaspack, the detected version in the current project and all that good stuff
image

crates/apvm/src/cmd/npm.rs
alshdavid, line 1:
```
apvm npm <command>
```
This is a subcommand that namespaces interactions with node_modules. It made sense to me at the start to seperate it but I am open to opinions here

crates/apvm/src/cmd/npm_link.rs
alshdavid, line 1:
```
apvm npm link
apvm npm link --install <version?>
```
This will link a version of atlaspack into the `node_modules` of the **current working directory**. If the version is omitted then it will try to resolve the version from the config. If `--install` is supplied it will install the version before linking it.

crates/apvm/src/cmd/npm_link.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+fn npm_link_local(
+  _ctx: Context,
+  _cmd: NpmLinkCommand,
+  _package: PackageDescriptor,
+) -> anyhow::Result<()> {
+  println!("TODO");
+  Ok(())
+}
```
alshdavid, line 64: The different project types need different link strategies - but I suspect once super is in play it won't matter and I can link all exactly the same way.

crates/apvm/src/main.rs
alshdavid, line 1: This isn't a concern here and I can roll this part back but one capability this executable has is to proxy commands to a managed version of Atlaspack.

```bash
apvm default 2.15.0
apvm atlaspack build ./src/index.js
```

And if the `apvm` executable is linked+renamed to `atlaspack`, it acts as a proxy that is sensitive to the version configured in the system.

```bash
apvm default 2.15.0
atlaspack build ./src/index.js
```
My thinking was that it allows for instant exits on the Atlaspack process when started from the terminal and potential for better handling of analytics.

crates/apvm/src/paths.rs
alshdavid, line 1: apvm creates a directory structure to manage the installed versions which is, by default (set with `$APVM_DIR`), located in:
```
$HOME/.local/apvm
```
It looks like
```
$HOME/.local/apvm
  /.temp
  /versions
    /npm
      /<name>
    /git
    /local
  /global -> [symlink]
```

crates/apvm/src/cmd/debug.rs
alshdavid, line 1:
```
apvm debug
```
This is a hidden command with an unstable private API used to debug apvm

Cargo.toml
```diff
@@ -81,9 +86,10 @@ pathdiff = "0.2.1"
 percent-encoding = "2.3.1"
 petgraph = "0.7.1"
 pretty_assertions = "1.4.1"
-rand = "0.8.5"
+rand = "0.9.0"
```
mattcompiles, line 89: Do we have to upgrade to `0.9.0` here? Guess it's fine if yes, but would probs be safer to just leave it for now as this will have an effect on the other builds.

crates/apvm/src/cmd/link_npm.rs
```diff
@@ -0,0 +1,151 @@
...
+      node_modules_atlaspack_pkg.join("package.json"),
+      (json::object! {
+        "name": format!("@atlaspack/{file_stem}"),
+        "main": "./index.js",
+        "type": "commonjs"
+      })
+      .pretty(2),
+    )?;
+
+    if file_stem == "runtime-js" {
```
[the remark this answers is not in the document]
mattcompiles: Personally I think we should remove this and deal with it directly in AFM.

Cargo.toml
alshdavid: Yes, there was a breaking change in this minor version (lol) that I depend on. I guess I could rewrite it to use the old API if it's contentious

crates/apvm/src/cmd/link_npm.rs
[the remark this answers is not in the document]
alshdavid: @JakeLane this is a temporary workaround for code in AFM that is importing something directly from an inner AFM path. This will be addressed directly in AFM at a later date but this unblocks the rollout in the mean time

crates/apvm/src/cmd/debug.rs
```diff
@@ -0,0 +1,100 @@
...
+  },
+}
+
+#[derive(Debug, Parser)]
+pub struct DebugCommand {
+  #[clap(subcommand)]
+  pub query: Option<DebugCommandType>,
+}
+
+#[rustfmt::skip]
```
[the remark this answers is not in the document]
alshdavid: Mistake, updated

Cargo.toml
alshdavid: This is only used in `lmdb-lite` as far as I can tell. @yamadapc, is this safe to bump?

alshdavid: Nevermind, reverted the version bump and moved to the older API.

mattcompiles (approved):

PR #483, opened 2025-05-01
Create public ESLint plugin 

dddlr:
Create public ESLint plugin, containing two ESLint rules to enforce how conditional imports should be used.

## Motivation

We need to enforce two specific things in how people use conditional imports:

* they should always use type annotations, and
* they should never export the conditional import function call (`importCond`)

ESLint rules are the easiest way to enforce this.

## Changes

* Added new package called `@atlaspack/eslint-plugin-public`
* Updated `typescript-eslint` and `prettier` packages to fix type, test and build issues

## Checklist

- [x] Test in Jira codebase, and make a draft PR to add this plugin to Jira codebase
- [ ] Verify that the package will be deployed correctly, with help from team


packages/core/eslint-plugin-public/src/rules/no-importcond-exports/README.md
```diff
@@ -0,0 +1,60 @@
...
+👍 Examples of **correct** code for this rule:
+
+```ts
+// `importCond` is used within the file but not directly exported
+const MyComponent = importCond<
+  typeof import('./old.tsx'),
+  typeof import('./new.tsx')
+>('gate-name', './old.tsx', './new.tsx');
+
+export function getComponent() {
```
AllySummers, line 29: `context:`

i’m curious as to why it’s considered bad practice to directly export something using `importCond` if it does work in production?

i could be wrong, but i feel like having a feature gate to control which component is used by dependents would be a semi common pattern, and having the additional function call to get the component makes code harder to read/understand, and puts us in a position where people may use wildly different names for a component, instead of at least needing to reference the component name?

e.g.:

```ts
/// @filename: foo-component.tsx
const FooComponent = importCond<typeof import('./old.tsx'), typeof import('./new.tsx')>(
    'gate-name',
    './old.tsx',
    './new.tsx',
);
export const getFooComponent = () => FooComponent;
/// @filename: bar-component.tsx
import { getFooComponent } from './foo-component';
// we can rename it here
const BarComponent = getFooComponent();
```

packages/core/eslint-plugin-public/src/utils/index.ts
```diff
@@ -0,0 +1,11 @@
+import {ESLintUtils} from '@typescript-eslint/utils';
+
+export interface ExamplePluginDocs {
```
AllySummers, line 3: `nit:`

was the inclusion of the word `Example` left in from docs, or was it intentional?

packages/core/eslint-plugin-public/test/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+import {RuleTester} from '@typescript-eslint/rule-tester';
+import path from 'node:path';
+
+RuleTester.afterAll = mocha.after;
+
+export const tsRuleTester = new RuleTester({
+  languageOptions: {
+    parser: tseslint.parser,
+    parserOptions: {
+      projectService: {
```
AllySummers, line 12: `question:` / `suggestion:`

is type information required to run this rule? if not, we can probably remove the tsconfig/project services options to make running unit tests a bit faster?

packages/core/eslint-plugin-public/package.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "dependencies": {
+    "@typescript-eslint/utils": "^8.31.1"
+  },
+  "devDependencies": {
+    "typescript-eslint": "^8.31.1",
+    "@typescript-eslint/rule-tester": "^8.31.1"
+  },
+  "peerDependencies": {
+    "eslint": ">8.0.0"
```
AllySummers, line 26: `suggestion:`

the semver `>8.0.0` will match anything greater than `8.0.0`, including `10.0.0` (which could have major breaking changes).

it would be considerably safer to use the range `^8.57.0 || ^9.0.0` to ensure that they only use a version that supports all the new properties in eslint rules (the ones mentioned here https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/ - while they are able to be used in in i believe `8.49.0`, there _might_ be some other things added between the versions? i can't remember)

package.json
```diff
@@ -64,7 +64,7 @@
     "@types/node": ">= 18",
     "buffer": "mischnic/buffer#b8a4fa94",
     "cross-env": "^7.0.0",
-    "eslint": "^8.41.0",
+    "eslint": "^8.57.0",
```
AllySummers, line 67: `suggestion:`

i would recommend using (or at least testing with) eslint v9 for the eslint plugin, even if it's a separate test (since that is what jira is using) if possible? :)

AllySummers: i realise it's probably difficult to update the whole repo to using eslint v9, and likely not in scope for this task, because `eslint-plugin-flowtype` and `eslint-plugin-monorepo` haven't been updated in ages, but i thought it might be useful to mention?

you may be able to get away with using [@eslint/compat](https://www.npmjs.com/package/@eslint/compat) and using `fixupPluginRules` though?

.changeset/fair-sheep-sing.md
JakeLane, line 1: I think the changesets tool is broken, best to remove all of these except the eslint plugin changes

packages/core/eslint-plugin-public/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@atlaspack/eslint-plugin-public",
```
JakeLane, line 2: I think better to call it `@atlaspack/eslint-plugin` and make the internal one named `internal` or `private`.

packages/core/eslint-plugin-public/src/rules/importcond-type-annotations/index.ts
```diff
@@ -0,0 +1,122 @@
...
+            sourceCode.getText(typeArguments.params[0]) !==
+              `typeof import(${secondArgText})` ||
+            sourceCode.getText(typeArguments.params[1]) !==
+              `typeof import(${thirdArgText})`
+          ) {
+            // Type annotation is present, but incorrect
+            context.report({
+              node,
+              messageId: 'wrongTypeAnnotation',
+              // We should be more cautious here, so we make it a suggestion instead of an autofix
```
JakeLane, line 100: Nice one

dddlr:
Screenshot 2025-05-02 at 11 25 23
Screenshot 2025-05-02 at 11 28 20

getting stuck on prettier failing and also making what appear to not be one-to-one changes, even after I reverted the prettier version bump and config changes

I'll add these to `.prettierignore` just to unblock this PR

but I recognise that we will need to update prettier and  at some point (maybe as part of an innovation week project?)

also noting that flow type comments are [no longer supported for prettier](https://github.com/prettier/prettier/pull/13687) and we will need to find an alternative solution: either moving to non-comment type annotations or just migrating to typescript

packages/core/eslint-plugin-public/src/rules/no-importcond-exports/README.md
JakeLane: Exporting `importCond` function calls will cause issues in test environments, so we don't want to encourage them.

As for why we made this decision, we have a couple of reasons:

- Moving the importCond means making the `fg` call not colocated into the relevant file for static analysis
- Another form of barrel files (with similar costs to bundling and bundle size)
- We don't want to encourage people to introduce code like in your example. Exporting an importCond call, even implicitly, means that we can't bundle split as efficiently. Each call to importCond is an opportunity to bundle split.

packages/core/eslint-plugin-public/src/utils/index.ts
dddlr: oops good catch, i was going to resolve this before making the PR before but completely forgot

AllySummers:
@dddlr based on this bit, it might be failing because of newer syntax?

e.g. the `??=` (nullish coalescing assignment operator) is only very new, so it may not support that
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment

Screenshot 2025-05-02 at 11 50 39 am

packages/core/eslint-plugin-public/test/utils.ts
dddlr: oh you're right, none of these options are actually needed 👀 

i had added this (and `filename`) because i was running into trouble getting typescript-eslint to work in this repo and this seemed to help, but i can't actually reproduce the issues i was running into anymore 🤔  so they must not have been necessary after all

package.json
dddlr: i will probably leave moving to eslint v9 for a future PR to be honest 😅  as it would involve checking several other packages around the repo (at least, i couldn't complete the version upgrade within a one-hour timebox)

i wonder if it's possible to test with eslint v9 just for the `@atlaspack/eslint-plugin` package? 🤔

dddlr: I did confirm that the tests added in this PR do work locally when i change the whole repo to eslint v9, so integrating into jira would be fine 😅

.changeset/fair-sheep-sing.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atlaspack/eslint-plugin-public': patch
```
mattcompiles, line 2: This should be `@atlaspack/eslint-plugin` right?

dddlr: gah i knew i forgot something

packages/core/eslint-plugin/src/rules/importcond-type-annotations/README.md
```diff
@@ -0,0 +1,49 @@
+# Use correct type annotations for `importCond` (atlaspack/importcond-type-annotations)
+
+This rule ensures that all `importCond` function calls have the correct type annotations.
+
+## Rule Details
+
+The `importCond` function is used for conditional imports, which allows developers to switch between
+two modules or components based on a feature gate value. It is a more performant alternative to
+`componentWithFG` as the user only loads the module or component matching the feature gate value,
```
JakeLane, line 9: This is a jira specific feature, maybe drop this context from the README given we can support any atlassian product in the future.

packages/core/eslint-plugin/src/rules/no-importcond-exports/index.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+      },
+      ExportNamedDeclaration(node) {
+        // Check if any named export references a tracked variable
+        if (node.specifiers) {
+          node.specifiers.forEach((specifier) => {
+            if (
+              specifier.type === 'ExportSpecifier' &&
+              specifier.local.type === 'Identifier' &&
+              importCondVariables.has(specifier.local.name)
```
JakeLane, line 53: will this be an ordering issue?

JakeLane: As in, is it posible for `importCondVariables` to not yet contain the variable used?

dddlr: good question

looks like default exports would be fine, as typescript enforces against ordering issues

Screenshot 2025-05-06 at 16 39 20

but named exports can run into ordering issues yeah - tbh i don't expect this to be a common syntax, but it'd be good for us to handle it anyway

Screenshot 2025-05-06 at 16 40 36

JakeLane: Yeah, even though it's not realistic to be visible in code, it's best to follow the correct implementation in ESLint (entry and exit phases).

JakeLane (approved):

PR #484, opened 2025-05-01
Super: copy over types

alshdavid:
## Motivation

Provide types from the super package

## Changes

- Add script to `@atlaspack/super` that copies project types into `super/types` and rewrites the import specifiers from `@atlaspack/*` to their respective relative paths.
- Added `exports` key to `@atlaspack/super` to route types and import paths
- Fixed type issue with `@atlaspack/fs`

image

## Checklist

[no-changeset]: Does this need one?


packages/core/super/package.json
```diff
@@ -25,6 +26,163 @@
...
+      "types": "./types/core/types/lib/index.d.ts"
+    },
+    "./utils": "./lib/utils.js",
+    "./watcher-watchman-js": "./lib/watcher-watchman-js.js",
+    "./worker": "./lib/worker.js",
+    "./workers": {
+      "default": "./lib/workers.js",
+      "types": "./types/core/workers/index.d.ts"
+    }
+  },
```
alshdavid, line 185: We should probably generate this from the entries list

packages/core/super/scripts/copy-types.mjs
```diff
@@ -0,0 +1,82 @@
...
+// Analyze repo to resolve paths to "package.json#types"
+for (const pkg of packageJsonFiles) {
+  const pkgDir = path.dirname(pkg);
+
+  const pkgJson = JSON.parse(
+    fs.readFileSync(path.join(__root, 'packages', pkg), 'utf8'),
+  );
+
+  // You cannot import "d.ts" extensions, must import ".js"
+  if (pkgJson.types) {
```
mattcompiles, line 40: I think we should just inline this in `generate-entries` as it already decides which packages should be added to the super package and loads their package.json.

packages/core/super/package.json
mattcompiles: Yeah, we either code gen it, or don't do it at all. Hardcoding will almost certainly lead to a mistake

mattcompiles (approved):

PR #488, opened 2025-05-05
Build release binaries on debian bullseye

yamadapc:
Builds release native add-on on debian bullseye instead of Ubuntu 22.04

https://github.com/atlassian-labs/atlaspack/actions/runs/14826771674/job/41620991987

https://github.com/atlassian-labs/atlaspack/actions/runs/14826833507

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,6 +36,7 @@ jobs:
           # Linux ARM64
           - name: aarch64-unknown-linux-gnu
             target: aarch64-unknown-linux-gnu
+            needs_cross_compile: true
 
           # # Linux ARM
           # We don't support 32 bit linux
@@ -57,6 +58,10 @@ jobs:
           apt-get update
           apt-get install -y build-essential curl nodejs npm
           npm i -g yarn
+      - name: Install cross-compiler
+        if: ${{ matrix.needs_cross_compile }}
+        run: |
+          apt-get install -y gcc-aarch64-linux-gnu
       - name: Install Rust
         uses: ./.github/actions/rust-toolchain
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,13 +36,13 @@ jobs:
 
           # Linux ARM64
           - name: aarch64-unknown-linux-gnu
-            os: ubuntu-24.04-arm
+            os: ubuntu-22.04-arm
             target: aarch64-unknown-linux-gnu
 
           # # Linux ARM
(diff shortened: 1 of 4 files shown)


JakeLane (approved):

.github/workflows/release.yml
```diff
@@ -24,39 +24,83 @@ permissions:
...
           # - name: arm-unknown-linux-gnueabihf
-          #   os: ubuntu-24.04
+          #   os: ubuntu-22.04
           #   target: arm-unknown-linux-gnueabihf
           #   run: apt install -y gcc-arm-linux-gnueabihf
 
+    name: ${{ matrix.name }}
+    runs-on: ubuntu-22.04
+    container:
+      image: debian:bullseye-slim
```
yamadapc, line 51: Need to duplicate this block again because this can't be conditional

.github/workflows/release.yml
```diff
@@ -24,39 +24,83 @@ permissions:
...
           # # Linux ARM
           # We don't support 32 bit linux
           # - name: arm-unknown-linux-gnueabihf
-          #   os: ubuntu-24.04
+          #   os: ubuntu-22.04
           #   target: arm-unknown-linux-gnueabihf
           #   run: apt install -y gcc-arm-linux-gnueabihf
 
+    name: ${{ matrix.name }}
+    runs-on: ubuntu-22.04
```
alshdavid, line 49: We can now use `ubuntu-24.04` on the host because the build happens in a docker container

.github/workflows/release.yml
```diff
@@ -24,39 +24,83 @@ permissions:
...
+        # We need to install nodejs because `setup-node` requires
+        # the runner to have nodejs installed (https://github.com/actions/setup-node/issues/1287)
+        run: |
+          apt-get update
+          apt-get install -y build-essential curl nodejs npm
+          npm i -g yarn
+      - name: Install cross-compiler
+        if: ${{ matrix.needs_cross_compile }}
+        run: |
+          apt-get install -y gcc-aarch64-linux-gnu
```
alshdavid, line 64: This is not required. Can use the `ubuntu-24.04-arm` host + an `debian:bullseye-slim` docker image

.github/workflows/release.yml
yamadapc: A lot of our consumers will be on 22.04, so it might be better to use that for now

alshdavid (approved):

mattcompiles (approved):

PR #496, opened 2025-05-08
Use the new GitHub app to auth changesets

benjervis:
## Motivation

Currently when the Changesets action opens the Version Packages PR, it uses the default `GITHUB_TOKEN` provided by GitHub Actions.

However PRs opened using this token don't have [permission to kick off] CI.

[permission to kick off]: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

## Changes

The [best practices solution] seems to be to make a GitHub app and then use that to authenticate and get a token that is allowed to run CI.

We've gotten the open source admins to make one for us, and now we can use the [official get-a-token action][a] to auth properly.

[best practices solution]: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#granting-additional-permissions
[a]: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow#authenticating-with-a-github-app

## Checklist

[no-changeset]: Updating github actions

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -184,9 +184,6 @@ jobs:
     id: authenticate-changesets
     if: github.ref_name == 'main'
     permissions: write-all
-    needs:
-      - build-native-linux
-      - build-native-macos
     runs-on: ubuntu-22.04
     uses: actions/create-github-app-token@v1
     with:
@@ -197,7 +194,10 @@ jobs:
     name: Build and release Changesets
     if: github.ref_name == 'main'
     permissions: write-all
-    needs: authenticate-changesets
+    needs:
+      - authenticate-changesets
+      - build-native-linux
+      - build-native-macos
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -179,14 +179,26 @@ jobs:
...
     runs-on: ubuntu-22.04
+    uses: actions/create-github-app-token@v1
+    with:
+      app-id: ${{ vars.ATLASPACK_CI_APP_ID }}
+      private-key: ${{ secrets.ATLASPACK_CI_PRIVATE_KEY }}
+
+  build-and-release-changesets:
+    name: Build and release Changesets
+    if: github.ref_name == 'main'
+    permissions: write-all
```
mattcompiles, line 199: I think this step should depend on all three steps. May as well get the token sorted while the previous steps are running.

mattcompiles (approved):

JakeLane (approved):

PR #498, opened 2025-05-09
Fix canary build features

yamadapc:
Pass `--features canary` to canary builds

### DIFF
--- a/scripts/build-native.js
+++ b/scripts/build-native.js
@@ -29,10 +29,6 @@ if (rustProfile !== 'debug') {
   cargoCommand.push('--profile', rustProfile);
 }
 
-if (rustProfile === 'canary') {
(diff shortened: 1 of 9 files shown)


scripts/build-native.js
```diff
@@ -29,6 +29,10 @@ if (rustProfile !== 'debug') {
   cargoCommand.push('--profile', rustProfile);
 }
 
+if (rustProfile === 'canary') {
```
mattcompiles, line 32: Why not just remove the canary profile all together and always build them?

mattcompiles: All good if you just wanna get this out the door quickly. But I think all build types are currently canary so it just creates confusion in the code base.

mattcompiles (approved):

PR #503, opened 2025-05-15
Fix conditional bundling reporter when condition is reused

JakeLane:
## Motivation

@flakeparadigm noticed that his code was not getting preloaded in Jira. We isolated this to reusing the same condition for two different imports. 

## Changes

This change fixes the conditional manifest reporter, which was erroneously overriding the previous condition bundles. It is gated behind a feature flag.

## Checklist

### DIFF
--- a/packages/core/core/src/public/BundleGraph.js
+++ b/packages/core/core/src/public/BundleGraph.js
@@ -31,6 +31,7 @@ import Dependency, {
 import {targetToInternalTarget} from './Target';
 import {fromInternalSourceLocation} from '../utils';
 import BundleGroup, {bundleGroupToInternalBundleGroup} from './BundleGroup';
+import {getFeatureFlag} from '@atlaspack/feature-flags';
 
 // Friendly access for other modules within this package that need access
 // to the internal bundle.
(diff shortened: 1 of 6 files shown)


packages/core/core/src/public/BundleGraph.js
```diff
@@ -489,10 +490,20 @@ export default class BundleGraph<TBundle: IBundle>
       for (let bundle of bundles) {
         const conditions = bundleConditions.get(bundle.id) ?? new Map();
 
+        const currentCondition = conditions.get(cond.key);
+
         conditions.set(cond.key, {
           bundle,
-          ifTrueBundles,
-          ifFalseBundles,
+          ifTrueBundles: getFeatureFlag(
```
mattcompiles, line 497: Not a blocker, but I feel like flagged code is easier to read (and clean up?) when the condition is just checked the once rather than forking inline like this.

mattcompiles (approved):

packages/core/core/src/public/BundleGraph.js
JakeLane: yeah I went back-and-forth on this a bit, as I initially forked it. I'll revert it to forking as I can see why this is hard to read a bit better now.

packages/reporters/conditional-manifest/src/ConditionalManifestReporter.js
```diff
@@ -60,18 +60,41 @@ export async function report({
...
+            bundle.target.distDir,
+            bundle.filePath,
+          );
+
+          const bundleInfo =
+            manifest[bundle.target.name]?.[relativeBundlePath] ?? {};
+
+          bundleInfo[key] = {
+            ifTrueBundles: mapBundles(cond.ifTrueBundles)
+              .concat(bundleInfo[key]?.ifTrueBundles ?? [])
```
liamqma, line 77: I’m new to bundlers, so I apologize if this is a very basic question. Could you provide an example where bundleInfo[key]?.ifTrueBundles might not be an empty array?

I ran [this test](https://github.com/atlassian-labs/atlaspack/pull/503/files#diff-2ca4afc6bf594e3520640e8064f433354fbfff4b9cbe643bc3d8d2d81e6bf8c1R953) locally, and noticed that bundleInfo[key]?.ifFalseBundles is an empty array ([]).

JakeLane: This occurs when there are multiple conditions with the same key. This can occur because each condition will get a new value in the `conditions` array collected from the transformer passes.

If the same bundle (e.g., jira-spa.js) has multiple conditions with the same, we'd end up overwriting the array. This is sometimes fine (and why it wasn't noticed) if the deps are the same. But given that they aren't always, we need to combine.

liamqma: Thanks for the explaination!

PR #511, opened 2025-05-20
Clean-up rolled-out feature-flag

yamadapc:


.changeset/blue-owls-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atlaspack/feature-flags': patch
```
mattcompiles, line 2: Should be core too

mattcompiles (approved):

marcins (approved):

PR #513, opened 2025-05-20
RFC: Unification

alshdavid:
## Motivation

Now that we are moving to the super package, some follow up work on normalizing the repo is required. 

This PR functions as an RFC of what brining the public packages under a unified package could look like (while offering effortless backwards compat for existing users).

This is important to do before getting people to use the super package directly as working through these details will help clarify what public API we want to support from the super package.

It will also simplify the version manager, which currently needs to support the distributed and super package models.

Lastly, it will help to normalize our integration tests, which have to run against both super and legacy.

It features:
- Shims for legacy consumers
- Only type strip the sources (minimal transformation)
  - Better debugging experience 
  - When we move to TypeScript, the output will be 1:1 less types and we can use first party Nodejs TypeScript support without any require hacks for local development and testing
- Vendored dependencies + their types (without a bundler)
  - Uses `node_modules` during development
  - Uses vendored packages during production
  - Long term would be better to use a bundler to combine the vendored packages, but the concept remains the same

## Changes

- Added `@atlaspack/unified` package
  - The bulk of this PR is the plumbing for this package (new package.json, configs, build scripts)
  - The intention is this would eventually be renamed to `atlaspack` and replace the super build
- Added `@atlaspack/diagnostics` to the unified package
  - This is a small change consisting of 5 files being moved + a shim package created
  - This is intended to demonstrate what the repo would look like when unified

## Backwards Compatibility

We cannot break users that depend on the distributed packages approach. To ensure backwards compatibility, shim packages are used. Shim packages are essentially single-file packages that re-export internals from the main package.

##### Example

While the actual code for the `@atlaspack/diagnostic` package resides in the main `atlaspack` package, a consumer trying to access it from `@atlaspack/diagnostic` will (transparently) install the shim package:

```json
// package.json
{
  "name": "@atlaspack/diagnostic",
  "main": "./shim.js",
  "types": "./shim.d.ts",
  "dependencies": {
    "atlaspack": "2.15.0"  // The actual implementation lives here
  }
}
```

```javascript
// shim.js
module.exports = require('atlaspack/diagnostic')
```

```typescript
// shim.d.ts
export * from 'atlaspack/diagnostic'
export {default} from 'atlaspack/diagnostic'
```

The shim package then has its `version` either `linked` or `fixed` against the main `atlaspack` package's `version` via changesets.

This requires no changes for consumers, though we will encourage consumers to migrate onto the unified package as it will be easier to support from the version manager.

## File size and install times

Despite the larger file size, vendoring the dependencies and distributing a "fat" platform-specific package does not negatively impact install times. We have seen install times actually improve with the super package, despite the large file size.

## Next Steps

This PR includes a lot of plumbing so it's small but touches a lot of files. Subsequent PR should be smaller.
The plan would be to migrate one package at a time to the unified package (one PR per package).

Changes to expect in subsequent PRs:
- Move files from their current directory to the unified package
- Shim package replacing the original that re-exports the moved implementation from the main package
- Changing import paths to be relative
- Declare dependencies to vendor

## "I don't like the vendoring approach / we can sort out vendoring later"

While I think it's fine as it is, that's reasonable position. We can unify the packages first and figure out the approach to vendoring after the packages have been combined.

## "What about TypeScript?"

I can migrate the packages to TypeScript at the same time they are unified. I haven't done so because the diff would be larger which makes for a harder review.

I'd be fine with doing both simultaniously, or doing the unification in one pass and typescript conversion in a second pass.

## Usage

```bash
yarn build-native
yarn build

# Supports the old approach via shim
node -e "console.log(require('@atlaspack/diagnostic'))"

# New API 
node -e "console.log(require('atlaspack/diagnostic'))" # Eventually
node -e "console.log(require('@atlaspack/unified/diagnostic'))"

# Use sources (for development)
node \
  --conditions="atlaspack:sources" \
  -r @atlaspack/babel-register \
  -e "console.log(require('@atlaspack/unified/diagnostic'))"
```

For JavaScript (via `import {} from 'atlaspack/*'`)
The suggested public API is:

```bash
atlaspack/build-cache
atlaspack/cache
atlaspack/cli
atlaspack/codeframe
atlaspack/conditional-import-types
atlaspack/core
atlaspack/eslint-plugin
atlaspack/feature-flags
atlaspack/fs
atlaspack/graph
atlaspack/logger
atlaspack/markdown-ansi
atlaspack/package-manager
atlaspack/plugin
atlaspack/profiler
atlaspack/register
atlaspack/rust
atlaspack/types
atlaspack/utils
atlaspack/workers

atlaspack/bundlers/*
atlaspack/compressors/*
atlaspack/configs/*
atlaspack/namers/*
atlaspack/optimizers/*
atlaspack/packagers/*
atlaspack/reporters/*
atlaspack/resolvers/*
atlaspack/runtimes/*
atlaspack/transformers/*
atlaspack/utils/*   # Merged with atlaspack/utils
atlaspack/validators/*
```

Where the contents of `packages/atlaspack/lib` mirror this public API

## Checklist

[no-changeset]: RFC


gulpfile.js
```diff
@@ -56,14 +50,69 @@ class TapStream extends Transform {
...
+    cwd: __dirname,
+    shell: true,
+    stdio: 'inherit',
+  });
+
+  execSync('node ./scripts/unified-build-ts.mjs', {
+    cwd: __dirname,
+    shell: true,
+    stdio: 'inherit',
+  });
```
alshdavid, line 99: This converts the flow types to `d.ts` files

gulpfile.js
```diff
@@ -56,14 +50,69 @@ class TapStream extends Transform {
...
+  cb();
+});
 
-exports.default = exports.build = gulp.parallel(buildBabel, copyOthers);
+gulp.task('unified:post', (cb) => {
+  execSync('node ./scripts/unified-build-vendor.mjs', {
+    cwd: __dirname,
+    shell: true,
+    stdio: 'inherit',
+  });
```
alshdavid, line 93: Includes the vendored packages in the build artifacts. Currently does not use a bundler but it would be better to in future

gulpfile.js
```diff
@@ -86,3 +135,12 @@ function renameStream(fn) {
...
 }
+
+exports.default = exports.build = gulp.series(
+  'clean',
+  'prepare',
+  buildUnified,
+  'unified:post',
+  buildBabel,
+  copyOthers,
+);
```
alshdavid, line 146: Takes 7 seconds to build

scripts/unified-build-ts.mjs
```diff
@@ -0,0 +1,98 @@
+/* eslint-disable no-console */
```
alshdavid, line 1: This looks through `packages/unified/src/**/*.js` and covert them to `packages/unified/lib/**/*.d.ts`

scripts/unified-build-vendor.mjs
```diff
@@ -0,0 +1,207 @@
+/* eslint-disable no-console */
```
alshdavid, line 1: This looks for files matching `package/unified/src/vendor/**/vendor.json` which includes metadata on the packages that should be vendored.

It then copies the direct dependencies to `packages/unified/lib/vendor/**` and recurses through the dependencies of the vendored dependencies and includes them as well.

It rewrites the import specifiers that point to `node_modules` to relative paths.

It's basically a crappy bundler that includes typescript types

packages/unified/src/diagnostic/index.js
```diff
@@ -1,8 +1,8 @@
 // @flow strict-local
 
 import invariant from 'assert';
-import nullthrows from 'nullthrows';
-import {parse, type Mapping} from '@mischnic/json-sourcemap';
+import nullthrows from '../vendor/nullthrows/index.js';
+import {parse, type Mapping} from '../vendor/@mischnic/json-sourcemap/index.js';
```
alshdavid, line 5: Rather than specifiying packages that need to be vendored by their node resolve path, e.g. `@mischnic/json-sourcemap`, the consumer specifies a relative import to the vendored package e.g. `../vendor/@mischnic/json-sourcemap/index.js`.

During development, the file at the destination (`../vendor/@mischnic/json-sourcemap/index.js`) re-exports the package in `node_modules`  however when the package is built, the folder is replaced with the vendored package code

alshdavid:
Closing in favor of https://github.com/atlassian-labs/atlaspack/pull/585

PR #514, opened 2025-05-20
Fix reading file contents on certain checkouts

yamadapc:
Read file contents from git CLI instead of libgit

### DIFF
--- a/crates/atlaspack_vcs/src/lib.rs
+++ b/crates/atlaspack_vcs/src/lib.rs
@@ -285,12 +285,15 @@ fn get_file_contents_at_commit(
     return Ok(None);
   }
 
+  let commit_id = commit.id();
   let contents = Command::new("git")
     .arg("show")
-    .arg(format!("{}:{}", commit.id(), path.display()))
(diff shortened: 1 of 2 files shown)


crates/atlaspack_vcs/src/lib.rs
```diff
@@ -279,17 +279,21 @@ fn get_file_contents_at_commit(
...
+  if tree.get_path(path).is_err() {
+    return Ok(None);
   }
+
+  let contents = Command::new("git")
+    .arg("show")
+    .arg(format!("{}:{}", commit.id(), path.display()))
+    .current_dir(repo.path().parent().unwrap())
+    .output()
+    .map_err(|err| anyhow::anyhow!("Failed to read yarn.lock from git: {err}"))?;
```
marcins, line 293: You're not specifically reading `yarn.lock` in this function, should the error display the actual `path` passed in?

yamadapc: You're right, thank for you for that

marcins (approved):

PR #520, opened 2025-05-21
Modify "large blobs" to be written to the LMDB cache

pancaspe87:
Eliminating large blob writes so that we can improve the time it takes to write to cache

### DIFF
--- a/packages/dev/query/src/index.js
+++ b/packages/dev/query/src/index.js
@@ -128,6 +128,7 @@ export async function loadGraphs(cacheDir: string): Promise<{|
   let assetGraph;
   if (assetGraphBlob) {
     try {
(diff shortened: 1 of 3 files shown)


packages/dev/query/src/index.js
```diff
@@ -128,6 +128,7 @@ export async function loadGraphs(cacheDir: string): Promise<{|
   let assetGraph;
   if (assetGraphBlob) {
     try {
+      // TODO: this should be reviewed when `cachePerformanceImprovements` flag is removed, as we'll be writing files to LMDB cache instead of large blobs
```
pancaspe87, line 131: @yamadapc I think there are other files that need to be reviewed once the large blobs are not longer used, e.g packages/core/core/src/PackagerRunner.js. They shouldn't be affected by these changes, AFAIU

yamadapc (approved):

PR #525, opened 2025-05-21
Fix bugs in the large blob cache changes

yamadapc:
- fix issue where `Buffer` objects were double serialized (use `set(x)` instead of `set(serialize(x))`)
- fix issue where request graph wasn't loaded when the flag was on
- make all large blobs always go through LMDB now regardless of other changes

### DIFF
--- a/packages/core/cache/src/LMDBLiteCache.js
+++ b/packages/core/cache/src/LMDBLiteCache.js
@@ -9,6 +9,7 @@ import {Lmdb} from '@atlaspack/rust';
 import type {FilePath} from '@atlaspack/types';
 import type {Cache} from './types';
 import type {Readable, Writable} from 'stream';
+import {getFeatureFlag} from '@atlaspack/feature-flags';
 
 import stream from 'stream';
 import path from 'path';
@@ -105,7 +106,9 @@ export class LMDBLiteCache implements Cache {
   }
 
   async ensure(): Promise<void> {
-    await this.fsCache.ensure();
+    if (!getFeatureFlag('cachePerformanceImprovements')) {
+      await this.fsCache.ensure();
+    }
     return Promise.resolve();
   }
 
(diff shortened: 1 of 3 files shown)


.changeset/violet-llamas-bake.md
```diff
@@ -0,0 +1,7 @@
+---
+'@atlaspack/cache': minor
+'@atlaspack/core': minor
```
mattcompiles, line 3: Doesn't really matter but this is only a patch for `@atlaspack/core`

crates/lmdb-js-lite/index.js
```diff
@@ -5,325 +5,312 @@
 /* auto-generated by NAPI-RS */
 
 const { existsSync, readFileSync } = require('fs')
-const {join} = require('path');
+const { join } = require('path')
```
mattcompiles, line 8: Did some weird formatting happen here?

packages/core/cache/src/LMDBLiteCache.js
```diff
@@ -105,7 +114,9 @@ export class LMDBLiteCache implements Cache {
   }
 
   async ensure(): Promise<void> {
-    await this.fsCache.ensure();
+    if (!getFeatureFlag('cachePerformanceImprovements')) {
+      await this.fsCache.ensure();
```
mattcompiles, line 118: Why are we adding this in? It was never used in the old `LMDBCache` implementation?

crates/lmdb-js-lite/index.js
yamadapc: This is a generated file

packages/core/cache/src/LMDBLiteCache.js
yamadapc: I am removing this, not adding it

mattcompiles (approved):

PR #529, opened 2025-05-22
Fix warnings in dev-dep-resolver

yamadapc:
Test Plan:
N/A

[no-changeset]: N/A


packages/utils/dev-dep-resolver/src/main.rs
```diff
@@ -24,7 +24,7 @@ fn main() {
     println!("------------ {} -----------", dep);
     let resolved = match cjs_resolver.resolve(dep, &cwd, SpecifierType::Esm).result {
       Ok(res) => res.0,
-      Err(e) => {
+      Err(_e) => {
         #[cfg(debug_assertions)]
         println!("FAILED TO RESOLVE {} {:?}", dep, e);
```
marcins, line 29: How does this line work if `debug_assertions` is enabled? Won't `e` be undefined? (Same below with `err`)

marcins (approved):

PR #530, opened 2025-05-22
Write metadata about the cache into it

yamadapc:
Test Plan: N/A

### DIFF
--- a/packages/core/core/src/RequestTracker.js
+++ b/packages/core/core/src/RequestTracker.js
@@ -1605,6 +1605,18 @@ export default class RequestTracker {
         nodes: undefined,
       });
 
+      await runCacheImprovements(
+        () =>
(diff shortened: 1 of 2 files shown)


JakeLane (approved):

packages/core/core/src/RequestTracker.js
```diff
@@ -1605,6 +1605,18 @@ export default class RequestTracker {
         nodes: undefined,
       });
 
+      await runCacheImprovements(
```
pancaspe87, line 1608: Does this change need to be feature flagged?

yamadapc: this function is the feature-flagging solution ; I had to make some changes - https://github.com/atlassian-labs/atlaspack/pull/525/files

pancaspe87 (approved):

PR #534, opened 2025-05-23
expose VCS timing metrics to analytics

pancaspe87:
Previous PRs introduced timings, but didn't expose the data:
- [PR](https://github.com/atlassian-labs/atlaspack/pull/344) with timing for dirty hashes listings: `dirty_files_execution_time`
- [PR](https://github.com/atlassian-labs/atlaspack/pull/346/files) with timing for yarn state scanning: `yarn_states_execution_time`

This PR exposes the timings info so that we can report them in dev metrics.

Tested locally:
Screenshot 2025-05-23 at 11 46 23 am

### DIFF
--- a/packages/core/fs/src/NodeVCSAwareFS.js
+++ b/packages/core/fs/src/NodeVCSAwareFS.js
@@ -82,7 +82,27 @@ export class NodeVCSAwareFS extends NodeFS {
         // Note: can't use toString() directly, or it won't resolve the promise
         const snapshotFile = await this.readFile(snapshot);
         const snapshotFileContent = snapshotFile.toString();
-        return JSON.parse(snapshotFileContent);
+        const vcsContent = JSON.parse(snapshotFileContent);
+
+        // Expose VCS timing metrics to analytics
+        if (
+          vcsContent.vcsState?.dirtyFilesExecutionTime != null ||
+          vcsContent.vcsState?.yarnStatesExecutionTime != null
+        ) {
+          logger.info({
+            origin: '@atlaspack/fs',
+            message: 'Expose VCS timing metrics',
+            meta: {
+              trackableEvent: 'vcs_timing_metrics',
+              dirtyFilesExecutionTime:
+                vcsContent.vcsState?.dirtyFilesExecutionTime,
+              yarnStatesExecutionTime:
+                vcsContent.vcsState?.yarnStatesExecutionTime,
+            },
+          });
+        }
+
+        return vcsContent;
       },
     );
     let watcherEventsSince = [];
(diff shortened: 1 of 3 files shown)


pancaspe87:
Hey @yamadapc Ticket only mentioned yarn state scanning, but I also include the dirty hash listings. IIRC these were the 2 PRs; let me know if there is any other data you want to include

> We are returning execution times but they don’t get reported anywhere → [Add tracing logs around yarn state scanning by pancaspe87 · Pull Request #346 · atlassian-labs/atlaspack](https://github.com/atlassian-labs/atlaspack/pull/346/files)

packages/core/fs/src/NodeVCSAwareFS.js
```diff
@@ -82,7 +82,27 @@ export class NodeVCSAwareFS extends NodeFS {
         // Note: can't use toString() directly, or it won't resolve the promise
         const snapshotFile = await this.readFile(snapshot);
         const snapshotFileContent = snapshotFile.toString();
-        return JSON.parse(snapshotFileContent);
+        const vcsContent = JSON.parse(snapshotFileContent);
```
pancaspe87, line 85: IIRC `getEventsSince` gets called only when a snapshot exists; just realised that we need to report the same for when writing the snapshot in the first place

packages/core/fs/src/NodeVCSAwareFS.js
```diff
@@ -160,6 +180,18 @@ export class NodeVCSAwareFS extends NodeFS {
         'NodeVCSAwareFS::getVcsStateSnapshot',
         () => getVcsStateSnapshot(gitRepoPath, this.#excludePatterns),
       );
+
+      logger.info({
+        origin: '@atlaspack/fs',
+        message: 'Expose VCS timing metrics on writeSnapshot',
+        meta: {
+          trackableEvent: 'vcs_timing_metrics-writeSnapshot',
+          // $FlowFixMe[prop-missing] Rust type includes these properties
```
pancaspe87, line 189: Didn't manage to solve these errors yet, but I saw `$FlowFixMe` used multiple times across the repo

packages/core/fs/src/NodeVCSAwareFS.js
yamadapc: We should do this on write though rather than on read

So in writeSnapshot

pancaspe87: Yeahhh, I realised this only after. Will remove this

JakeLane (approved):

yamadapc (approved):

PR #535, opened 2025-05-23
Add shared bundle merging feature 

mattcompiles:
Add `sharedBundleMergeThreshold` config option

In apps with lots of dynamic imports, many shared bundles are often removed
from the output to prevent an overload in network requests according to the
`maxParallelRequests` config. In these cases, setting `sharedBundleMergeThreshold` can
merge shared bundles with a high overlap in their source bundles (bundles that share the bundle).
This config trades-off potential overfetching to reduce asset duplication.

The following config would merge shared bundles that have a 75% or higher overlap in source bundles.

```json
{
  "@atlaspack/bundler-default": {
    "sharedBundleMergeThreshold": 0.75
  }
}
```

## Checklist

### DIFF
--- a/packages/bundlers/default/src/idealGraph.js
+++ b/packages/bundlers/default/src/idealGraph.js
@@ -910,6 +910,7 @@ export function createIdealGraph(
       if (!manualSharedMap.has(manualSharedBundleKey)) {
         let firstSourceBundle = nullthrows(
           bundleGraph.getNode(sourceBundles[0]),
+          'Bro are you sure this is the error',
         );
         invariant(firstSourceBundle !== 'root');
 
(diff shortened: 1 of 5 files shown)


JakeLane (approved):
nice one!

packages/bundlers/default/src/idealGraph.js
```diff
@@ -1249,6 +1256,96 @@ export function createIdealGraph(
...
+  function mergeBundles(
+    bundleGraph: IdealBundleGraph,
+    bundleToKeepId: NodeId,
+    bundleToRemoveId: NodeId,
+    assetReference: DefaultMap<Asset, Array<[Dependency, Bundle]>>,
+  ) {
+    let bundleToKeep = nullthrows(bundleGraph.getNode(bundleToKeepId), '18');
+    let bundleToRemove = nullthrows(
+      bundleGraph.getNode(bundleToRemoveId),
+      '19',
```
mgfalzon, line 1268: @mattcompiles Is this hardcoded value supposed to be in there?

PR #539, opened 2025-05-24
Remove left over nullthrows numbers used for troubleshooting

benjervis:
## Motivation

When we were [adding shared bundle merging](https://github.com/atlassian-labs/atlaspack/pull/535), we were getting some nullthrows errors in the build but it wasn't clear where from. We added numbers to each instance of nullthrows, but forgot to take a few of them out.

## Checklist

[no-changeset]: Tweaks to unreleased code

### DIFF
--- a/.changeset/rare-countries-chew.md
+++ b/.changeset/rare-countries-chew.md
@@ -1,5 +1,0 @@
----
-'@atlaspack/bundler-default': patch
----
-
-Removed some unhelpful identifiers from internal error conditions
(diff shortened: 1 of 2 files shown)


.changeset/rare-countries-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atlaspack/bundler-default': patch
```
mattcompiles, line 2: I wouldn't even add a changeset as this is just tweaking unreleased code

JakeLane (approved):

marcins (approved):

PR #541, opened 2025-05-26
Add database compaction command

yamadapc:
Adds a command to copy over a database compacting free pages.

This should make the database files more space efficient.

Test Plan: yarn test:unit

### DIFF
--- a/packages/core/cache/src/LMDBLiteCache.js
+++ b/packages/core/cache/src/LMDBLiteCache.js
@@ -10,17 +10,17 @@ import {Lmdb} from '@atlaspack/rust';
 import type {FilePath} from '@atlaspack/types';
 import type {Cache} from './types';
 import type {Readable, Writable} from 'stream';
-
+import fs from 'fs';
+import ncp from 'ncp';
+import {promisify} from 'util';
 import stream from 'stream';
 import path from 'path';
-import {promisify} from 'util';
-
 import {NodeFS} from '@atlaspack/fs';
-
 // $FlowFixMe
 import packageJson from '../package.json';
-
 import {FSCache} from './FSCache';
+
+const ncpAsync = promisify(ncp);
 
 interface DBOpenOptions {
   name: string;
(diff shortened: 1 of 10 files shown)


packages/core/cache/src/LMDBLiteCache.js
```diff
@@ -241,13 +243,25 @@ export class LMDBLiteCache implements Cache {
...
-    // the next read will see the latest changes to the store.
-    // Useful in scenarios where reads and writes are multi-threaded.
-    // See https://github.com/kriszyp/lmdb-js#resetreadtxn-void
-    this.store.resetReadTxn();
+  async compact(targetPath: string): Promise<void> {
+    await fs.promises.mkdir(targetPath, {recursive: true});
+
+    const files = await fs.promises.readdir(this.dir);
+    // copy all files except data.mdb and lock.mdb to the target path (recursive)
+    for (const file of files) {
```
marcins, line 251: I'm not sure of the use case for this compacting change (during startup? for debugging?) - but this could be more efficient with a `Promise.all` or similar. Though if we're talking about milliseconds off the critical path though it doesn't matter.

packages/core/cache/src/LMDBLiteCache.js
```diff
@@ -241,13 +243,25 @@ export class LMDBLiteCache implements Cache {
...
+        continue;
+      }
+
+      await ncpAsync(filePath, path.join(targetPath, file));
+    }
+
+    this.store.compact(path.join(targetPath, 'data.mdb'));
   }
+
+  refresh(): void {}
```
marcins, line 264: Why was this code removed in this change?

I remember implementing this years back to fix a race condition in packaging - is it no longer required?

marcins (approved):

packages/core/cache/src/LMDBLiteCache.js
yamadapc: This isn't required with the new back-end and was just no-oping

packages/core/cache/src/LMDBLiteCache.js
yamadapc: We'll do this on a background job building caches so it should be okay to do sequential copy here.

JakeLane (approved):

PR #545, opened 2025-05-26
Packager optimisation around getting referenced assets

yamadapc:
When running packaging, up to roughly 30% of the time on some threads is spent
on `ScopeHoistingPackager::loadAssets`.

In this function we are trying to determine, for each asset in a bundle,
whether it is "referenced" by the bundle.

A referenced asset is an asset that is part of the bundle's sub-tree asset
graph, but does not have a "contains" edge against the bundle.

The current algorithm for determining this will:

* For each bundle being packaged
  * For each asset
    * Check if asset is referenced by bundle
      * Check bundle edges against this asset
      * Get all bundles in this bundle's bundle group
      * Recursive - Traverse the bundle graph starting at each of these
        * Check candidate bundle edges against this asset

On certain bundles in an application (I suppose root bundles), this takes up to
2 minutes (out of 5min total runtime on my machine for all 6k bundles).

In this commit, I am prototyping a few changes to this. First, we do
reference checking in bulk.

* For each bundle being packaged
  * Before starting `loadAssets`
    * Build Set of all referenced assets for this bundle
      * Get all candidate assets to check
      * Get all bundles in this bundle's bundle group
      * Traverse the bundle graph starting at each of these
        * List directly referenced assets from this bundle (not recursive)
        * Cache this list for all other packaging in the build
      * Check if any directly referenced asset was a candidate
  * For each asset in bundle
    * Check if asset is in Set built on previous step

I am still measuring / validating my changes.

Test Plan: pending

### DIFF
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -1284,6 +1284,143 @@ export default class BundleGraph {
     );
   }
 
+  getReferencedAssets(
+    bundle: Bundle,
+    cache: Map<Bundle, Set<string>>,
+  ): Set<string> {
+    const result = new Set();
+
+    const siblingBundles = new Set(
+      this.getBundleGroupsContainingBundle(bundle).flatMap((bundleGroup) =>
+        this.getBundlesInBundleGroup(bundleGroup, {includeInline: true}),
+      ),
+    );
+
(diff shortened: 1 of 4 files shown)


JakeLane (commented):
Should this be feature flagged?

PR #552, opened 2025-05-27
Clean-up quadratic cache invalidation ff

yamadapc:
This change has long been rolled-out internally.

There are broken tests which weren't addressed.

We need to fix these on a separate change.

Test Plan: yarn test:unit & yarn test:integration

### DIFF
--- a/packages/core/core/test/RequestTracker.test.js
+++ b/packages/core/core/test/RequestTracker.test.js
@@ -453,93 +453,87 @@ describe('RequestTracker', () => {
   });
 
   describe('respondToFSEvents', () => {
-    [true, false].forEach((value) => {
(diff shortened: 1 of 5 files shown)


packages/core/core/test/RequestTracker.test.js
```diff
@@ -453,93 +453,87 @@ describe('RequestTracker', () => {
...
-                  '',
-                  './node_modules/something/package.json',
-                ),
-              });
-              let result = await Promise.resolve('hello');
-              api.storeResult(result);
-            },
-            input: null,
+    /**
+     * This test-case has been missed and invalidation is broken on this case.
```
yamadapc, line 500: Note this

PR #555, opened 2025-05-28
Native: Added ATLASPACK_NATIVE_THREADS

alshdavid:
## Motivation

We need the ability to control the number of threads used by Tokio in the native context

## Changes

Added an environment variable `ATLASPACK_NATIVE_THREADS` to control the number of Tokio threads

## Checklist

### DIFF
--- a/packages/core/core/src/Atlaspack.js
+++ b/packages/core/core/src/Atlaspack.js
@@ -165,10 +165,17 @@ export default class Atlaspack {
       const version = require('../package.json').version;
       await lmdb.put('current_session_version', Buffer.from(version));
 
+      let threads = undefined;
+      if (process.env.ATLASPACK_NATIVE_THREADS !== undefined) {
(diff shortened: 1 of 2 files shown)


packages/core/core/src/Atlaspack.js
```diff
@@ -165,10 +165,17 @@ export default class Atlaspack {
       const version = require('../package.json').version;
       await lmdb.put('current_session_version', Buffer.from(version));
 
+      let threads = undefined;
```
marcins, line 168: Is the default if undefined just CPUs - 1 or a similar heuristic?

yamadapc (approved):

packages/core/core/src/Atlaspack.js
mattcompiles: Yes, but that's calculated in the native code.

PR #558, opened 2025-05-28
Fix bug cache where deleted files cause snapshot to fail

yamadapc:
Allows files to be missing when deleted.

### DIFF
--- a/crates/atlaspack_vcs/src/lib.rs
+++ b/crates/atlaspack_vcs/src/lib.rs
@@ -267,11 +267,18 @@ pub struct YarnSnapshot {
   pub yarn_state: YarnStateFile,
 }
 
+/// "Dirty" files are files modified in the current work-tree and uncommitted.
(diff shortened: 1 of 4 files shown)


crates/atlaspack_vcs/src/lib.rs
```diff
@@ -625,7 +632,7 @@ pub fn vcs_list_dirty_files(
     return Err(anyhow::anyhow!("Git ls-files failed:\n{}", stderr));
   }
   let output = String::from_utf8(command.stdout).unwrap();
-  let lines: Vec<_> = output.split_terminator('\0').map(String::from).collect();
+  let lines: HashSet<_> = output.split_terminator('\0').map(String::from).collect();
```
matt-koko, line 635: any concern for a collision if lines/files are the same?

mattcompiles (approved):

PR #559, opened 2025-05-28
Fix a few VCS cache invalidation bugs

yamadapc:
Fix bug where switching backwards in history would not properly report changes.

Fix bug where duplicate events were logged for diffed, dirty or previously
(snapshot) dirty files.

### DIFF
--- a/.changeset/wild-brooms-hide.md
+++ b/.changeset/wild-brooms-hide.md
@@ -1,0 +1,6 @@
+---
+'@atlaspack/rust': patch
+'@atlaspack/fs': patch
+---
+
+Fix bugs in VCS cache invalidation
--- a/crates/atlaspack_vcs/src/lib.rs
+++ b/crates/atlaspack_vcs/src/lib.rs
@@ -400,6 +400,9 @@ fn get_diff_with_git_cli(
   let output = Command::new("git")
     .arg("diff")
     .arg("--name-status")
+    // We need to list all changes even if `new_commit` is an ancestor of `old_commit`
+    // https://git-scm.com/docs/git-diff
+    // https://git-scm.com/docs/gitrevisions
     .arg(format!("{}..{}", old_commit.id(), new_commit.id()))
     .current_dir(repo_path)
     .output()?;
--- a/crates/atlaspack_vcs/src/lib.rs
+++ b/crates/atlaspack_vcs/src/lib.rs
@@ -393,7 +393,7 @@ fn get_diff_with_git_cli(
   let output = Command::new("git")
     .arg("diff")
     .arg("--name-status")
-    .arg(format!("{}...{}", old_commit.id(), new_commit.id()))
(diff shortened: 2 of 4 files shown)


.changeset/wild-brooms-hide.md
```diff
@@ -0,0 +1,6 @@
+---
+'@atlaspack/rust': patch
```
pancaspe87, line 2: Are the packages in this changeset correct?

yamadapc: Yes!

packages/core/integration-tests/test/vcs-cache.js
```diff
@@ -531,27 +524,20 @@ describe('vcs cache', () => {
...
-      assert.equal(changes[2].path, path.join(lodashDir, 'src', 'file2.js'));
-      assert.equal(changes[2].type, 'delete');
-
-      assert.equal(changes[3].path, path.join(lodashDir, 'src', 'file2.js'));
-      assert.equal(changes[3].type, 'create');
-
-      assert.equal(changes[4].path, path.join(root, 'yarn.lock'));
-      assert.equal(changes[4].type, 'update');
+      assert.deepEqual(changes, [
+        {path: path.join(lodashDir, 'file1.js'), type: 'delete'},
```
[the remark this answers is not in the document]
yamadapc: For node_modules updates we do `delete` and `create` since we don't really know if the file was present before or not. Possibly `update` would be fine.

crates/atlaspack_vcs/src/lib.rs
```diff
@@ -400,7 +400,7 @@ fn get_diff_with_git_cli(
   let output = Command::new("git")
     .arg("diff")
     .arg("--name-status")
-    .arg(format!("{}...{}", old_commit.id(), new_commit.id()))
+    .arg(format!("{}..{}", old_commit.id(), new_commit.id()))
```
csowden-atlassian, line 403: optional: consider comments on why `..` is used over `...`

csowden-atlassian (approved):

crates/atlaspack_vcs/src/lib.rs
yamadapc: This might actually still be wrong and should be `git diff commit1 commit2`

yamadapc: Actually this should be fine.

The problem was using `...` syntax meant if we had:
```
git diff snapshot...current
```

and `current` was behind `snapshot` then the diff would be empty as no commits in `current` are unreachable from `snapshot`, but we must invalidate all of the differences regardless

.changeset/wild-brooms-hide.md
yamadapc: The `fs` package is "functionally" changed, even though it isn't touched here, because it is what consumes this diff.

PR #562, opened 2025-05-29
Fix packager writing big files to cache

yamadapc:
Write packager outputs to disk instead of LMDB. This will work-around issues
with the cache potentially growing too big.

atlaspack-query needs to be supported on a subsequent diff.

There are a few issues related to MemoryFS that this change has uncovered and
fixed.

### DIFF
--- a/crates/atlaspack/src/atlaspack.rs
+++ b/crates/atlaspack/src/atlaspack.rs
@@ -192,14 +192,14 @@ mod tests {
   use atlaspack_core::types::{Asset, Code};
   use atlaspack_filesystem::in_memory_file_system::InMemoryFileSystem;
   use atlaspack_plugin_rpc::{MockRpcFactory, MockRpcWorker};
-  use lmdb_js_lite::{DatabaseWriterError, LMDBOptions};
+  use lmdb_js_lite::{get_database, LMDBOptions};
 
   use super::*;
 
   #[test]
   fn build_asset_graph_commits_assets_to_lmdb() -> Result<(), anyhow::Error> {
     // TODO: Create overlay fs for integration test
-    let db = Arc::new(create_db()?);
+    let db = create_db()?;
     let fs = InMemoryFileSystem::default();
 
     fs.write_file(
@@ -240,16 +240,16 @@ mod tests {
 
     atlaspack.commit_assets(assets.iter().collect())?;
 
-    let txn = db.read_txn()?;
+    let txn = db.database().read_txn()?;
     for (idx, asset) in assets_names.iter().enumerate() {
-      let entry = db.get(&txn, &idx.to_string())?;
+      let entry = db.database().get(&txn, &idx.to_string())?;
       assert_eq!(entry, Some(asset.to_string().into()));
     }
 
     Ok(())
   }
 
-  fn create_db() -> Result<DatabaseHandle, DatabaseWriterError> {
+  fn create_db() -> anyhow::Result<Arc<DatabaseHandle>> {
     let path = temp_dir().join("atlaspack").join("asset-graph-tests");
     let _ = std::fs::remove_dir_all(&path);
 
--- a/crates/lmdb-js-lite/src/writer.rs
+++ b/crates/lmdb-js-lite/src/writer.rs
@@ -18,13 +18,13 @@ type Result<R> = std::result::Result<R, DatabaseWriterError>;
 #[derive(thiserror::Error, Debug)]
 pub enum DatabaseWriterError {
   #[error("heed error: {0}")]
-  HeedError(#[from] heed::Error),
+  Heed(#[from] heed::Error),
   #[error("IO error: {0}")]
-  IOError(#[from] std::io::Error),
+  IO(#[from] std::io::Error),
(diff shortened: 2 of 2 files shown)


packages/core/core/src/UncommittedAsset.js
```diff
@@ -296,6 +303,12 @@ export default class UncommittedAsset {
   }
 
   getCacheKey(key: string): string {
+    if (getFeatureFlag('cachePerformanceImprovements')) {
+      let filePath = fromProjectPathRelative(this.value.filePath);
+      // replace ./../something, ../something with $$atlaspack__parent_dir$$/something
```
mattcompiles, line 308: Is this a todo?

yamadapc: It's done inside `LmdbLiteCache` now ;

packages/core/feature-flags/src/index.js
```diff
@@ -20,7 +20,7 @@ export const DEFAULT_FEATURE_FLAGS: FeatureFlags = {
   inlineBundlesSourceMapFixes: false,
   conditionalBundlingNestedRuntime: false,
   patchProjectPaths: false,
-  cachePerformanceImprovements: false,
+  cachePerformanceImprovements: process.env.NODE_ENV === 'test' ? true : false,
```
yamadapc, line 23: Tests are running with flag turned on, except cache tests which are running for both sides

packages/core/core/test/test-utils.js
```diff
@@ -3,18 +3,12 @@
...
-import tempy from 'tempy';
-import {inputFS, outputFS} from '@atlaspack/test-utils';
+import {inputFS, outputFS, cache, cacheDir} from '@atlaspack/test-utils';
 import {relativePath} from '@atlaspack/utils';
 import {NodePackageManager} from '@atlaspack/package-manager';
 import {createEnvironment} from '../src/Environment';
 import {toProjectPath} from '../src/projectPath';
 
-let cacheDir = tempy.directory();
-export let cache: FSCache = new FSCache(outputFS, cacheDir);
```
yamadapc, line 15: All tests will now run against LMDB database instead of FSCache

packages/core/fs/src/MemoryFS.js
```diff
@@ -497,6 +497,7 @@ export class MemoryFS implements FileSystem {
   }
 
   createWriteStream(filePath: FilePath, options: ?FileOptions): WriteStream {
+    this.mkdirp(path.dirname(filePath));
```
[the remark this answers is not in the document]
yamadapc: In the context this is in (MemoryFS) it's not worth trying to address this

The directory will be created synchronously, but worker events are async.

packages/core/core/src/RequestTracker.js
```diff
@@ -1499,130 +1499,139 @@ export default class RequestTracker {
...
     let total = 0;
-    report({
-      type: 'cache',
-      phase: 'start',
-      total,
-      size: this.graph.nodes.length,
-    });
+    await runCacheImprovements(
+      async (cache) => {
+        await cache.getNativeRef().startWriteTransaction();
```
yamadapc, line 1517: I moved the write transaction into the try/catch because failing to commit the transaction will cause the database to be locked.

We need to commit/abort the transaction always. I willchange the LMDB wrapper API to make it impossible to permanently lock the DB.

packages/core/core/src/PackagerRunner.js
```diff
@@ -700,20 +716,27 @@ export default class PackagerRunner {
...
     }
 
-    let mapExists = await this.options.cache.has(mapKey);
+    let mapExists = getFeatureFlag('cachePerformanceImprovements')
+      ? await this.options.cache.hasLargeBlob(mapKey)
+      : await this.options.cache.has(mapKey);
 
     return {
       contents: isLargeBlob
         ? this.options.cache.getStream(contentKey)
```
yamadapc, line 732: I'm unsure about this part.
I'm changing all package reads to now stream, whereas previously they'd be read from LMDB at once mostly.

I think that should be okay, but possibly we should just read the entire file eagerly.

packages/core/cache/src/LMDBLiteCache.js
```diff
@@ -193,39 +210,36 @@ export class LMDBLiteCache implements Cache {
...
     }
-    return Promise.resolve(this.getBlobSync(key));
+    return fs.promises.readFile(this.getFileKey(key));
   }
 
-  /**
-   * @deprecated Use setBlob instead.
-   */
-  setLargeBlob(
+  async setLargeBlob(
```
yamadapc, line 231: I think it's really bad to have multiple types of entries and methods to access them:

* raw buffer objects
* v8 serialized objects
* compressed entries vs. uncompressed entries
* file entries (large blobs)
* LMDB entries

It'd be much if there was a more cohesive set of entries. Or at least that if a developer makes a mistake like:
```
cache.set(key, buffer); // this needs to be `setBlob`
cache.getBlob(key);
```

Which will v8 serialize `buffer`, we should be handling it gracefully.

With the large blobs, we might add a LMDB entry which points at the file for example.

PR #563, opened 2025-05-29
AFB-1244 - add operational only statsig flags to atlaspack for AFM statsig client validation

matt-koko:
Add two temporary flags to atlaspack for ATM statsig client validation.

## Motivation

To facilitate validation of new statsig client in AFM integration with Atlaspack flags.
See issue AFB-1244 for more info.

## Changes

Added two operational only flags.
- kokoTestGateOne
- kokoTestGateTwo

## Checklist

- [x] Existing or new tests cover this change - n/a

[no-changeset]: temporary flags


pancaspe87:
Note for future PRs; IIRC we usually don't reference ticket numbers in open source projects. As general rule, we try to avoid referencing internal info.

PR #572, opened 2025-06-01
Do not store environment field on entities

yamadapc:
Deduplicates environment field and only keep a string around.

Store all environments on a global 'RwLock<...>' hashmap (this should be the
LMDB cache instead).


yamadapc:
WIP because:

* no feature-flag
* no storage / loading from cache

packages/core/rust/test/EnvironmentManager.test.js
```diff
@@ -0,0 +1,67 @@
...
+    const storedEnvironment = getEnvironment(environment1.id);
+    assert.deepEqual(storedEnvironment, environment1);
+
+    const storedEnvironment2 = getEnvironment(environment2.id);
+    assert.deepEqual(storedEnvironment2, environment2);
+
+    const environments = getAllEnvironments();
+    assert.deepEqual(environments, [environment1, environment2]);
+
+    setAllEnvironments([]);
```
pancaspe87, line 56: Is there any scenario where we would need to empty all the existing environments?

pancaspe87:
Re: https://github.com/atlassian-labs/atlaspack/pull/572#issuecomment-2928126340

I was actually wondering if these changes need to be feature flagged before landing this PR

packages/core/core/src/EnvironmentManager.js
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * When deduplication is cleaned-up this will always be a string.
+ */
+export type EnvironmentRef = EnvironmentId | CoreEnvironment;
+
+/**
+ * Convert environment to a ref.
+ * This is what we should be using to store environments.
+ */
+export function toEnvironmentRef(env: CoreEnvironment): EnvironmentRef {
```
yamadapc, line 22: Feature-flags are here

When the feature-flag is off, we will continue to pass objects around.

When the feature-flag is on, we switch objects by an opaque string.

packages/core/core/src/AssetGraph.js
```diff
@@ -147,19 +148,6 @@ export default class AssetGraph extends ContentGraph<AssetGraphNode> {
     };
   }
 
-  // Deduplicates Environments by making them referentially equal
-  normalizeEnvironment(input: Asset | Dependency | AssetGroup) {
```
yamadapc, line 151: This shouldn't be removed without a feature-flag

PR #578, opened 2025-06-02
Do not run CI on node 18

yamadapc:
Remove node 18 on test suites and CI jobs.

[no-changeset] CI changes

Test Plan: CI


dddlr (approved):
looks good 👍

can we add node v22 to integration_tests and integration_tests_v3 in a follow-up PR? (as @mattcompiles mentioned on slack)

and curious re: the reason for not adding v22 for macos?

pancaspe87 (approved):

PR #584, opened 2025-06-03
Pr 572/store env to cache

pancaspe87:
WIP PR, based on https://github.com/atlassian-labs/atlaspack/pull/572
Open to gather initial feedbacks from @yamadapc 

---
- We want to list all environments using `getAllEnvironments` then store them into the cache inside of `RequestTracker::writeToCache`
- Environments should be stored as: `Environment/$ATLASPACK_VERSION/$ENVIRONMENT_ID`
- A separate key should be stored: `EnvironmentManager/$ATLASPACK_VERSION`. This should contain the list of environment IDs used in the last build.
- When loading the request tracker in `loadRequestGraph`, we should load the environments required for the previous build.


packages/core/core/src/RequestTracker.js
```diff
@@ -1499,130 +1444,177 @@ export default class RequestTracker {
...
+      });
 
-    const serialiseAndSet = async (
-      key: string,
-      // $FlowFixMe serialise input is any type
-      contents: any,
-    ): Promise<void> => {
-      if (signal?.aborted) {
-        throw new Error('Serialization was aborted');
+      const environments = getAllEnvironments();
```
pancaspe87, line 1474: Write list of environments to cache

packages/core/core/src/RequestTracker.js
```diff
@@ -1774,7 +1791,52 @@ async function loadRequestGraph(options): Async<RequestGraph> {
     },
   });
 
-  if (await options.cache.hasLargeBlob(requestGraphKey)) {
+  if (getFeatureFlag('cachePerformanceImprovements')) {
```
pancaspe87, line 1794: Loading the  environments required for the previous build.

packages/core/core/src/RequestTracker.js
yamadapc: could you extract both read / write to functions at the module level?

pancaspe87:
Closing as duplicated of https://github.com/atlassian-labs/atlaspack/pull/599, which was opened to better highlight the diff with the SPR it is based off

PR #588, opened 2025-06-04
Support granular tsconfig invalidations

yamadapc:
At the moment if tsconfig files change many javascript transformation outputs
are invalidated.

This commit patches the js transformer that so that transformations are only
invalidated if the 'compilerOptions' key changes.

Test Plan:
See added integration tests replicating current and desired behavior.


mattcompiles (approved):
Nice to see `invalidateOnConfigKeyChange` getting some more use 😄

PR #593, opened 2025-06-05
Improve arrow function support

mgfalzon:
## Motivation

This change decreases the JS output for Confluence by 366kb. For large applications the bytes add up quick!

Other smaller optimizations to consider:
* Replacing other helper function calls with arrow function
* Changing `parcelRequire.register` to `parcelRequire.r`
* Reducing the size of `parcelRequireName`, the current definition for confluence is `parcelRequired477` which shows up in 3721 files in Conflunece. Reducing it to `rd477` would save 44kb.

## Changes

* Use arrow functions in parcelRegister calls
* Reduce calls to `this.bundle.env.supports('arrow-functions', true);`

## Checklist

### DIFF
--- a/.changeset/stupid-kiwis-change.md
+++ b/.changeset/stupid-kiwis-change.md
@@ -1,0 +1,5 @@
+---
+'@atlaspack/packager-js': minor
+---
+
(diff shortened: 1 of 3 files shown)


yamadapc (commented):
LGTM but you should feature-flag it

PR #594, opened 2025-06-05
Fix LMDB database running out of reader slots

yamadapc:
It is possible on certain use cases that the LMDB database will run out of
reader slots if processes are quit while keeping the database open.

This adds a call which will check and clean stale readers, fixing a database
that has become unable to be opened.

Test Plan: N/A

### DIFF
--- a/crates/lmdb-js-lite/src/writer.rs
+++ b/crates/lmdb-js-lite/src/writer.rs
@@ -375,6 +375,11 @@ impl DatabaseWriter {
       env
     }?;
 
+    // If processes are terminated, they might leave stale readers on the DB
+    // the database will be unusable after it reaches a certain number of
+    // readers, so clean-up is useful on start-up.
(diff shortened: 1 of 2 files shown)


mattcompiles (approved):

crates/lmdb-js-lite/src/writer.rs
```diff
@@ -375,6 +375,11 @@ impl DatabaseWriter {
       env
     }?;
 
+    // If processes are terminated, they might leave stale readers on the DB
+    // the database will be unusable after it reaches a certain number of
+    // readers, so clean-up is useful on start-up.
+    environment.clear_stale_readers()?;
```
pancaspe87, line 381: Do we need to feature flag this change?

yamadapc: No this should be fine

PR #599, opened 2025-06-06
Write and load deduplicated environments from cache

pancaspe87:
PR, based on https://github.com/atlassian-labs/atlaspack/pull/572

---
- We want to list all environments using `getAllEnvironments` then store them into the cache inside of `RequestTracker::writeToCache`
- Environments should be stored as: `Environment/$ATLASPACK_VERSION/$ENVIRONMENT_ID`
- A separate key should be stored: `EnvironmentManager/$ATLASPACK_VERSION`. This should contain the list of environment IDs used in the last build.
- When loading the request tracker in `loadRequestGraph`, we should load the environments required for the previous build.

### DIFF
--- a/packages/core/core/src/RequestTracker.js
+++ b/packages/core/core/src/RequestTracker.js
@@ -71,6 +71,7 @@ import type {
   RequestInvalidation,
   InternalFileCreateInvalidation,
   InternalGlob,
+  Environment,
 } from './types';
 import {BuildAbortError, assertSignalNotAborted, hashFromOption} from './utils';
 import {performance} from 'perf_hooks';
@@ -2263,7 +2264,7 @@ export function getBiggestFSEventsInvalidations(
  * @param {LMDBLiteCache} cache
  * @param {Environment} env
  */
-export async function storeEnvById(cache, env) {
+export async function storeEnvById(cache: Cache, env: Environment) {
   const envKey = `Environment/${ATLASPACK_VERSION}/${env.id}`;
 
   if (await cache.get(envKey)) {
@@ -2283,7 +2284,10 @@ export async function storeEnvById(cache, env) {
  * @param {LMDBLiteCache} cache
  * @param {Set<string>} environmentIds
  */
-export async function storeEnvManager(cache, environmentIds) {
+export async function storeEnvManager(
+  cache: Cache,
+  environmentIds: Iterable<string>,
+) {
   await instrument(
     `RequestTracker::writeToCache::cache.put(${`EnvironmentManager/${ATLASPACK_VERSION}`})`,
     async () => {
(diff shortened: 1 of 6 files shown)


packages/core/core/test/RequestTracker.test.js
```diff
@@ -590,3 +600,127 @@ root --- node1 --- node2 ----------- orphan1 --- orphan2
...
+    await options.cache.ensure();
+
+    for (const key of options.cache.keys()) {
+      await options.cache.getNativeRef().delete(key);
+    }
+    setAllEnvironments([]);
+
+    setFeatureFlags({
+      ...DEFAULT_FEATURE_FLAGS,
+      cachePerformanceImprovements: true,
```
pancaspe87, line 647: This could be redundant since I then realised we do `cachePerformanceImprovements: process.env.NODE_ENV === 'test' ? true : false` in [here](https://github.com/atlassian-labs/atlaspack/blob/main/packages/core/feature-flags/src/index.js#L23), but keeping it to be more explicit of the intent of the tests

pancaspe87:
[] replace flag and use `environmentDeduplication`

packages/core/core/test/RequestTracker.test.js
pancaspe87: - [ ] replace with `environmentDeduplication`

packages/core/core/src/RequestTracker.js
```diff
@@ -2225,3 +2258,89 @@ export function getBiggestFSEventsInvalidations(
...
+
+/**
+ * Stores an environment object by its ID
+ * @param {LMDBLiteCache} cache
+ * @param {Environment} env
+ */
+export async function storeEnvById(cache: Cache, env: Environment) {
+  const envKey = `Environment/${ATLASPACK_VERSION}/${env.id}`;
+
+  if (await cache.get(envKey)) {
```
yamadapc, line 2270: There's no need to do this ; just call cache.set

packages/core/core/src/RequestTracker.js
```diff
@@ -1856,6 +1867,28 @@ async function loadRequestGraph(options): Async<RequestGraph> {
...
+    try {
+      await loadEnvironmentsFromCache(options.cache);
+      logger.verbose({
+        origin: '@atlaspack/core',
+        message: 'Environments were loaded from cache',
+        meta: {
+          ...commonMeta,
+        },
+      });
+    } catch (err) {
```
yamadapc, line 1880: This is a hard failure so it's better not to handle the error

packages/core/core/src/RequestTracker.js
```diff
@@ -1572,6 +1577,12 @@ export default class RequestTracker {
         size: this.graph.nodes.length,
       });
 
+      if (getFeatureFlag('environmentDeduplication')) {
+        async (cache) => {
```
yamadapc, line 1581: I'm not sure if I messed it up on rebase but it should just be called

yamadapc (approved):

PR #601, opened 2025-06-06
Improve granular config invalidation

yamadapc:
Modifies previous changes to allow for granular tsconfig invalidations to also
support package.json reads on the JSTransformer.

Adds something akin to the environment and options read tracking. Uses a proxy
object to perform read tracking, then registers granular invalidations based on
which fields are read.

I had to change integration tests I had added previously. The reason being
any create above invalidation will get hit when a file create event is received
and sometimes due behavior of `NodeFS.writeFile` (which uses file move) causes
create events to be fired for the writes.

### DIFF
--- a/packages/core/core/test/requests/ConfigRequest.test.js
+++ b/packages/core/core/test/requests/ConfigRequest.test.js
@@ -12,7 +12,10 @@ import type {
   ConfigRequestResult,
 } from '../../src/requests/ConfigRequest';
 import type {RunAPI} from '../../src/RequestTracker';
-import {getObjectKey, runConfigRequest} from '../../src/requests/ConfigRequest';
+import {
+  getValueAtPath,
+  runConfigRequest,
+} from '../../src/requests/ConfigRequest';
 import {toProjectPath} from '../../src/projectPath';
 
 // $FlowFixMe unclear-type forgive me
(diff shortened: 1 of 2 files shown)


packages/core/core/test/requests/ConfigRequest.test.js
```diff
@@ -249,3 +249,226 @@ describe('ConfigRequest tests', () => {
     );
   });
 });
+
+describe('getObjectKey', () => {
```
yamadapc, line 253: Mostly written by AI

packages/core/core/src/public/Config.js
```diff
@@ -27,6 +27,74 @@ const internalConfigToConfig: DefaultWeakMap<
...
+ *        usedPaths.add(path);
+ *     };
+ *
+ *     const config = makeConfigProxy(onRead, {a: {b: {c: 'd'}}})
+ *     console.log(config.a.b.c);
+ *     console.log(Array.from(usedPaths)); // ['a', 'b', 'c']
+ *
+ * In case the value is null or an array, we will track the read as well.
+ */
+export function makeConfigProxy<T>(
```
yamadapc, line 54: Whenever reading a JSON configuration file, we can track properties read.

There are a few details of how changes will be handled, which are tested in the unit-tests.

packages/core/types-internal/src/index.js
```diff
@@ -978,17 +978,18 @@ export interface Config {
...
           exclude?: boolean,
         |}
       | {|
           /**
-           * If specified, only invalidate when this config key changes.
+           * If specified, this function will return a proxy object that will track reads to
+           * config fields and only register invalidations for when those keys change.
            */
-          configKey?: string,
+          readTracking?: boolean,
```
pancaspe87, line 992: Is `configKey` used in any product?

yamadapc: I had added it under this flag last week

packages/core/core/test/requests/ConfigRequest.test.js
```diff
@@ -244,8 +244,180 @@ describe('ConfigRequest tests', () => {
...
       call.args,
-      ['config.json', 'key1', hashString('"value1"')],
+      ['config.json', ['key1'], hashString('"value1"')],
       'Invalidate was called for key1',
     );
   });
 });
+
+describe('getObjectKey', () => {
+  it('can get a key from an object', () => {
```
pancaspe87, line 254: nit
should the function be named something like `getObjectByKey/getObjectValueByKey/etc` OR should the test case renamed to something like `retrieves a nested value from an object when given a path array'

pancaspe87 (approved):

PR #612, opened 2025-06-12
Scope hoisting within wrapped assets

mgfalzon:
## Motivation

To add scope-hoisting within wrapped assets.

## Changes
* Pass `depContent` from the toplevel instead of within the recursive function. This prevents `parcelRequire` wrapped within another `parcelRequire` when performing wrappedAsset inlining.
* Add `addInlineAssetMetadata` which adds the `inline` metadata to tell `ScopeHoistingPackager.js` which assets should be inlined.
* Add a unit test to test for wrapped asset inlining
* Attempt to inline assets within `ScopeHoistingPackager.js` but there are issues

**This implementation would be MUCH easier if we could just modify the assetGraph after it's done building in all the transformers. If we did that we wouldn't need to change `ScopeHoistingPackager.js` or have logic in the bundler to add metadata.** I looked into this, but I couldn't figure out a way to do it. We'd likely have to add an additional plugin type.

## Checklist


packages/bundlers/default/src/addInlineAssetMetadata.js
```diff
@@ -0,0 +1,92 @@
...
+     * the incoming dependency counts for their resolved assets.
+     */
+    for (const dependency of assetToScopeHoistDependencies) {
+      const asset = assetGraph.getResolvedAsset(dependency);
+      if (asset && assetToScopeHoistIntoDependencyAssets.has(asset)) {
+        const previousIncomingDependencies = incomingDependencyMap.get(asset);
+        if (previousIncomingDependencies !== undefined) {
+          const incomingDependencies = previousIncomingDependencies.filter(
+            (dep) => {
+              return !assetToScopeHoistDependencies.has(dep);
```
mgfalzon, line 83: This check is incorrect. I think we just want to check the current dependency/asset, not filter them all out

PR #613, opened 2025-06-13
Fix constant modules not being included in MSBs

marcins:
## Motivation

When enabling inline constants for one of our internal products, we had errors where "constant modules" weren't being included in MSBs if they didn't match the MSB regex which would cause an error during packaging.

## Changes

A bug was fixed in `makeManualAssetToConfigLookup` to ensure constant modules were added to the mapping _before_ the regex check and early bailout.

## Checklist

- [x] Existing or new tests cover this change (new integration test)

### DIFF
--- a/.changeset/dirty-nails-double.md
+++ b/.changeset/dirty-nails-double.md
@@ -1,0 +1,6 @@
+---
+'@atlaspack/integration-tests': patch
+'@atlaspack/bundler-default': patch
+---
+
+Ensure that constant modules are correctly included in MSBs even if they wouldn't otherwise be.
--- a/packages/bundlers/default/src/idealGraph.js
+++ b/packages/bundlers/default/src/idealGraph.js
@@ -204,14 +204,15 @@ export function createIdealGraph(
             config.projectRoot,
             node.value.filePath,
           );
-          if (!assetRegexes.some((regex) => regex.test(projectRelativePath))) {
-            return;
-          }
 
           // We track all matching MSB's for constant modules as they are never duplicated
           // and need to be assigned to all matching bundles
           if (node.value.meta.isConstantModule === true) {
             constantModuleToMSB.get(node.value).push(c);
+          }
+
+          if (!assetRegexes.some((regex) => regex.test(projectRelativePath))) {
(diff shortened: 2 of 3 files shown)


packages/bundlers/default/src/idealGraph.js
```diff
@@ -204,15 +204,16 @@ export function createIdealGraph(
...
-            return;
-          }
 
           // We track all matching MSB's for constant modules as they are never duplicated
           // and need to be assigned to all matching bundles
           if (node.value.meta.isConstantModule === true) {
             constantModuleToMSB.get(node.value).push(c);
           }
+
+          if (!assetRegexes.some((regex) => regex.test(projectRelativePath))) {
```
marcins, line 214: I guess I could make this neater now: 
```js
 if (assetRegexes.some((regex) => regex.test(projectRelativePath))) {
     manualAssetToConfig.set(node.value, c);
}
return;

File: .changeset/dirty-nails-double.md [hunk 2]
```diff
@@ -0,0 +1,6 @@
+---
+'@atlaspack/integration-tests': patch
```
mattcompiles, line 2: Don't really need to patch private stuff btw

packages/bundlers/default/src/idealGraph.js
```diff
@@ -204,17 +204,16 @@ export function createIdealGraph(
...
 
           // We track all matching MSB's for constant modules as they are never duplicated
           // and need to be assigned to all matching bundles
           if (node.value.meta.isConstantModule === true) {
             constantModuleToMSB.get(node.value).push(c);
           }
-          manualAssetToConfig.set(node.value, c);
-          return;
+
+          if (assetRegexes.some((regex) => regex.test(projectRelativePath))) {
```
mattcompiles, line 214: Wow this was a fail lol

.changeset/dirty-nails-double.md
marcins: Should I remove this or is it fine to stay for this one?

mattcompiles (approved):

PR #615, opened 2025-06-13
Use direct reference to private babel-register package in APVM Atlaspack script

marcins:
## Motivation

The `ATLASPACK_DEV` branch of the `.bin/atlaspack` script didn't work because `@atlaspack/babel-register` is a private package, so the `reqiure('@atlaspack/register')` would fail.

## Changes

 Change the `require` call to instead directly reference the path, using `APVM_ATLASPACK_LOCAL`. Seems to work!

## Checklist

### DIFF
--- a/.changeset/large-moons-sleep.md
+++ b/.changeset/large-moons-sleep.md
@@ -1,0 +1,9 @@
+---
+'@atlaspack/apvm': patch
+'@atlaspack/apvm-linux-amd64': patch
+'@atlaspack/apvm-linux-arm64': patch
+'@atlaspack/apvm-macos-amd64': patch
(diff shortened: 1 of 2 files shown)


.changeset/large-moons-sleep.md
```diff
@@ -0,0 +1,9 @@
+---
+'@atlaspack/apvm': patch
```
marcins, line 2: @alshdavid Wasn't any changes detected since change is in Rust code - not sure if this is the right set of pacakges?

alshdavid: Yeah, this is correct. It's just a Rust binary distributed via npm. The main package and the os-specific packages will update in sync

PR #627, opened 2025-06-19
Add scope hoisting of assets inside wrapped assets

benjervis:
## Motivation

We did some testing of the scope hoisting packager, and it turns out that for one of our products, only 0.6% of assets are actually hoisted, with the rest being wrapped in `parcelRegister`.

This is because anything that's dynamically imported needs to be wrapped, and all children of wrapped assets also get wrapped. It seems like the original intention for this was to handle the case of a circular dependency, but it's a pretty big blast radius.

## Changes

This change adds a new feature flag `applyScopeHoistingImprovement` which allows children of wrapped assets to be inlined into the wrapped asset. This is achieved by assigning assets to their wrapped parents and if multiple parents are detected then they becomes wrapped themselves. 

We've also had to exclude assets that contain symbols that are re-exported. Ideally this wouldn't be the case, but we need to solve more complicated cases around symbol resolution to make this work. 

## Checklist

[webpack does]: https://webpack.js.org/plugins/module-concatenation-plugin/#optimization-bailouts

### DIFF
--- a/packages/core/integration-tests/test/scope-hoisting.js
+++ b/packages/core/integration-tests/test/scope-hoisting.js
@@ -6731,5 +6731,72 @@ describe('scope hoisting', function () {
 
       assert.deepEqual(result, ['leaf-one', 'fourleaf-one']);
     });
+
+    it('should handle * re-exports in wrapped groups', async function () {
+      await fsFixture(overlayFS, __dirname)`
+        star-reexports-in-wrapped-groups
+          index.html:
+            <script type="module" src="./index.js"></script>
+
+          index.js:
+            import {one} from './first-barrel';
+            import {four} from './second-barrel.js';
+            result([one, four()]);
+
+          first-barrel.js:
+            export * from './second-barrel.js';
+
+          second-barrel.js:
+            import {valueOne as one} from './leaf-one';
+            export {one};
+            // Force this file to exclusive re-export
+            export function four() { return 'four' + one; }
+
+          leaf-one.js:
+            export const valueOne = 'leaf-one';
+
+          package.json:
+            {
+              "name": "scope-hosting-test",
+              "sideEffects": ["index.js"],
+              "@atlaspack/bundler-default": {
+                "manualSharedBundles": [{
+                  "name": "shared",
+                  "assets": ["!index.js"]
+                }]
+              }
+            }
+          yarn.lock:
+      `;
+
+      let b = await bundle(
+        path.join(__dirname, 'star-reexports-in-wrapped-groups/index.html'),
+        {
+          inputFS: overlayFS,
+          featureFlags: {
+            applyScopeHoistingImprovement: true,
+          },
+        },
+      );
+
+      let sharedBundle = nullthrows(
+        b.getBundles().find((b) => b.manualSharedBundle === 'shared'),
+      );
+
+      assertWrappedAssets(b, sharedBundle, [
+        'first-barrel.js',
+        'second-barrel.js',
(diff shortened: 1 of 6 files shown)


packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -418,6 +431,16 @@ export class ScopeHoistingPackager {
...
         }
+
+        if (getFeatureFlag('applyScopeHoistingImprovement')) {
+          let outgoingDeps = this.bundleGraph.getDependencies(asset).length;
+          if (outgoingDeps === 0) {
+            // If an asset has no dependencies, then there are no deps to worry
+            // about loading first, so it's safe to hoist
+            return;
+          }
+        }
```
mgfalzon, line 442: @benjervis Do we want to do this? Won't this lead to lots of duplicated code? Shouldn't we only do this if incomingDependencies is also 1?

packages/core/integration-tests/test/scope-hoisting.js
```diff
@@ -6491,4 +6492,311 @@ describe('scope hoisting', function () {
...
+      await run(b, {
+        result(o) {
+          result = o;
+        },
+      });
+
+      assert.deepEqual(result, {valueOne: 'leaf-one', valueTwo: 'leaf-two'});
+    });
+
+    it.only('should handle complex re-exports', async function () {
```
marcins, line 6744: I guess the `.only` here needs to be removed...

marcins (approved):

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -387,48 +388,97 @@ export class ScopeHoistingPackager {
...
+      });
+
+    let assetSymbols = this.bundleGraph.getExportedSymbols(asset, this.bundle);
+
+    return assetSymbols.some((assetSymbol) =>
+      parentSymbols.some(
+        (parentSymbol) => parentSymbol.symbol === assetSymbol.symbol,
+      ),
+    );
+  }
```
mgfalzon, line 480: Does this check account for namespace re-exports?
```js
export * from 'my-module'
```
should we add a test for that?

mgfalzon (approved):

packages/packagers/js/src/ScopeHoistingPackager.js
mattcompiles: Added one 👍

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -387,48 +388,97 @@ export class ScopeHoistingPackager {
...
-          actions.skipChildren();
-          return;
-        }
-        if (!asset.meta.isConstantModule) {
-          this.wrappedAssets.add(asset.id);
-          wrapped.push(asset);
-        }
-      }, wrappedAssetRoot);
+      for (let wrappedAsset of wrapped) {
+        this.bundle.traverseAssets((asset, _, actions) => {
```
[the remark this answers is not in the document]
mattcompiles: Yes that's roughly right. 
> Assets end up in assignedAssets if they are used more than once in the bundle
Used in more than one wrapped asset is more correct. An asset can be used twice inside a wrapped asset (we often refer to them as wrap groups) and still get inlined.
> if they are used more than once in the bundle, or we end up back at at asset via a circular dependency.
This algo doesn't really catch circular deps as they don't seem to cause issues in this case.

PR #628, opened 2025-06-19
Add technical docs for conditional imports

dddlr:
Add technical docs for conditional imports. This is designed to give a technical overview of conditional imports, which complements our existing conditional imports docs which are more focused on adoption.

## Motivation

We do not currently have technical docs on conditional imports, so it would be good to add some.

## Changes

Added as Markdown file for now, we can decide on a long-term home later.

## Checklist

[no-changeset]: Changes documentation only


docs/features/Conditional Bundling.md
```diff
@@ -0,0 +1,260 @@
...
+
+Recall that `importCond` is converted to this in the build output:
+
+```js
+const MyComponent = globalThis.__MCOND('my-feature-gate-name')
+  ? require('./true-component.tsx').default
+  : require('./false-component.tsx').default;
+```
+
+> **Sidenote**
```
dddlr, line 234: is `server: true` still used anywhere? and if so, is my explanation actually correct?

@JakeLane

docs/features/Conditional Bundling.md
```diff
@@ -0,0 +1,260 @@
...
+If:
+
+- `TrueComponent` and `FalseComponent` are quite large, or
+- `TrueComponent` has dependencies (e.g. it imports code from other places) that `FalseComponent` doesn't have (or vice versa)
+
+then the user will end up downloading code that they never need to execute. For example, if `fg('my-feature-gate-name')` is true, then they will never need `FalseComponent`, but the user is still forced to download it.
+
+### The solution
+
+Atlaspack's conditional imports API solve this problem by providing a way to notify the Atlaspack bundler that there is conditional code. At build time, Atlaspack will then ensure that `TrueComponent` and `FalseComponent` are separate bundles in the build output, and it will generate a **conditional manifest** which is a JSON file that details which bundle(s) should be loaded for each value of `fg('my-feature-gate-name')`. The web server (e.g. Bifrost) would then check the conditional manifest to determine which bundles to serve to any given user, depending on the value of `fg('my-feature-gate-name')`.
```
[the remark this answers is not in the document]
JakeLane: > The web server (e.g. Bifrost)
probably good to add that the runtime will do it when loading an async chunk as well

docs/features/Conditional Bundling.md
```diff
@@ -0,0 +1,260 @@
...
+> **Type safety:** If you have two files, you'd have to consider the types of each export from the files. That would cause issues if one file exported things the other didn't. If we only consider the default export, that means you only have to match one export and handle one export in the callsite.
+>
+> **Bundling complexity:** Simplifying to default export means we can simplify the runtime code. If we don't, that means we have to handle a bunch of bundling edge cases to make sure we get optimised code (e.g. dead code removal, side-effects etc.)
+
+At runtime, the `importCond` function call will be converted to something like this:
+
+```js
+// index.js
+const MyComponent = globalThis.__MCOND('my-feature-gate-name')
+  ? require('./true-component.tsx').default
```
JakeLane, line 78: This is misleading because `require` is a CJS import, which is not the case here.

Maybe replace this with:
```
register("xxx", function(e, t) {
    e.exports = parcelRequire("avM1e")("my-feature-gate-name", function() {
        return parcelRequire("1feRM")
    }, function() {
        return parcelRequire("dqBV0")
    })
}),
// ...
const MyComponent = register("xxx").default;
```

docs/features/Conditional Bundling.md
JakeLane: `node: true` is used in SSR rendering

docs/features/Conditional Bundling.md
```diff
@@ -0,0 +1,260 @@
...
+  </head>
+  <body>
+    <script src="shared.xxx.js"></script>
+    <script src="some-new-bundle.xxx.js"></script>
+    <script src="jira-spa-issue-view.xxx.js"></script>
+  </body>
+</html>
+```
+
+With this kind of setup, we can easily update the web server to parse the conditional manifest and ensure that the user loads the correct bundles:
```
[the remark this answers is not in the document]
JakeLane: I think it would be good to do this, but I was worried about the risk that the behaviour may need to be different between Jira-Fe-Gateway and Bifrost (and updating atlaspack is slow). I think in practice it's probably fixable with the second param like above

JakeLane (approved):

dddlr:
thanks all for the feedback and suggestions 🙇 will merge this in PR now just _some_ docs merged in before general availability, but will make a follow-up PR asap to address all of the feedback properly

docs/features/Conditional Bundling.md
dddlr: but that's separate to `server: true` right?

PR #631, opened 2025-06-19
Lock package versions

yamadapc:
Test Plan: N/A


pancaspe87 (approved):

.changeset/config.json
```diff
@@ -11,7 +11,147 @@
       "@atlaspack/apvm-linux-amd64",
       "@atlaspack/apvm-linux-arm64",
       "@atlaspack/apvm-macos-amd64",
-      "@atlaspack/apvm-macos-arm64"
+      "@atlaspack/apvm-macos-arm64",
```
pancaspe87, line 14: Did you just list all the available packages?

matt-koko:
Is there a way we can use a glob expression to match on all the packages in Atlaspack?

[https://github.com/changesets/changesets/blob/main/docs/fixed-packages.md#using-glob-expressions](https://github.com/changesets/changesets/blob/main/docs/fixed-packages.md#using-glob-expressions)

If a new package is added in the future, will the developer know to include the new package in the fixed changeset array?

.changeset/config.json
```diff
@@ -5,155 +5,7 @@
...
-      "@atlaspack/types-internal",
-      "@atlaspack/typescript-example",
-      "@atlaspack/utils",
-      "@atlaspack/validator-eslint",
-      "@atlaspack/validator-typescript",
-      "@atlaspack/watcher-watchman-js",
-      "@atlaspack/workers"
-    ]
-  ],
+  "fixed": [["@atlaspack/*"]],
```
matt-koko, line 8: nice 🙌

yamadapc:
We will instead lock package versions and re-export core packages from `@atlaspack/core`

PR #632, opened 2025-06-20
Optimise inline consts and namespaces

marcins:
## Motivation

With constant inlining, in some cases constant modules have namespace objects prepended by the packager which reduces their efficiency. For example, we might end up with code that looks like this:

```js
var s={};
$parcel$export(s,"CONSTANT",()=>a);
$parcel$export(s,"OTHER",()=>u)
const a=42,u="other";
e.result=`The constant is ${a}`;
```

Where we would expect to see:
```js
e.result=`The constant is 42`;
```

In some cases, particularly where code does an `import * as consts from './constants` this namespacing _is_ required. However with the current logic, the constant module will have the namespace object added regardless of whether that namespace import is in the current bundle or a different bundle. As const modules are effectively duplicated but not wrapped, this doesn't make sense to do - it would make sense if the module _was_ wrapped because then you'd expect every wrapped duplicate to be identical.

## Changes

This change makes it so that `usedNamespace` is `true` for constant modules only if the `*` import is in the current bundle. A new integration test has been added to verify this.

- [] One thing I'm not sure about is the other bits of logic that contribute to `usedNamespace` and whether they should also apply? I need to double check this and see if anything breaks in real-world usage.

## Checklist


packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -1220,34 +1220,49 @@ ${code}
...
+      // Only set usedNamespace if there is an incoming dependency in the current bundle that uses '*'
+      usedNamespace = this.bundleGraph
+        .getIncomingDependencies(asset)
+        .some(
+          (dep) =>
+            this.bundle.hasDependency(dep) &&
+            nullthrows(this.bundleGraph.getUsedSymbols(dep)).has('*'),
+        );
+    } else {
+      usedNamespace =
```
marcins, line 1237: This should be the existing logic, unchanged - the logic above only applies to constant modules.

JakeLane (approved):

packages/core/feature-flags/src/index.js
```diff
@@ -38,6 +38,7 @@ export const DEFAULT_FEATURE_FLAGS: FeatureFlags = {
   resolveBundlerConfigFromCwd: false,
   conditionalBundlingReporterSameConditionFix: false,
   condbHtmlPackagerChange: false,
+  inlineConstOptimisationFix: false,
```
marcins, line 41: I've used one feature flag for two related fixes.. not sure if I should split this into two flags?

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -717,7 +728,17 @@ export class ScopeHoistingPackager {
       sourceMap?.offsetLines(1, 1);
       lineCount++;
 
-      code = `parcelRegister(${JSON.stringify(
+      let earlyCode = '';
```
marcins, line 731: This isn't gated, but the only thing that populates `earlyDepContent` is gated.. I think that's safe enough?

packages/packagers/js/src/ScopeHoistingPackager.js
marcins: I originally had this condition in the uber-condition below but saw some issues - it's possible I had just gotten something wrong.. could any of the other cases below apply to constant modules?

packages/core/feature-flags/src/index.js
mattcompiles: I think it makes sense to have a single flag here

mattcompiles (approved):
This looks good to me mate. Probably a good idea to validate it in Jira too?

PR #633, opened 2025-06-20
Port HMR changes from parcel

sbhuiyan-atlassian:
## Motivation

This PR pulls in some recent HMR changes from Parcel.

## Changes

See the individual commit messages for the list of changes.

## Checklist


JakeLane:
We also will need a changeset (just run `yarn changeset`) so that we get everything released correctly

sbhuiyan-atlassian:
> We also will need a changeset (just run `yarn changeset`) so that we get everything released correctly

Done

JakeLane (approved):

packages/core/core/src/AssetGraph.js
```diff
@@ -538,6 +538,14 @@ export default class AssetGraph extends ContentGraph<AssetGraphNode> {
           ...depNode.value.meta,
           ...existing.value.resolverMeta,
         };
+        depNode.value.resolverMeta = existing.value.resolverMeta;
```
marcins, line 541: Should changes like this also be gated behind the ff?

packages/core/core/src/requests/PathRequest.js
```diff
@@ -319,7 +319,8 @@ export class ResolverRunner {
           }
 
           if (result.priority != null) {
-            dependency.priority = Priority[result.priority];
+            dependency.priority = dependency.resolverPriority =
```
marcins, line 322: ff here too?

packages/core/integration-tests/test/integration/hmr-duplicate/a.js
marcins, line 1: Should use `fsFixture` for these tests to avoid the need for adding lots of fixture files to the repo.

packages/packagers/js/src/DevPackager.js
```diff
@@ -112,6 +112,16 @@ export class DevPackager {
           }
         }
 
+        // Add dependencies for parcelRequire calls added by runtimes
+        // so that the HMR runtime can correctly traverse parents.
+        let hmrDeps = asset.meta.hmrDeps;
```
marcins, line 117: ff this too.

marcins (changes requested):
There's a bunch of places where code changes have been made without wrapping them in a feature flag check - for safety we should ensure as much of the change as practical is gated so that we can easily respond to any issues that might occur when rolling these changes out.

packages/transformers/js/core/src/dependency_collector.rs
```diff
@@ -522,6 +525,39 @@ impl VisitMut for DependencyCollector<'_> {
 
                 return;
               }
+              "parcelRequire" => {
+                if self.config.hmr_improvements {
```
yamadapc, line 529: Could you write tests for this change?

sbhuiyan-atlassian: Done. Added tests for the FF on and off cases.

packages/core/core/src/requests/PathRequest.js
sbhuiyan-atlassian: Added FF check

packages/packagers/js/src/DevPackager.js
sbhuiyan-atlassian: Added FF check

packages/core/integration-tests/test/integration/hmr-duplicate/a.js
sbhuiyan-atlassian: As discussed, we'll skip this one

marcins (approved):
Appreciate your effort porting this over and addressing our feedback! 👍

JakeLane (approved):

PR #634, opened 2025-06-20
Update conditional bundling docs

dddlr:
Update technical docs for conditional imports based on PR suggestions and feedback. This is a follow-up to #628, see there for more context.

## Checklist

[no-changeset]: docs change

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,8 @@ You should also add documentation to `docs/` if you are adding new features or a
 
 For bigger changes, please make sure you start a discussion first by creating an issue and explaining the intended change. -->
 
+> **If you're an Atlassian employee:** you shouldn't sign the CLA. Instead, please search for the "Link your Github & Atlassian accounts" page on Confluence, and follow the instructions there to link your GitHub account to the Atlassian system.
+
 Atlassian requires contributors to sign a Contributor License Agreement, known as a CLA. This serves as a record stating that the contributor is entitled to contribute the code / documentation / translation to the project and is willing to have it used in distributions and derivative works (or is willing to transfer ownership).
 
 Prior to accepting your contributions we ask that you please follow the appropriate link below to digitally sign the CLA. The Corporate CLA is for those who are contributing as a member of an organization and the individual CLA is for those contributing as an individual.
--- a/docs/features/Conditional Bundling.md
+++ b/docs/features/Conditional Bundling.md
@@ -8,7 +8,7 @@ If you wish to use conditional imports in Jira, you should use our adoption guid
 
 ### The problem
 
-Take the following example code:
+Take the following example code that uses synchronous imports:
 
 ```tsx
 // my-component.js
@@ -34,7 +34,9 @@ then the user will end up downloading code that they never need to execute. For 
 
 ### The solution
 
-Atlaspack's conditional imports API solve this problem by providing a way to notify the Atlaspack bundler that there is conditional code. At build time, Atlaspack will then ensure that `TrueComponent` and `FalseComponent` are separate bundles in the build output, and it will generate a **conditional manifest** which is a JSON file that details which bundle(s) should be loaded for each value of `fg('my-feature-gate-name')`. The web server (e.g. Bifrost) would then check the conditional manifest to determine which bundles to serve to any given user, depending on the value of `fg('my-feature-gate-name')`.
+Atlaspack's conditional imports API solve this problem by providing a way to hint to the Atlaspack bundler that there is conditional code. At build time, Atlaspack will then ensure that `TrueComponent` and `FalseComponent` are separate bundles in the build output, and it will generate a **conditional manifest** which is a JSON file that details which bundle(s) should be loaded for each value of `fg('my-feature-gate-name')`. The web server (e.g. Bifrost) would then check the conditional manifest to determine which bundles to serve to any given user, depending on the value of `fg('my-feature-gate-name')`. (The conditional manifest is also checked at runtime when loading an async chunk.)
+
+All of the bundles are still loaded _synchronously_, making conditional imports more versatile and performant than dynamic imports. We discuss the benefits of conditional imports compared to dynamic imports in the "How are conditional imports better than dynamic imports?" section below.
 
 ### Syntax
 
@@ -70,27 +72,38 @@ Caveats:
 >
 > **Bundling complexity:** Simplifying to default export means we can simplify the runtime code. If we don't, that means we have to handle a bunch of bundling edge cases to make sure we get optimised code (e.g. dead code removal, side-effects etc.)
 
-At runtime, the `importCond` function call will be converted to something like this:
+At build time, the `importCond` function call will be converted to something like this:
 
 ```js
 // index.js
-const MyComponent = globalThis.__MCOND('my-feature-gate-name')
-  ? require('./true-component.tsx').default
-  : require('./false-component.tsx').default;
+
+register("avM1e", function(e, t) {
+  e.exports = function(e, ifTrue, ifFalse) {
+      return globalThis.__MCOND && globalThis.__MCOND(e) ? ifTrue() : ifFalse()
+  }
+}),
+register("4qbUE", function(e, t) {
+    e.exports = parcelRequire("avM1e")("my-feature-gate-name", function() {
+        return parcelRequire("esWld");
+    }, function() {
+        return parcelRequire("dqBV0");
+    })
+}),
+// ...
+const MyComponent = parcelRequire("4qbUE");
 ```
 
-`globalThis.__MCOND` is the feature gate function that the user evaluates at runtime. More about this later...
+`globalThis.__MCOND` ("**m**odule **cond**ition") is the feature gate function that the user evaluates at runtime. More about this later...
 
-## How are conditional imports better?
+## How are conditional imports better than dynamic imports?
 
 Before conditional imports, there were two main approaches used in Jira for feature gating different modules:
 
 - Conventional imports, e.g. `componentWithFG`
 - Dynamic imports
 
-The `componentWithFG` function call has the same impact to bundle size and load time as typing `fg('...') ? TrueComponent : FalseComponent` (with a bit of magic to make it work). It uses synchronous imports so there will be no impact to the idle time, but there will be a bundle size impact as the user will always load both versions of a component synchronously:
-
 ```tsx
+// Conventional imports
 import TrueComponent from './true-component.tsx';
 import FalseComponent from './false-component.tsx';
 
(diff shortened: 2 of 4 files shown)


docs/features/Conditional Bundling.md
```diff
@@ -34,7 +34,9 @@ then the user will end up downloading code that they never need to execute. For
...
+Atlaspack's conditional imports API solve this problem by providing a way to hint to the Atlaspack bundler that there is conditional code. At build time, Atlaspack will then ensure that `TrueComponent` and `FalseComponent` are separate bundles in the build output, and it will generate a **conditional manifest** which is a JSON file that details which bundle(s) should be loaded for each value of `fg('my-feature-gate-name')`. The web server (e.g. Bifrost) would then check the conditional manifest to determine which bundles to serve to any given user, depending on the value of `fg('my-feature-gate-name')`. (The conditional manifest is also checked at runtime when loading an async chunk.)
```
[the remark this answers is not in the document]
JakeLane: The manifest isn't checked, but we inline bundle fetching as a dependency for async chunks. See https://github.com/atlassian-labs/atlaspack/blob/main/packages/runtimes/js/src/JSRuntime.js#L581 for the relevant code (and you can use the example in the repo to generate the code).

docs/features/Conditional Bundling.md
```diff
@@ -226,35 +255,112 @@ Now that we have set up the correct bundles to be loaded from the server side, w
...
+  module.exports = function (cond, ifTrue, ifFalse) {
+    return globalThis.__MCOND && globalThis.__MCOND(cond)
+      ? ifTrue()
+      : ifFalse();
+  };
+});
+parcelRegister('e2VvX', function (module, exports) {
+  // function that resolves bundles
+  // see `packages/runtimes/js/src/helpers/bundle-manifest.js`
+  var $dgmOi = parcelRequire('dgmOi');
```
dddlr, line 331: is this a correct description of what `dgmOi` does? i just took an educated guess

docs/features/Conditional Bundling.md
```diff
@@ -226,35 +255,112 @@ Now that we have set up the correct bundles to be loaded from the server side, w
...
+You will need to update the web server to add an implementation for the runtime feature gate function, `__MCOND`. If your feature gate function is called `fg`, it is as simple as:
 
 ```tsx
 // somewhere in your client-side code
 globalThis.__MCOND = (cond: string) => fg(cond);
 ```
 
 You will also need to add this for your Jest configuration so that `importCond` works in unit tests.
+
+### Aside: how conditional imports inside asynchronous bundles are handled
```
dddlr, line 285: can I get a fact check for this section?

docs/features/Conditional Bundling.md
```diff
@@ -34,7 +34,19 @@ then the user will end up downloading code that they never need to execute. For
...
+
+At build time, Atlaspack will parse all `importCond` function calls, then ensure that `TrueComponent` and `FalseComponent` are separate bundles in the build output. It will then generate a **conditional manifest** which is a JSON file that details which bundle(s) should be loaded for each value of `fg('my-feature-gate-name')`. The web server (e.g. Bifrost) would then check the conditional manifest to determine which bundles to serve to any given user, depending on the value of `fg('my-feature-gate-name')`.
+
+> **Note on async bundles**
+>
+> This document covers the case where conditional imports are used in a synchronous bundle. In this case, the web server needs to ensure the correct bundles are loaded on the page, using the bundles listed in the conditional manifest.
+>
+> However, when conditional imports are used in an asynchronous bundle (most of Jira's bundles are asynchronous), Atlaspack will load the conditional bundles asynchronously alongside the asynchronous bundle.
```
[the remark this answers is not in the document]
JakeLane: yes this is correct but a bit confusing wording. Maybe something like

`Atlaspack will load the conditional bundles as a dependency alongside the asynchronous bundle.`

packages/examples/conditional-bundling/package.json
```diff
@@ -5,13 +5,13 @@
...
-    "build:inspect:on": "PARCEL_WORKERS=0 node --inspect-brk $(yarn bin atlaspack) build --no-cache --feature-flag conditionalBundlingApi=true src/index.html",
+    "build:inspect:on": "PARCEL_WORKERS=0 node --inspect-brk $(yarn bin atlaspack) build --no-cache --feature-flag conditionalBundlingApi=true --feature-flag conditionalBundlingNestedRuntime=true --feature-flag conditionalBundlingReporterDuplicateFix=true --feature-flag conditionalBundlingReporterSameConditionFix=true --feature-flag condbHtmlPackagerChange=true src/index.html",
     "build:off": "atlaspack build --no-cache --feature-flag conditionalBundlingApi=false src/index.html",
-    "build:on": "atlaspack build --no-cache --feature-flag conditionalBundlingApi=true --no-optimize src/index.html",
+    "build:on": "atlaspack build --no-cache --feature-flag conditionalBundlingApi=true --feature-flag conditionalBundlingNestedRuntime=true --feature-flag conditionalBundlingReporterDuplicateFix=true --feature-flag conditionalBundlingReporterSameConditionFix=true --feature-flag condbHtmlPackagerChange=true --no-optimize src/index.html",
```
dddlr, line 10: will need to undo once https://github.com/atlassian-labs/atlaspack/pull/640 is merged in, depending on which PR is merged in first

CONTRIBUTING.md
```diff
@@ -16,6 +16,8 @@ For pull requests, please:
 
 For bigger changes, please make sure you start a discussion first by creating an issue and explaining the intended change. -->
 
+> **If you're an Atlassian employee:** you shouldn't sign the CLA. Instead, please search for the "Link your Github & Atlassian accounts" page on Confluence, and follow the instructions there to link your GitHub account to the Atlassian system.
```
pancaspe87, line 19: Is there anything else we need to update? IIRC there may be a task/message when raising a PR

docs/features/Conditional Bundling.md
```diff
@@ -34,7 +34,19 @@ then the user will end up downloading code that they never need to execute. For
...
+>
+> Scroll down to see an example of this.
+
+All of the bundles are still loaded _synchronously_, making conditional imports more versatile and performant than dynamic imports. We discuss the benefits of conditional imports compared to dynamic imports in the "How are conditional imports better than dynamic imports?" section below.
```
pancaspe87, line 49: nit

you can probably link the section itself in the markdown

docs/features/Conditional Bundling.md
```diff
@@ -70,27 +82,38 @@ Caveats:
...
 
-At runtime, the `importCond` function call will be converted to something like this:
+At build time, the `importCond` function call will be converted to something like this:
 
 ```js
 // index.js
-const MyComponent = globalThis.__MCOND('my-feature-gate-name')
-  ? require('./true-component.tsx').default
-  : require('./false-component.tsx').default;
+
```
pancaspe87, line 89: This could be introduced by the description you have in line 39. Maybe move that section on here?

> At build time, Atlaspack will parse all `importCond` function calls, then ensure that `TrueComponent` and `FalseComponent` are separate bundles in the build output. It will then generate a **conditional manifest** which is a JSON file that details which bundle(s) should be loaded for each value of `fg('my-feature-gate-name')`. The web server (e.g. Bifrost) would then check the conditional manifest to determine which bundles to serve to any given user, depending on the value of `fg('my-feature-gate-name')`.

CONTRIBUTING.md
dddlr: Screenshot 2025-06-24 at 13 30 25

not that i know of 🤔

if there's any changes you can think of that we need to make to the pull request template etc, i'll just spin off that and the `CONTRIBUTING.md` changes to a separate PR 😅 as this is technically not in the current PR scope

docs/features/Conditional Bundling.md
```diff
@@ -70,27 +82,38 @@ Caveats:
...
+}),
+// ...
+const MyComponent = parcelRequire("4qbUE");
 ```
 
-`globalThis.__MCOND` is the feature gate function that the user evaluates at runtime. More about this later...
+`globalThis.__MCOND` ("**m**odule **cond**ition") is the feature gate function that the user evaluates at runtime. More about this later...
 
-## How are conditional imports better?
+## How are conditional imports better than dynamic imports?
```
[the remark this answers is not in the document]
dddlr: ah like when we have a bundle that we only want to load when we have a relatively uncommon user interaction, like clicking a button to open a specific modal right? (in this case, i suppose conditional imports wouldn't be the best option because we don't want the bundle to be preloaded)

docs/features/Conditional Bundling.md
```diff
@@ -34,7 +34,19 @@ then the user will end up downloading code that they never need to execute. For
 
 ### The solution
 
-Atlaspack's conditional imports API solve this problem by providing a way to notify the Atlaspack bundler that there is conditional code. At build time, Atlaspack will then ensure that `TrueComponent` and `FalseComponent` are separate bundles in the build output, and it will generate a **conditional manifest** which is a JSON file that details which bundle(s) should be loaded for each value of `fg('my-feature-gate-name')`. The web server (e.g. Bifrost) would then check the conditional manifest to determine which bundles to serve to any given user, depending on the value of `fg('my-feature-gate-name')`.
+Atlaspack's conditional imports API solves this problem by providing a way to hint to the Atlaspack bundler that there is conditional code, and providing the means for the web server to only serve the bundles to the user that correspond to the feature gate values for that user.
```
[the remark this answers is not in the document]
dddlr: hmm thoughts on this?

> Atlaspack's conditional imports API solves this problem by giving a way for the Atlaspack bundler to recognise conditional code. It allows for the web server to then only deliver specific bundles to the user, based on their feature gate values.

i changed a few things, due to nitpicky stuff like how atlaspack doesn't really ensure that the web server serves the correct bundles (it just assumes that the web server will serve the right bundles and complains loudly if not), and how feature gate settings is ambiguous

docs/features/Conditional Bundling.md
dddlr: done, let me know how it looks

docs/features/Conditional Bundling.md
```diff
@@ -226,35 +256,112 @@ Now that we have set up the correct bundles to be loaded from the server side, w
...
 ```tsx
 // somewhere in your client-side code
 globalThis.__MCOND = (cond: string) => fg(cond);
 ```
 
 You will also need to add this for your Jest configuration so that `importCond` works in unit tests.
+
+### Aside: how conditional imports inside asynchronous bundles are handled
+
+As noted earlier, when conditional imports are used in an asynchronous bundle, Atlaspack will add a special loader to load the conditional bundles as a dependency alongside the asynchronous bundle.
```
JakeLane, line 288:
```suggestion
As noted earlier, when conditional imports are used in an asynchronous bundle, Atlaspack will add to the loader the conditional bundles as a dependency alongside the asynchronous bundle.
```

docs/features/Conditional Bundling.md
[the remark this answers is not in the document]
pancaspe87: Sounds good :)

PR #635, opened 2025-06-20
Add custom GitHub action and pull request template to enforce that documentation is updated

dddlr:
Added a pull request template with a checklist that asks that documentation be updated for every PR.

This is enforced by a custom GitHub action that checks whether any tasks in the checklist are marked as incomplete. If any are incomplete, the GitHub action will give an error.

You can exempt parts the PR description from the incomplete task check by adding  and  around the sections where you want to disable the check.

## Motivation

Compiled has something similar: https://github.com/atlassian-labs/compiled/pull/1694

I copied that over to Atlaspack.

## Changes

* Updated PR template to include updating docs
* Added GitHub action to enforce that all items in the checklist are ✅ 

## Checklist

[no-changeset]: does not affect atlaspack code


dddlr:
> LGTM! Is it worth noting in the [CONTRIBUTING.md](https://github.com/atlassian-labs/atlaspack/blob/main/CONTRIBUTING.md) file that features should be appropriately documented via the docs folder?

good idea, let me add that

CONTRIBUTING.md
```diff
@@ -12,6 +12,8 @@ For pull requests, please:
 - Follow the existing style
 - Separate unrelated changes into multiple pull requests
 
+You should also add documentation to `docs/` if you are adding new features or adding new API options. Note that our documentation is currently very barebones, so every little bit helps!
```
matt-koko, line 15: nice

matt-koko (approved):

dddlr:
hi @pancaspe87

> Even though we had the templates, this was not preventing the PR to land?

currently on master, you could have incomplete tasks in the checklist and that wouldn't block the PR from being merged (because there was no github action that would look for incomplete tasks)

after this PR is landed, those incomplete tasks in the checklist will block the PR from being merged, because of the github action we added

> Can we just click on the checkbox?

yes 😢  i am pinning my hopes on the fact that we will think twice before blindly ticking checkboxes 😅 there's only so far one could go with enforcement

PR #636, opened 2025-06-20
Migrate @atlaspack/cli to TypeScript

alshdavid:
Migrate `@atlaspack/cli` to TypeScript

## Motivation

🌊
🔥

## The Plan

- The entire codebase will be in TypeScript 
  - Will raise a PR package-by-package, starting with the packages that no internal packages depend on
  - Packages will have missing types during the transition
  - All `TS:MIGRATE` comments will be addressed before the end of the migration
- The build system will be simplified to only do type stripping
- `@babel/register` and `esbuild-register` will be removed, relying on the built-in TypeScript support in Nodejs

## Changes

- Removed self build
- Added `lerna run build-tsc` for TypeScript packages
  - Will eventually use `swc` to type strip, but keeping it with `tsc` for simplicity for now 
- Migrated `@atlaspack/cli` to TypeScript

## Checklist

[no-changeset]: WIP


packages/dev/babel-register/index.js
alshdavid, line 1: reserve `.mts` and `.cts` for the built-in typescript transpiler. Babel register will eventually be removed from the repo entirely

packages/core/cli/bin/atlaspack.js
alshdavid, line 1: This is gated on Nodejs's built-in contitions

```bash
# Use "lib"
npx atlaspack --version

# Use TypeScript sources
NODE_OPTIONS="--condition @atlaspack::sources" npx atlaspack --version
```

PR #638, opened 2025-06-20
TypeScript Ep.1 - Initial changes

alshdavid:
## Motivation

Provide relevant context and motivation behind the change, if any

## Changes

Describe the solution and changes you have made

## Checklist

- [ ] Existing or new tests cover this change
- [ ] There is a changeset for this change, or one is not required

PR #639, opened 2025-06-21
TypeScript Ep.2 - @atlaspack/diagnostic

alshdavid:
## Motivation

## Changes

## Checklist

[no-changeset]: TODO


flow-typed/atlaspack/diagnostic.js
```diff
@@ -0,0 +1,270 @@
+declare module "@atlaspack/diagnostic" {
```
yamadapc, line 1: This was generated with the `flowgen`?

packages/core/diagnostic/package.json
```diff
@@ -11,18 +11,28 @@
     "url": "https://github.com/atlassian-labs/atlaspack.git"
   },
   "main": "lib/diagnostic.js",
-  "source": "src/diagnostic.js",
```
yamadapc, line 14: Isn't it breaking register?

packages/core/diagnostic/src/diagnostic.mts
```diff
@@ -234,18 +229,18 @@ export default class ThrowableDiagnostic extends Error {
...
+    | {
+        data: any;
+        pointers: {[key: string]: jsonSourcemap.Mapping};
+      },
+  ids: Array<{key: string; type?: 'key' | 'value'; message?: string}>,
 ): Array<DiagnosticCodeHighlight> {
   let map =
     typeof data == 'string'
-      ? parse(data, undefined, {dialect: 'JSON5', tabWidth: 1})
+      ? jsonSourcemap.default.parse(data, undefined, {dialect: 'JSON5', tabWidth: 1})
```
yamadapc, line 240: Why `.default`?

packages/core/diagnostic/lib/index.js.flow
```diff
@@ -0,0 +1,270 @@
+// @flow
```
alshdavid, line 1: This can be generated with `node ./scripts/flowgen index.d.ts index.js.flow` however the generated flow types break the `flow check` because they are too imprecise. 

Managing the exports manually is a temporary solution for this

marcins: Is it possible to validate they're correct / comprehensive at all? Did you at least start with the generated types, and then refine?

packages/core/diagnostic/src/index.mts
```diff
@@ -1,106 +1,102 @@
...
 import nullthrows from 'nullthrows';
-import {parse, type Mapping} from '@mischnic/json-sourcemap';
+import * as jsonSourcemap from '@mischnic/json-sourcemap';
 
 /** These positions are 1-based (so <code>1</code> is the first line/column) */
-export type DiagnosticHighlightLocation = {|
-  +line: number,
-  +column: number,
-|};
+export type DiagnosticHighlightLocation = {
```
marcins, line 6: I guess there's no exact object types in TypeScript?

pancaspe87 (approved):

packages/core/diagnostic/src/index.mts
```diff
@@ -1,106 +1,102 @@
...
-  language?: string,
-  codeHighlights: Array<DiagnosticCodeHighlight>,
-|};
+  language?: string;
+  codeHighlights: Array<DiagnosticCodeHighlight>;
+};
 
 /** A JSON object (as in "map") */
-type JSONObject = {
+export type JSONObject = {
```
marcins, line 48: Does this need to be exported? It wasn't before. I'm guessing it's for generating types somewhere else.

package.json
```diff
@@ -52,7 +52,7 @@
     "test:js": "yarn test:js:unit && yarn test:integration",
     "test:js:coverage": "yarn nyc yarn test:js:unit",
     "test:js:coverage:report": "yarn nyc report --reporter=html-spa",
-    "test:js:unit": "cross-env NODE_ENV=test mocha --conditions=atlaspack::sources --timeout 5000",
+    "test:js:unit": "cross-env NODE_ENV=test mocha --disable-warning=ExperimentalWarning --conditions=atlaspack::sources --experimental-strip-types --timeout 5000",
```
pancaspe87, line 55: Is this an unrelated change to migration?

alshdavid: This actually should be moved to the [first PR](https://github.com/atlassian-labs/atlaspack/pull/638). I've updated that PR to include it there.

It's needed to support TypeScript in the unit tests (at least for now)

packages/core/diagnostic/lib/index.js.flow
alshdavid: They are validated with `npx flow check`/`yarn check`. The packages that depend on them will error if the types don't fit properly.

I started with the generated types and refined them using the original public API (in flow) until the flow checks pass

packages/core/diagnostic/src/index.mts
alshdavid: TypeScript's objective is to apply a type system that models JavaScript's unmodified behaviours rather than be a language unto itself.

JavaScript doesn't have a concept of exact objects, only structural "type contracts" so TypeScript hasn't modeled exact objects or opaque types.

...Just don't ask about `enums` & `namespaces`

matt-koko:
the `feature-flags` package LGTM

liamqma (approved):

alshdavid:
Going to try the codemod approach here: https://github.com/atlassian-labs/atlaspack/pull/696

PR #641, opened 2025-06-23
Option to use compiled Atlaspack in the integration tests

alshdavid:
## Motivation

Decouple the version of Nodejs used in the integration tests from the version of Nodejs used during development of Atlaspack.

We need newer Nodejs features for development but Atlaspack itself should be able to run with an older version of Nodejs

## Changes

Removed the Atlaspack self build and inline env vars plugin

Added env var which is used to tell the Atlaspack development JIT transformers (`@atlaspack/register`) to point to `./lib` rather than `./src`.

```bash
# Manually
ATLASPACK_REGISTER_USE_LIB=true yarn test:integration

# Included in
yarn test:integration-ci
```

### Tests Skipped

I have fixed as many tests as I can but this is a list of the skipped tests. They fail because:

- Some tests have dependencies on ./src objects and paths
- Some tests use object spys against src targets while the test runs against lib code
   - spys and module mocks won't work in newer versions of Nodejs+Typescript so we may need to consider alternatives (they are more of a unit test thing anyway)
- Some tests just don't like me

**Skipped tests:**

- packages/core/integration-tests/test/cache.js
  - 'should support yarn pnp'
  - 'should support adding a ${name}'
  - 'should support adding a nested ${name}'
  - 'should support adding a .babelignore'
  
## Checklist

[no-changeset]: CI

### DIFF
--- a/packages/core/integration-tests/test/cache.js
+++ b/packages/core/integration-tests/test/cache.js
@@ -28,6 +28,7 @@ import {
   findAsset,
   bundle,
   fsFixture,
+  USE_LIB,
 } from '@atlaspack/test-utils';
 import {md} from '@atlaspack/diagnostic';
 import fs from 'fs';
@@ -223,7 +224,9 @@ let packageManager = new NodePackageManager(inputFS, '/');
               await overlayFS.unlink(path.join(inputDir, 'src/nested/test.js'));
             });
           },
-          {message: "Failed to resolve './nested/test' from './src/index.js'"},
+          {
+            message: "Failed to resolve './nested/test' from './src/index.js'",
+          },
         );
       });
 
@@ -241,7 +244,9 @@ let packageManager = new NodePackageManager(inputFS, '/');
           async () => {
             await runBundle();
           },
-          {message: "Failed to resolve './nested/test' from './src/index.js'"},
+          {
+            message: "Failed to resolve './nested/test' from './src/index.js'",
+          },
         );
       });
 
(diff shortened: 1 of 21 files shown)


packages/core/integration-tests/test/domain-sharding.js
alshdavid, line 1: This appears to produce different hashes when using `./lib` rather than `./src`. @benjervis is that expected?

packages/core/test-utils/src/assertCode.js
alshdavid, line 1: Credit @mattcompiles https://github.com/atlassian-labs/atlaspack/pull/469/files#diff-9bf8c87eaf6350222dbfe7aea3da725e0c33f7eef89bf5deb1d6dd094a0c1ac6R217

packages/core/integration-tests/test/cache.js
```diff
@@ -7104,7 +7108,10 @@ let packageManager = new NodePackageManager(inputFS, '/');
 
   describe('environment caching', function () {
     it('should cache and load environments between builds', async function () {
-      const EnvironmentManager = require('@atlaspack/core/src/EnvironmentManager');
+      const EnvironmentManager = eval('require')(
```
yamadapc, line 7111: What's this?

alshdavid: Oh, right. This [JIT babel transformer](https://github.com/atlassian-labs/atlaspack/blob/6d9e9939c74913685005b01acc6dc991a7d76c52/packages/dev/babel-register/babel-plugin-module-translate.js) looks for `require`/`import` statements that import local specifiers and rewrites them.

This is my hacky approach to hiding the require from babel-register and preserve the URL

packages/core/test-utils/test/fsFixture.test.js
alshdavid, line 1: This is a unit test for the integration test utils (i think)

PR #642, opened 2025-06-24
De-duplicate ifTrueBundles and ifFalseBundles arrays when creating conditional manifest

dddlr:
Conditional bundling bug fix:

Add de-duplication so that assets don't appear duplicated inside the conditional manifest.

In other words, we want to de-duplicate stuff like this:
Screenshot 2025-06-24 at 16 00 46

## Motivation

When `importCond` is used twice in the same bundle with the exact same arguments, we will add the conditional bundles to `ifTrueBundles` and `ifFalseBundles` twice.

For example, imagine we have two files that get bundled into the same asset (let's call it `navigation-stuff.4abe98ff.js`) at build-time:

```jsx
// src/navigation/dev-panel/index.tsx

import React from 'react';

const Component = importCond('my-feature-gate', '@private/component/index-new.tsx', '@private/component/index-old.tsx')
export const DevPanel = (props) => {
    return (
        <>
            <DevStuff />
            <Component ... />
            {/* and so on... */}
        </>
    );
};
```

```jsx
// src/navigation/product-panel/index.tsx

import React from 'react';

const Component = importCond('my-feature-gate', '@private/component/index-new.tsx', '@private/component/index-old.tsx')
export const ProductPanel = (props) => {
    return (
        <>
            <ProductStuff ... />
            <Component ... />
            {/* and so on... */}
        </>
    );
};
```

`index-new.js` and `index-old.js` will get added to the `ifTrueBundles` and `ifFalseBundles` arrays for `navigation-stuff.4abe98ff.js` _twice_, because there are two `importCond` function calls that end up in the same bundle.

## Changes

Add some basic de-duplication in `packages/reporters/conditional-manifest/src/ConditionalManifestReporter.js`, which is where we create the conditional manifest.

I considered adding this de-duplication into `getConditionalBundleMapping` directly too, but it seemed superfluous - @JakeLane let me know if you have a different opinion here.

Note that this PR will need to be updated once https://github.com/atlassian-labs/atlaspack/pull/640 is merged in

## Checklist

- [x] Make a PR to add de-duplication to the old Jira SSR too

### DIFF
--- a/.changeset/kind-doors-change.md
+++ b/.changeset/kind-doors-change.md
@@ -1,0 +1,9 @@
+---
+'@atlaspack/reporter-conditional-manifest': patch
+'@atlaspack/integration-tests': patch
+'@atlaspack/feature-flags': patch
+'@atlaspack/bundler-default': patch
+'@atlaspack/utils': patch
+---
+
+Fix duplicate assets appearing in conditional manifest when two conditional imports use the same file paths
--- a/packages/core/integration-tests/test/conditional-bundling.js
+++ b/packages/core/integration-tests/test/conditional-bundling.js
@@ -1090,4 +1090,81 @@ describe('conditional bundling', function () {
       },
     });
   });
+
+  it(`should de-duplicate bundles, if the same arguments are passed in two conditional bundling usages`, async function () {
+    const dir = path.join(
+      __dirname,
+      'import-cond-cond-manifest-same-condition-and-paths',
+    );
+    await overlayFS.mkdirp(dir);
+
+    await fsFixture(overlayFS, dir)`
+        .parcelrc:
+          {
+            "extends": "@atlaspack/config-default",
+            "reporters": [
+              "@atlaspack/reporter-conditional-manifest",
+              "..."
+            ]
+          }
+
+        index.js:
+          import { Component1 } from './c';
+          import { Component2 } from './d';
+
+          export const App = (props) => {
+            Component1;
+            Component2;
+          };
+
+        a.js:
+          export default 'module-a';
+
+        b.js:
+          export default 'module-b';
+
+        c.js:
+          import React from 'react';
+          const Imported1 = importCond('cond', './a', './b');
+          export const Component1 = () => { return Imported1 };
+
+        d.js:
+          import React from 'react';
+          const Imported2 = importCond('cond', './a', './b');
+          export const Component2 = () => { return Imported2 };
+      `;
+
+    let bundleGraph = await bundle(path.join(dir, '/index.js'), {
+      inputFS: overlayFS,
+      featureFlags: {
+        conditionalBundlingApi: true,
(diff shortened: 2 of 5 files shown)


packages/core/integration-tests/test/conditional-bundling.js
```diff
@@ -1090,4 +1090,81 @@ describe('conditional bundling', function () {
       },
     });
   });
+
+  it(`should de-duplicate bundles, if the same arguments are passed in two conditional bundling usages`, async function () {
```
dddlr, line 1094: ah i just realised how to get JSX syntax working, hold on while i update the PR

.changeset/kind-doors-change.md
```diff
@@ -0,0 +1,9 @@
+---
+'@atlaspack/reporter-conditional-manifest': patch
+'@atlaspack/integration-tests': patch
+'@atlaspack/feature-flags': patch
+'@atlaspack/bundler-default': patch
```
dddlr, line 5: `@atlaspack/bundler-default` and `@atlaspack/utils` were included by default, is this correct?

packages/reporters/conditional-manifest/src/ConditionalManifestReporter.js
```diff
@@ -72,14 +72,33 @@ export async function report({
...
-            ifFalseBundles: mapBundles(cond.ifFalseBundles)
-              .concat(bundleInfo[key]?.ifFalseBundles ?? [])
-              .sort(),
-          };
+          if (getFeatureFlag('conditionalBundlingDeduplicateBundles')) {
+            bundleInfo[key] = {
+              ifTrueBundles: [
+                ...new Set(
+                  mapBundles(cond.ifTrueBundles)
+                    .concat(bundleInfo[key]?.ifTrueBundles ?? [])
```
marcins, line 80: Probably not important right now but might be a factor as the feature scales - but does it make sense to do the sort/set for every bundle, or do it once as a pass at the end?

Maybe even consider something like using two maps like `ifTrueBundles: Map<string, Set<string>> = new Map()` (and one for false), adding to the sets in this pass, and at the end producing the `bundleInfo` object and sorting the sets then.

.changeset/kind-doors-change.md
marcins: Is it because of the change in feature flags? Though I'd expect more packages to be touched by that?

packages/reporters/conditional-manifest/src/ConditionalManifestReporter.js
JakeLane: @dddlr might be a good opportunity to improve this code, given you've raised a PR to mitigate in jira - you can take your time on this one

JakeLane (approved):
looks good! Did you check if the JSRuntime code needs to be updated?

PR #652, opened 2025-06-26
Fix state corruption by disabling build aborting

yamadapc:
This aims to fix two known cases where internal state would be corrupted due to
a build being aborted.

The first would cause `PackagerRunner::loadConfig` to throw due to
`bundle.name` being undefined. We have observed this happening after build
aborts, but are unsure which request is doing the non-transactional update to a bundle.

Possibly the `applyRuntimes` function seems to create some bundles, then run
requests, then do more patches to those bundles. On the event of a a build
abort, the second section of that function (between runRequest boundaries)
would not run.

Another case is in `DepDevRequest::getDevDepRequests`. On this function, we are
listing all dev dep requests on the graph and then asserting they have a result
stored.

However, in the case of a build abort, those requests may exist in the graph
but not have been executed. It is also reasonable that there might be other
cases where those nodes exist in the graph but haven't yet been executed at
time of packaging.

For this case, we filter out dev-dep requests that have missing results instead
of throwing.

As a catch-all solution, we will remove all build aborting functionality.
Generally, the incremental builds are fast enough that we don't need to
maintain this. We should revisit this functionality with more transactional
semantics and less distributed mutable state.

The cache writes are not particularly fast, however, the sections of the cache
write that can be aborted aren't the bottle-neck. Furthermore, once a cache
write is already executing it is unlikely that aborting it is a good idea, as
the user should benefit from the cache being written. Instead, we should write
to cache less often (not on every build end) or incrementally (break-up the
results and write them immediately after running the task.

Test Plan: N/A


JakeLane (approved):

yamadapc:
I have removed cache write abort changes from this diff.

PR #653, opened 2025-06-26
Improve option invalidation granularity and reduce unnecessary cache invalidations

csowden-atlassian:
## Motivation

Our build cache system was invalidating more often than necessary due to changes in options that don't actually affect build output. For example:

- Changes to `instanceId` would invalidate the cache even though it doesn't affect builds
- Reading `Object.keys()` on option objects would trigger invalidation  
- Changes to nested feature flag values would invalidate entire feature flag object
- No way to blocklist non-impactful options from triggering invalidation

This resulted in unnecessary rebuilds and reduced development performance, especially in watch mode where options might change frequently.

## Changes

Implemented a more granular option invalidation system with the following improvements:

1. **Path-based tracking**: Track exact paths of accessed options (e.g., `['featureFlags', 'granularOptionInvalidation']`) instead of just top-level keys
   ```js
   // Before: Invalidates all of featureFlags
   options.featureFlags.someFlag
   // After: Only invalidates specific flag
   invalidateOnOptionChange(['featureFlags', 'someFlag'])
   ```
2. **Option blocklist**: Added configuration to prevent certain options from triggering invalidation:
```js
optionInvalidation: {
  blocklist: ['instanceId', 'metrics.*'],
  trackMetrics: true
}
```
   3. **Smarter enumeration handling**:  Skip invalidation for `Object.keys()` operations when they don't affect build output
   4. **Invalidation metrics**:  Added detailed tracking of which options cause the most invalidations
   5. **Feature flag controlled**: All changes are behind granularOptionInvalidation flag for safe rollout

   The changes maintain backward compatibility while significantly reducing unnecessary cache invalidations.

## Changes
Initial testing shows:
- Reduced cache invalidations by preventing non-impactful options from triggering rebuilds
- More precise invalidation of nested options
- Better visibility into what causes cache invalidations through metrics

## Checklist
- [x] Extensive test suite added covering all new functionality
- [x] Added integration tests for cache behavior
- [x] Added documentation for option invalidation configuration
- [x] Maintained backward compatibility
- [x] Feature flag controlled for safe rollout


packages/core/core/src/RequestTracker.js
```diff
@@ -506,29 +522,172 @@ export class RequestGraph extends ContentGraph<
...
+          for (let parentNode of parentNodes) {
+            this.invalidateNode(parentNode, OPTION_CHANGE);
+          }
+        }
+      }
+
+      return invalidatedKeys;
+    }
+
+    // New behavior with granular path tracking and blocklist
```
csowden-atlassian, line 563: The option invalidation blocklist allows users to specify configuration options that shouldn't trigger rebuilds when changed. This improves performance by preventing unnecessary invalidations for options that rarely affect build output.

The implementation:
1. Uses a default blocklist for options known to be non-impactful (e.g., 'instanceId')
2. Supports user-configured blocklists through options.optionInvalidation.blocklist
3. Supports prefix-based blocking with wildcard notation (e.g., 'metrics.*')
4. Optionally tracks skipped invalidations for reporting and diagnostics

packages/core/core/src/public/Config.js
```diff
@@ -65,17 +69,47 @@ export function makeConfigProxy<T>(
...
   };
 
   const makeProxy = (target, path) => {
     return new Proxy(target, {
       ownKeys(target) {
-        reportPath(path);
+        // Check if we should track object enumeration operations
+        // This is controlled by a feature flag - the previous behavior was to track these
+        // but it can cause over-invalidation
+        const granularOptionInvalidationEnabled = getFeatureFlag(
```
marcins, line 102: Unnecessary to assign this to a variable? (i.e. just stick the `getFeatureFlag` call inside the `if` - makes it easier to cleanup)

packages/core/core/src/utils.js
```diff
@@ -125,8 +175,32 @@ function proxyPackageManager(
...
+ * @param {mixed} value - The option value to hash
+ * @returns {string} A string hash representing the value
+ */
 export function hashFromOption(value: mixed): string {
-  if (typeof value === 'object' && value != null) {
+  if (value == null) {
+    return String(value);
+  }
+
+  if (typeof value === 'object') {
```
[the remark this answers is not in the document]
marcins: How does that work when an element that's not in the hash changes that doesn't affect the length? Are these arrays of options from config getting passed here? Do we actually have any large collections in practice?

We hash multi-megabyte bundles, I'd argue that hashing a string with hundreds of options concatenated can't be too expensive.

packages/core/core/test/ConfigProxy.test.js
```diff
@@ -0,0 +1,318 @@
...
+  it('tracks reads to nested fields with path arrays', () => {
+    const onRead = sinon.spy();
+    const target = {a: {b: {c: 'd'}}};
+    const config = makeConfigProxy(onRead, target);
+    config.a.b.c;
+    assert.ok(onRead.calledWith(['a', 'b', 'c']));
+    assert.ok(onRead.calledOnce);
+  });
+
+  it('handles Object.keys() operations based on the granularOptionInvalidation feature flag', () => {
```
marcins, line 18: Should the test run instead with `granularOptionInvalidationEnabled` set to `true` and again set to `false` (FWIW with Mocha I've been doing this lately for some tests:

```
// intentionally verbose so it's obvious rather than just `[true, false].forEach`
[{myFlag: true}, {myFlag: false}].forEach(({myFlag}) => {
   it(`does something (myFlag: ${myFlag.toString()})`, () => {
      .. set the feature flag whether it's in package.json or by using `setFeatureFlags`)
   }
});
```
Example: https://github.com/atlassian-labs/atlaspack/blob/4d7abde0a6faa148bcf330b08d820f171f853b76/packages/core/integration-tests/test/scope-hoisting.js#L3620

packages/core/core/test/OptionInvalidationIntegration.test.js
```diff
@@ -0,0 +1,211 @@
...
+describe.skip('Option invalidation in cache integration test', () => {
+  let inputDir;
+
+  beforeEach(async () => {
+    inputDir = path.join(__dirname, '../../../integration-tests/input');
+    await inputFS.rimraf(inputDir);
+    await inputFS.mkdirp(inputDir);
+  });
+
+  async function createSimpleProject() {
```
marcins, line 38: You can use `fsFixture` for inline test code for integration tests that makes this kind of setup easier to reason about.. Just search the codebase, plenty of examples. Avoids the need for physical files for fixtures as well.

marcins (approved):
Nice one! Some minor feedback, but mostly looks good! (TBF I didn't go to deep on the hundreds of lines of new tests... but appreciate they're there!)

packages/core/core/src/RequestTracker.js
```diff
@@ -280,11 +281,26 @@ const nodeFromEnv = (env: string, value: string | void): RequestGraphNode => ({
...
+ * @param {string[] | string} option - Option path as an array or dot-notation string
+ * @param {mixed} value - The current value of the option
+ * @returns {RequestGraphNode} The created node
+ */
+const nodeFromOption = (
+  option: string[] | string,
+  value: mixed,
+): RequestGraphNode => {
+  // Normalize option to string format for node ID
+  const optionKey = Array.isArray(option) ? option.join('.') : option;
```
yamadapc, line 297: I think we should `JSON.stringify` this

packages/core/core/src/utils.js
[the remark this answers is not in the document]
yamadapc: I also think we should hash everything

packages/core/core/src/public/Config.js
```diff
@@ -42,8 +44,10 @@ const internalConfigToConfig: DefaultWeakMap<
  *
  * In case the value is null or an array, we will track the read as well.
  *
- * Iterating over `Object.keys(obj.field)` will register a read for the `['field']` path.
- * Other reads work normally.
+ * NOTE: By default, we DO track Object.keys()/ownKeys operations for backward compatibility,
+ * but this can be disabled with the 'granularOptionInvalidation' feature flag. When disabled,
+ * property enumeration won't trigger cache invalidation, which prevents unnecessary invalidations
```
yamadapc, line 49: Enumerating a field should invalidate when it changes.

Because output from code listing a property might change if there are new items or keys.

It there code that was triggering this? Maybe we can try to rework it to avoid invalidation.

csowden-atlassian: [specific commit](https://github.com/atlassian-labs/atlaspack/pull/653/commits/b65c1e34d2f60df6beb112982af63d8d2eb74832) - was a bit of a tricky one to figure out, but I'm fairly confident this addresses the root cause

packages/core/integration-tests/test/option-invalidation.js
```diff
@@ -0,0 +1,65 @@
...
+        export const value = "test";
+    `;
+
+    // Test that the basic setup works with feature flag enabled
+    const firstBuild = await bundle(path.join(dir, '/index.js'), {
+      inputFS: overlayFS,
+      defaultConfig: path.join(dir, '.parcelrc'),
+    });
+
+    const secondBuild = await bundle(path.join(dir, '/index.js'), {
```
yamadapc, line 56: ideally there'd be a test for the option changes not invalidating caches ; there are a few similar tests in `cache.js`

packages/core/core/src/requests/ConfigRequest.js
```diff
@@ -89,8 +89,14 @@ export async function loadPluginConfig<T: PluginWithLoadConfig>(
     config.result = await loadConfig({
       config: new PublicConfig(config, options),
       options: new PluginOptions(
-        optionsProxy(options, (option) => {
-          config.invalidateOnOptionChange.add(option);
+        optionsProxy(options, (optionPath) => {
+          // Check if we're using the new granular option tracking (path arrays)
+          if (Array.isArray(optionPath)) {
+            config.invalidateOnOptionChange.add(optionPath.join('.'));
```
yamadapc, line 95: We should change the type of `invalidateOnOptionChange`

packages/core/core/src/types.js
```diff
@@ -332,6 +336,20 @@ export type AtlaspackOptions = {|
...
+  // Configuration for option invalidation
+  optionInvalidation?: {|
+    // Complete list of options to block from invalidation
+    blocklist?: Array<string>,
+    // Options that should be deep-compared instead of just hash-compared
+    deepCompare?: Array<string>,
+    // Maximum number of parent nodes to invalidate per option change
+    maxInvalidationsPerOption?: number,
+    // Batch option invalidations together
+    batchInvalidations?: boolean,
```
yamadapc, line 349: What does this and the `maxInvalidationsPerOptions` keys do?

packages/core/core/src/Transformation.js
```diff
@@ -123,7 +123,14 @@ export default class Transformation {
       optionsProxy(
         this.options,
         (option) => {
-          this.invalidations.invalidateOnOptionChange.add(option);
+          // Check if we're using the new granular option tracking (path arrays)
+          if (Array.isArray(option)) {
+            // Convert array paths to dot-notation for compatibility with Set<string>
+            this.invalidations.invalidateOnOptionChange.add(option.join('.'));
```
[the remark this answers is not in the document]
csowden-atlassian: Updated

packages/core/core/src/types.js
csowden-atlassian: This isn't actually used - just needed cleaning up

packages/core/core/src/utils.js
```diff
@@ -84,19 +99,58 @@ export function optionsProxy(
...
+    // return {...proxiedOptions, packageManager};
+    // This would trigger Object.keys() enumeration on the proxy object.
+    // (The ownKeys trap is defined in the makeConfigProxy function in Config.js)
+    //
+    // The spread operator (...) internally calls Object.keys() which triggers the ownKeys trap
+    // on proxied objects. For proxies that track property access, this can cause performance
+    // issues when done repeatedly, as it enumerates all properties instead of accessing specific ones.
+    // Object.assign() uses getOwnPropertyDescriptor() instead, avoiding the ownKeys trap.
+    // $FlowFixMe[incompatible-exact] - Object.assign creates inexact type, but proxy contains all AtlaspackOptions properties
+    return Object.assign({}, proxiedOptions, {packageManager});
```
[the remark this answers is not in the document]
csowden-atlassian: Np - I reworked this to handle ignored items inside `makeConfigProxy`

PR #654, opened 2025-06-27
Fix issues with scope hoisting optimisation and inline constants

marcins:
## Motivation

If `unstable_inlineConstants` is enabled along with `applyScopeHoistingImprovement` then things break, this is because constant modules end up wrapped but not the references.

## Changes

To fix this there are two changes:
* don't wrap constant modules when `applyScopeHoistingImprovement` is enabled
* bring in the fix from `inlineConstOptimisationFix` to render all constant modules at the top of the bundle when `applyScopeHoistingImprovement` is enabled.

The integration tests for the wrapping changes have been modified to run with inline constants enabled and disabled to verify (they were failing before the fixes)

## Checklist


packages/core/integration-tests/test/scope-hoisting.js
```diff
@@ -3616,74 +3616,79 @@ describe('scope hoisting', function () {
...
-          defaultTargetOptions: {
-            sourceMaps: false,
-          },
-        },
-      );
-
-      let constants = ['BLOGGER', 'PREMIUM', 'MONTHS_IN_YEAR'];
+    describe('constant inlining', function () {
+      [
+        {applyScopeHoistingImprovement: true},
```
marcins, line 3621: Strictly speaking the const inlining tests were passing for both branches without the fix - I guess they aren't comprehensive enough..

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -707,7 +714,10 @@ export class ScopeHoistingPackager {
                   // assets referenced by this asset will also be wrapped. Otherwise, inline the
                   // asset content where the import statement was.
                   if (getFeatureFlag('applyScopeHoistingImprovement')) {
-                    if (!this.wrappedAssets.has(resolved.id)) {
+                    if (
+                      !resolved.meta.isConstantModule &&
```
mattcompiles, line 718: Is this needed? Don't the changes above ensure constant modules are never wrapped?

mattcompiles (approved):
Nice one bud

PR #643, opened 2025-06-24
[no-changeset] introduce ENV variables docs

pancaspe87:
[no-changeset]: Another PR to improve our docs.

Created through _multiple iterations_ with Cursor AI assistance.

----
As mentioned in previous [PR](https://github.com/atlassian-labs/atlaspack/pull/624)

> 
Over the upcoming weeks, I'll try to improve our docs by leveraging AI tools and also checking what is available in the Parcel website. Things may change/evolve over time as I learn how to best leverage tools; eventually we could explore tools that could update/auto-generate documentation as we make changes.

PR #659, opened 2025-07-01
Add a handful of debugging tools

benjervis:
## Motivation

When I'm trying to identify issues with Atlaspack builds, there are a couple of things I always change to my local version to make things a bit easier.

Probably makes sense to move them out of my permanent stash and into the repo for everyone to enjoy.

## Changes

This PR introduces the concept of "debug tools", and implements the first two. It's intended that we can add others in future as ideas come up.

Debug tools are enabled via an env var (which I'm setting permanently in my shell config), with instructions added to the `debug-tools.js` file.

The two tools that are implemented here are:

### Simple CLI Reporter

When doing a build in our internal applications, several thousand bundles are created, and listing the file path of every single one takes up an annoying amount of space in the terminal and is not useful to me at all. It also means that if I have any console.log entries added, I have to scroll way up to see them.

Adding the `simple-cli-reporter` tool will change this output to just report the number of bundles created.

### Add asset paths to output

When looking at an output bundle, it can be difficult to know exactly what asset a particular `parcelRegister` call corresponds to. This is often important information, to know how the source code is structured and to be able to find things like the package.json it corresponds to.

Adding the `asset-file-names-in-output` tool will insert comments with the file path of the asset (relative to the project root):

- Above the `parcelRegister` statement when scope hoisting
- Wrapping a block of code that has been scope hoisted into a `parcelRegister` body
- At the start of the module function in the Dev Packager

## Checklist

This is only for internal Atlaspack developers, so no public documentation for this.

### DIFF
--- a/packages/core/utils/src/debug-tools.js
+++ b/packages/core/utils/src/debug-tools.js
@@ -11,7 +11,12 @@
  * You can enable all tools by setting `ATLASPACK_DEBUG_TOOLS=all`.
  */
 
-export let debugTools = {
+type DebugTools = {|
+  'asset-file-names-in-output': boolean,
+  'simple-cli-reporter': boolean,
+|};
+
+export let debugTools: DebugTools = {
   'asset-file-names-in-output': false,
   'simple-cli-reporter': false,
 };
--- a/packages/core/utils/src/debug-tools.js
+++ b/packages/core/utils/src/debug-tools.js
@@ -1,8 +1,8 @@
 // @flow
 
 /*
- * These tools are intended for Atlaspack developers, to provide extra tools to
- * make debugging Atlaspack issues more straightforward.
+ * These tools are intended for Atlaspack developers, to provide extra utilities
+ * to make debugging Atlaspack issues more straightforward.
  *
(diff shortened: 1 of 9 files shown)


packages/core/utils/src/debug-tools.js
```diff
@@ -0,0 +1,45 @@
...
+
+/*
+ * These tools are intended for Atlaspack developers, to provide extra utilities
+ * to make debugging Atlaspack issues more straightforward.
+ *
+ * To enable a tool, set the `ATLASPACK_DEBUG_TOOLS` environment variable to a
+ * comma-separated list of tool names. For example:
+ * `ATLASPACK_DEBUG_TOOLS="asset-file-names-in-output,simple-cli-reporter"`
+ *
+ * You can enable all tools by setting `ATLASPACK_DEBUG_TOOLS=all`.
```
pancaspe87, line 11: Got an open [PR](https://github.com/atlassian-labs/atlaspack/pull/643) with ENV variables, but tbh you could just add a quick .md file in the docs file with some debugging notes

marcins (approved):
Awesome stuff! I have definitely hacked in something like the asset mapping thing before, very useful.

PR #661, opened 2025-07-01
Cleanup the NAPI Worker Pool when disposing of the Atlaspack instance

marcins:
## Motivation

If Atlaspack V3 is run via the API (i.e. `new Atlaspack( { featureFlag: { atlaspackV3: true } })`) then the process it is run from will never terminate. This is because the NAPI Worker Pool Atlaspack V3 uses is never terminated.

## Changes

Adds a new feature flag `atlaspackV3CleanShutdown`, when enabled will add a `disposable` to the Atlaspack API so that when Atlaspack is destroyed the NAPI Worker Pool will also be destroyed.

A new `end()` method is added to the `AtlaspackV3` class to support this.

**Note** it's not currently feasible to test this change because we run Mocha with `exit: true` to force termination so this issue cannot be reproduced in an integration test. I will separately investigate whether we can remove that force exit as the underlying issue is long fixed. 

## Checklist

- [x] There is a changeset for this change, or one is not required **(n/a)**
- [x] Added documentation for any new features to the `docs/` folder **(n/a)**


mattcompiles (approved):

matt-koko (approved):

marcins:
macOS tests seem to be hanging (completed in ~43m according to logs, test timed out after 45m).. still investigating, doesn't appear to be the suite I added (that's V3 anyway)

- [ ] fix hanging CI??

PR #663, opened 2025-07-01
Remove exit:true for Mocha tests

marcins:
## Motivation

We've had Mocha doing a force `exit` at the end of tests when they "hang". This potentially masks issues where resources (such as worker pools) are not properly cleaned up. I was able to reproduce a "hang" when using Atlaspack V3 via the `Atlaspack` JS API that wasn't reproducible in a test because of this.

The reason for having that `exit: true` dates back to an issue from 2019 or earlier which is long resolved.

## Changes

* Removed the "exit: true" from all the Mocha configs, and restored the `after()` hook to dispose of shared test worker pools in test-utils.
* Added a `ATLASPACK_MOCHA_HANG_DEBUG` env var for tests that can help to debug hangs by setting up https://github.com/mafintosh/why-is-node-running and allowing you to dump out what handles are still open
* Fixed several unit and integration tests which were keeping open handles and stopping the process from exit'ing

## Checklist

### DIFF
--- a/.changeset/fair-insects-sell.md
+++ b/.changeset/fair-insects-sell.md
@@ -1,0 +1,7 @@
+---
+'@atlaspack/test-utils': patch
+---
+
+- Add ATLASPACK_MOCHA_HANG_DEBUG for debugging hanging tests
+- Ensure at least one test always runs to clean up worker farms
+- Cleanup AtlaspackV3 worker farm
--- a/packages/core/integration-tests/test/react-refresh.js
+++ b/packages/core/integration-tests/test/react-refresh.js
@@ -37,12 +37,7 @@ if (MessageChannel) {
         '/integration/react-refresh-automatic',
       );
 
-      let b,
-        root,
-        randoms,
-        subscription,
-        ports,
-        window = {};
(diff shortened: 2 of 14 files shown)


packages/core/integration-tests/test/react-refresh.js
```diff
@@ -37,14 +37,10 @@ if (MessageChannel) {
         '/integration/react-refresh-automatic',
       );
 
-      let b,
```
marcins, line 40: This didn't actually do what the author probably intended.. I think they were intending

`let {root, randoms, window} = {}` - which really is the same as just `let root, randoms, window;`

Instead this assigns an empty object to `window`, but leaves all the other variables `undefined`.

They can just all be undefined as they're assigned in `beforeEach` anyway..

.changeset/fair-insects-sell.md
```diff
@@ -0,0 +1,7 @@
+---
+'@atlaspack/test-utils': patch
```
mattcompiles, line 2: It doesn't hurt, but FYI you don't need to do changesets for private packages.

mattcompiles (approved):
Nice one bud 👏

PR #674, opened 2025-07-04
Fix @atlaspack/eslint-plugin ts errors

liamqma:
## Motivation

Fix the ts errors which occurred in https://github.com/atlassian-labs/atlaspack/actions/runs/16063496636/job/45334282985#step:6:62

## Changes

I will explain details in the code diff.

[no-changeset]


packages/core/eslint-plugin/package.json
```diff
@@ -9,6 +9,9 @@
     "type": "git",
     "url": "https://github.com/atlassian-labs/atlaspack.git"
   },
+  "scripts": {
```
liamqma, line 12: Add check-ts to the script here so it can be checked in branch build.

packages/core/eslint-plugin/package.json
```diff
@@ -19,6 +22,7 @@
     "@typescript-eslint/utils": "^8.31.1"
   },
   "devDependencies": {
+    "@typescript-eslint/types": "^8.31.1",
```
liamqma, line 25: Add `"@typescript-eslint/types"` as devDep because it's used at https://github.com/atlassian-labs/atlaspack/blob/main/packages/core/eslint-plugin/src/rules/importcond-type-annotations/index.ts#L5

alshdavid (approved):

PR #676, opened 2025-07-04
"atlaspack" npm barrel package

alshdavid:
## Motivation

Single package referencing all of the Atlaspack packages as dependencies and re-exporting them from subpaths.

This is to simplify distribution and consumption of the bundler. 

Basically one big [package.json](https://github.com/atlassian-labs/atlaspack/pull/676/files#diff-9e85083752a1774d53a12e05ff28ed90704a4a7532f7f499aec84c27df1a0ed1R142) with all of the Atlaspack dependencies, then lots of `index.js` files that basically do `export * from '@atlaspack/fs'`

## Changes

- added `atlaspack` npm package
  - depends on all public `@atlaspack/*` packages and re-exports them

## Usage

```bash
npm install atlaspack
npx atlaspack --version

node -e "console.log(require('atlaspack'))" # core
node -e "console.log(require('atlaspack/fs'))"
node -e "console.log(require('atlaspack/core'))"

# Original imports work too via transient dependencies
node -e "console.log(require('@atlaspack/fs'))"
node -e "console.log(require('@atlaspack/core'))"
```

### `.parcelrc`

This works with `.parcelrc` files both directly 

```json
{
  "extends": "atlaspack/config/default",
  "transformers": {
    "*.{js,mjs,jsm,jsx,es6,cjs,ts,tsx}": [
      "atlaspack/transformer/js",
      "..."
    ]
  }
}
```

and transiently

```json
{
  "extends": "@atlaspack/config-default",
  "transformers": {
    "*.{js,mjs,jsm,jsx,es6,cjs,ts,tsx}": [
      "@atlaspack/transformer-js",
      "..."
    ]
  }
}
```

# Install times

Without lockfile: 76s
With lockfile and packages not cached: 19s
With lockfile and packages cached: 1s

_Note: This can be brought down substantially by bundling or vendoring, which me may do in future_

# Dev Release

Currently available for testing under

```
npm install atlaspack@unified
```

## Checklist

[no-changeset]: Nothing is modified


packages/core/eslint-plugin/src/rules/importcond-type-annotations/index.ts
```diff
@@ -90,8 +92,10 @@ const rule = createRule<Options, keyof typeof messages>({
             });
           } else if (
             typeArguments.params.length !== 2 ||
+            // @ts-ignore Argument of type 'TypeNode' is not assignable to parameter of type 'Node | Token | undefined'.
```
liamqma, line 95: I've fixed this at [PR](https://github.com/atlassian-labs/atlaspack/pull/674)

alshdavid: Thanks

.changeset/config.json
```diff
@@ -14,6 +14,118 @@
       "@atlaspack/apvm-macos-arm64"
     ]
   ],
+  "linked": [
```
yamadapc, line 17: Did you mean to fix them?

alshdavid: I think this should be linked because I don't want to bump all the packages in unison (at least not in this PR). 

[https://github.com/changesets/changesets/blob/main/docs/linked-packages.md#example-with-dependants](https://github.com/changesets/changesets/blob/main/docs/linked-packages.md#example-with-dependants)

yamadapc (approved):

PR #684, opened 2025-07-08
Change USE_LIB to USE_SRC

liamqma:
## Motivation

There have been instances where src/ and lib/ were incorrectly used interchangeably. This PR reverses the logic of ATLASPACK_REGISTER_USE_SRC, so that lib/ is now used by default. The src/ directory should only be used when ATLASPACK_REGISTER_USE_SRC is explicitly specified.

We anticipate that src/ will be used in the following scenarios:

- Unit tests
- Integration tests executed locally
- self build
- gulp

In all other cases, lib/ should be used. Please let me know if there are any exceptions to this rule.

## Changes

- Renamed ATLASPACK_REGISTER_USE_LIB to ATLASPACK_REGISTER_USE_SRC
- Specified ATLASPACK_REGISTER_USE_SRC for unit tests and local integration tests

## Checklist

### DIFF
--- a/packages/core/package-manager/src/NodePackageManager.js
+++ b/packages/core/package-manager/src/NodePackageManager.js
@@ -41,14 +41,10 @@ import {transformSync} from '@swc/core';
 // Package.json fields. Must match package_json.rs.
 const MAIN = 1 << 0;
 const SOURCE = 1 << 2;
-const additionalEntries =
-  process.env.ATLASPACK_BUILD_ENV !== 'production' ||
-  process.env.ATLASPACK_SELF_BUILD
-    ? SOURCE
-    : 0;
-const mainWithAdditional = MAIN | additionalEntries;
-const ENTRIES =
-  process.env.ATLASPACK_REGISTER_USE_SRC === 'true' ? mainWithAdditional : MAIN;
+let ENTRIES = MAIN;
+if (process.env.ATLASPACK_REGISTER_USE_SRC === 'true') {
+  ENTRIES |= SOURCE;
+}
 
 const NODE_MODULES = `${path.sep}node_modules${path.sep}`;
 
--- a/packages/dev/repl/src/atlaspack/BrowserPackageManager.js
+++ b/packages/dev/repl/src/atlaspack/BrowserPackageManager.js
@@ -75,14 +75,10 @@ export const BUILTINS = {
 // Package.json fields. Must match package_json.rs.
 const MAIN = 1 << 0;
 const SOURCE = 1 << 2;
-const additionalEntries =
-  process.env.ATLASPACK_BUILD_ENV !== 'production' ||
-  process.env.ATLASPACK_SELF_BUILD
-    ? SOURCE
-    : 0;
-const mainWithAdditional = MAIN | additionalEntries;
-const ENTRIES =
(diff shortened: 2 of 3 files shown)


packages/core/package-manager/src/NodePackageManager.js
```diff
@@ -41,14 +41,14 @@ import {transformSync} from '@swc/core';
 // Package.json fields. Must match package_json.rs.
 const MAIN = 1 << 0;
 const SOURCE = 1 << 2;
+const additionalEntries =
+  process.env.ATLASPACK_BUILD_ENV !== 'production' ||
+  process.env.ATLASPACK_SELF_BUILD
+    ? SOURCE
+    : 0;
+const mainWithAdditional = MAIN | additionalEntries;
```
mattcompiles, line 49: This reads very confusing. From what I can see `additionalEntries` is either source or nothing. Maybe something like the following would be easier to read?

```
let ENTRIES = MAIN;

if (process.env.ATLASPACK_REGISTER_USE_SRC === 'true' ) {
  if (process.env.ATLASPACK_BUILD_ENV !== 'production' || process.env.ATLASPACK_SELF_BUILD) {
    ENTRIES |= SOURCE
  }
}
```

packages/core/package-manager/src/NodePackageManager.js
```diff
@@ -41,14 +41,14 @@ import {transformSync} from '@swc/core';
 // Package.json fields. Must match package_json.rs.
 const MAIN = 1 << 0;
 const SOURCE = 1 << 2;
+const additionalEntries =
+  process.env.ATLASPACK_BUILD_ENV !== 'production' ||
+  process.env.ATLASPACK_SELF_BUILD
```
mattcompiles, line 46: Do we still need `ATLASPACK_BUILD_ENV` &&  `ATLASPACK_SELF_BUILD` here?

liamqma: Yea, good point! I am confident `ATLASPACK_SELF_BUILD` is not needed. Not sure about `ATLASPACK_BUILD_ENV `. I will look into it.

alshdavid: Probably not needed but messing with the build system is like threading a needle so I'd hesitate to tidy those up in this PR in case they unintentionally cause issues.

I observed that if the self build is broken, the integration tests will still pass so we have to tread lightly when making changes

packages/core/package-manager/src/NodePackageManager.js
```diff
@@ -41,14 +41,10 @@ import {transformSync} from '@swc/core';
...
 const SOURCE = 1 << 2;
-const ENTRIES =
-  process.env.ATLASPACK_REGISTER_USE_LIB === 'true'
-    ? MAIN
-    : MAIN |
-      (process.env.ATLASPACK_BUILD_ENV !== 'production' ||
-      process.env.ATLASPACK_SELF_BUILD
-        ? SOURCE
-        : 0);
+let ENTRIES = MAIN;
```
liamqma, line 44: @alshdavid 

To streamline the logic, I've removed the checks for process.env.ATLASPACK_BUILD_ENV and process.env.ATLASPACK_SELF_BUILD. I've ensured that ATLASPACK_REGISTER_USE_SRC=true is manually set wherever necessary.

Risk: There is a possibility that I might have overlooked something.

PR #685, opened 2025-07-09
Fix star re-export of empty files causing runtime errors

mattcompiles:
## Motivation

This PR fixes an issue where star re-exports of empty files (usually occurring in compiled typescript libraries) could cause exports to be undefined at runtime. This can occur as empty files are treated as CJS which potentially can have their exports mutated from outside sources and therefore get marked as potentially containing the superset of the exports for the library. 

In the past we have worked around this by [only performing this behaviour for Assets that are marked as having side effects](https://github.com/parcel-bundler/parcel/pull/9079). However in NPM packages (older versions of `@apollo/client` for example) this value is out of our control. 

The solution we've gone with here is to remove the `*` wildcard export from these empty files, when all incoming deps are ESM type, meaning export mutation shouldn't occur. 

The fix is behind the feature-flag `emptyFileStarRexportFix`.

## Checklist

### DIFF
--- a/.changeset/gorgeous-teachers-reflect.md
+++ b/.changeset/gorgeous-teachers-reflect.md
@@ -1,0 +1,8 @@
+---
+'@atlaspack/feature-flags': patch
+'@atlaspack/core': patch
+'@atlaspack/rust': patch
+---
+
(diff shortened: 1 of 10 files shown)


marcins (approved):
Nice one! Seems good to me.

PR #686, opened 2025-07-09
Fix incorrect marking of computed property access as unused

benjervis:
## Motivation

We had a report of a bug where a variable was coming up as undefined. The code in question looked like:
```ts
import { SOURCES_CONFIG } from './config';

export const Component = ({ source }) => {
  const config = SOURCES_CONFIG[source];
  // ...
```

It turns out that the Collector for symbol prop only looks at MemberExpressions (accessing a property on an object) to consider whether they're from a require (` const config = require('./config'); config.SOURCES_CONFIG`), but otherwise ignores them. This means that the imported object is not being marked as used, and the bundler removes the asset.

## Changes

Seems like this was just a logic error in the series of `if` blocks that make up the Ident block in MemberExpr.

To make it easier to follow, I broke the chain of if-else blocks into separate ifs with returns.

Previously, the logic was that if it was a computed property, the property would get added to non-static access, and it would never add it to the used imports field.

Now, with the feature flag, the used imports check always happens, and if it is used then it can also be added to non-static access.

## Checklist


packages/core/feature-flags/src/index.js
```diff
@@ -42,6 +42,7 @@ export const DEFAULT_FEATURE_FLAGS: FeatureFlags = {
   inlineConstOptimisationFix: false,
   hmrImprovements: false,
   atlaspackV3CleanShutdown: false,
+  unusedComputedPropertyFix: false,
```
mattcompiles, line 45: Let's default to true for the tests.
`unusedComputedPropertyFix: process.env.NODE_ENV === 'test'`

packages/transformers/js/core/src/hoist.rs
```diff
@@ -2733,6 +2733,7 @@ mod tests {
         true,
         context.is_module,
         false,
+        false,
```
mattcompiles, line 2736: Make this `true`

packages/transformers/js/core/src/fs.rs
```diff
@@ -443,6 +445,7 @@ mod tests {
         &mut deps,
         true,
         false,
+        false,
```
mattcompiles, line 448: Make this true

mattcompiles (approved):

PR #721, opened 2025-07-23
Add support for bundle merging based on webpackChunkName comments

benjervis:
## Motivation

As part of the ongoing effort to bring Atlaspack-built artifacts up to performance parity with webpack, we've identified that there are a number of imports that webpack merges using the `webpackChunkName` comments, that aren't being merged in Atlaspack.

To bring them into closer alignment, we want to support those comments to merge async bundles in Atlaspack

## Changes

The Rust side of the JS Transformer already extracts "magic comments", so we can use that existing information to update the bundling algorithm to merge bundles pointed to by dependencies that have the same magic comment.

This did also require some changes to the packager, as we've had to support there being more than one entry asset.

I've made the integration test v2 for now, I'll work on adding this to the native side in a follow up, keen to get this into canary for now.

## Checklist


packages/bundlers/default/src/decorateLegacyGraph.js
```diff
@@ -52,26 +54,41 @@ export function decorateLegacyGraph(
...
         }),
       );
 
-      bundleGraph.addBundleToBundleGroup(bundle, bundleGroup);
+      if (getFeatureFlag('supportWebpackChunkName')) {
+        for (let bundleGroup of bundleGroups.values()) {
+          bundleGraph.addBundleToBundleGroup(bundle, bundleGroup);
+        }
+      } else {
+        invariant(bundleGroup);
```
marcins, line 89: /nit does this invariant need to be here? it's already asserted on line 69. (Doesn't hurt anything..)

packages/bundlers/default/src/idealGraph.js
```diff
@@ -1296,12 +1382,25 @@ export function createIdealGraph(
...
-    );
-    bundleToKeep.internalizedAssets.union(bundleToRemove.internalizedAssets);
+    if (getFeatureFlag('supportWebpackChunkName')) {
+      if (bundleToKeep.internalizedAssets != null) {
+        if (bundleToRemove.internalizedAssets != null) {
+          bundleToKeep.internalizedAssets.intersect(
+            bundleToRemove.internalizedAssets,
+          );
+        } else {
+          bundleToKeep.internalizedAssets.clear();
```
marcins, line 1392: How does this work? (I guess it requires more knowledge of internalization)

If the `bundleToKeep` has `internalizedAssets` they'll be removed if the bundle being removed doesn't have `internalizedAssets`?

packages/core/core/src/public/MutableBundleGraph.js
```diff
@@ -229,6 +230,17 @@ export default class MutableBundleGraph
       isPlaceholder = entryAssetNode.requested === false;
     }
 
+    let entryAssetIds = [];
+    if (entryAsset) {
```
marcins, line 234: Is this the same as `opts.entryAsset`? some code specific to that was removed on line 269.

marcins (approved):
Seems good to me!

I don't have enough deep knowledge of the bundling algo to understand the changes in context so have mostly reviewed just the logic that's there and whether it seems to make conceptual sense.

packages/bundlers/default/src/decorateLegacyGraph.js
benjervis: For logical reasons, no, but Flow isn't smart enough to work out that it's already been invariantated unfortunately

packages/core/core/src/public/MutableBundleGraph.js
benjervis: It is, it's just that it needs converting on input.

See: https://github.com/atlassian-labs/atlaspack/blob/8c776a1d4a1b28dbb5aed6e9ded46023c8f46196/packages/core/core/src/public/MutableBundleGraph.js#L200-L202

PR #723, opened 2025-07-25
adding projectRoot and gitRoot options to InitialAtlaspackOptions

vykimnguyen:
## Motivation

([AFB-1408](https://product-fabric.atlassian.net/browse/AFB-1408)) This PR allows `projectRoot` and `gitRoot` to be specified for initial options.

## Changes

- Adds `projectRoot` and `gitRoot` to `initialAtlaspackOptions`
- Modify cli to allow specifying of these options
- Defaults back to auto detection of lockfile to determine root if these aren't specified
- Integration tests with new options

## Checklist


marcins:
Looks like there's still some flow check errors, as well as one of the integration tests that's not behaving as expected.

packages/core/cli/src/normalizeOptions.js
```diff
@@ -37,6 +37,8 @@ export interface Options {
   trace?: boolean;
   cache?: boolean;
   cacheDir?: string;
+  projectRoot?: string;
+  gitRoot?: string;
```
pancaspe87, line 41: IIRC we also have `gitRepoPath` which should be specific to VCS. Worth checking if anything should be consolidated now that we making the option available in `InitialAtlaspackOptions`

cc @yamadapc

yamadapc: Passing this is a bit weird because this should be easy to resolve on our own

vykimnguyen: I'll remove the option for `gitRoot` as discussed in standup

packages/core/core/src/resolveOptions.js
```diff
@@ -103,22 +103,40 @@ export default async function resolveOptions(
...
-      [...LOCK_FILE_NAMES, '.git', '.hg'],
-      path.parse(entryRoot).root,
-    )) || path.join(inputCwd, 'index'); // ? Should this just be rootDir
+  let projectRoot;
+  if (initialOptions.projectRoot != null) {
+    // Use explicitly provided projectRoot, resolving relative paths to cwd
+    const providedProjectRoot = initialOptions.projectRoot;
+    projectRoot = path.isAbsolute(providedProjectRoot)
+      ? providedProjectRoot
+      : path.resolve(inputCwd, providedProjectRoot);
```
mattcompiles, line 112: I think we should enforce this path to be absolute. CWD can potentially change between calls which is definitely not what you want. I think if you pass projectRoot it should be an error if it's not absolute.

.changeset/healthy-humans-breathe.md
```diff
@@ -0,0 +1,8 @@
+---
+'@atlaspack/integration-tests': minor
+'@atlaspack/types-internal': minor
+'@atlaspack/core': minor
+'@atlaspack/cli': minor
+---
+
+adding projectRoot and gitRoot options
```
marcins, line 8: /nit the changelog still mentions `gitRoot`.

PR #724, opened 2025-07-28
Fix Node builtin polyfilling for libraries

benjervis:
## Motivation

Under the particular set of circumstances that:
- We're building a library (`isLibrary: true`)
- We have a non-global output format (`commonjs`)
- We want to bundle node modules (`includeNodeModules: true`)
- We're targeting a browser-like environment (`web-worker`)

The node resolver would incorrectly insert require statements for these node builtins, instead of polyfilling them.

This may seem like a niche case, but it's the recommended setup for targeting SSR environments.

This is caused by the fact that the `shouldIncludeNodeModule` that determines whether a builtin is external or not kinda just doesn't consider that `includeNodeModules` might be "`true`".

## Changes

This change modifies the first check of the function, whether `includeNodeModules` is false, to just whether it's a boolean, and returns that if so.

As ever, locked behind a feature flag, `libraryBuiltinsFix`.

## Checklist

### DIFF
--- a/.changeset/bright-badgers-behave.md
+++ b/.changeset/bright-badgers-behave.md
@@ -1,0 +1,14 @@
+---
+'@atlaspack/node-resolver-core': patch
+'@atlaspack/feature-flags': patch
+---
+
+Fixes the handling of node builtins in library mode.
+
+Previously, the function to check whether a node builtin should be external didn't account
+for the case where `includeNodeModules` was set to just `true`.
+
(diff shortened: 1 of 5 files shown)


liamqma (approved):

benjervis:
Seems that this is deliberate, in that library builds want to exclude Node builtins even when they include node_modules. While that's not what I want in this case, it can be fixed with aliasing.

In future, when we look to support a Tesseract target, this should all happen internally.

PR #727, opened 2025-07-30
Improve macOS integration test performance

marcins:
## Motivation

macOS integration tests seem to be getting slower again. [Profiling](https://github.com/atlassian-labs/atlaspack/pull/603) revealed two large hot spots both originating from the Cache tests:
* https://github.com/atlassian-labs/atlaspack/blob/main/packages/core/test-utils/src/utils.ts#L1327-L1381
* writing snapshots in the watcher

## Changes

This PR changes the watcher backend to `fs-events` on macOS. This reduces the CI time by approx 4-5 min (from ~30min to ~26min) based on two test runs.

~~I did also try and optimise the `assertNoFilePathInCache` but the few things I tried weren't any faster (using a chunked faster string search algo, scanning in a promise queue, using a Rust based function for searching for the string in the file).~~

`assertNoFilePathInCache` was optimised to use a promise queue to check files - testing with Hyperfine locally showed the best value for concurrency was 50, this reduces macOS test time by another ~5 min and Linux by ~2 min.

## Checklist

[no-changeset]: private packages only

### DIFF
--- a/packages/core/integration-tests/test/cache.ts
+++ b/packages/core/integration-tests/test/cache.ts
@@ -40,6 +40,7 @@ import {deserialize} from '@atlaspack/build-cache';
 import {hashString} from '@atlaspack/rust';
 import {getAllEnvironments} from '@atlaspack/rust';
 import type {FeatureFlags} from '@atlaspack/feature-flags';
+import os from 'os';
 
 let inputDir: string;
 let packageManager = new NodePackageManager(inputFS, '/');
@@ -56,6 +57,7 @@ let packageManager = new NodePackageManager(inputFS, '/');
       {
         inputFS: overlayFS,
         shouldDisableCache: false,
+        watchBackend: os.platform() === 'darwin' ? 'fs-events' : undefined,
         featureFlags: {
           ...featureFlags,
           cachePerformanceImprovements,
--- a/packages/core/test-utils/package.json
+++ b/packages/core/test-utils/package.json
@@ -24,6 +24,7 @@
     "@atlaspack/fs": "2.15.16",
     "@atlaspack/package-manager": "2.14.21",
     "@atlaspack/utils": "2.17.3",
+    "async": "^3.2.6",
     "chalk": "^4.1.0",
     "expect": "^29.7.0",
     "ncp": "^2.0.0",
@@ -33,6 +34,7 @@
     "resolve": "^1.12.0",
     "rimraf": "^5.0.5",
     "ws": "^7.0.0",
+    "@types/async": "^3.2.6",
     "tempy": "^1.0.0",
     "why-is-node-running": "^3.2.2"
   }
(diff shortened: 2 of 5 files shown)


packages/core/test-utils/package.json
```diff
@@ -24,14 +24,17 @@
...
     "@atlaspack/utils": "2.17.3",
+    "async": "^3.2.6",
     "chalk": "^4.1.0",
     "expect": "^29.7.0",
     "ncp": "^2.0.0",
     "nullthrows": "^1.1.1",
     "posthtml": "^0.16.5",
     "posthtml-parser": "^0.10.1",
     "resolve": "^1.12.0",
+    "rimraf": "^5.0.5",
```
marcins, line 35: I still don't get why I have to add these as half the time running `yarn add` in a workspace will throw an error about `rimraf` being required but not being present in the workspace.

packages/dev/mocha-profiler/package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Profile mocha tests",
   "license": "(MIT OR Apache-2.0)",
   "private": true,
-  "main": "src/index.js",
+  "main": "./lib/index.js",
```
marcins, line 7: Fixes running the Mocha profiler (via `ATLASPACK_PROFILE_MOCHA=true`) with TypeScript.

packages/core/integration-tests/test/cache.ts
```diff
@@ -56,6 +57,7 @@ let packageManager = new NodePackageManager(inputFS, '/');
       {
         inputFS: overlayFS,
         shouldDisableCache: false,
+        watchBackend: os.platform() === 'darwin' ? 'fs-events' : undefined,
```
mattcompiles, line 60: Isn't `fs-events` the default on mac anyway?

packages/core/test-utils/src/utils.ts
```diff
@@ -1328,6 +1330,52 @@ export async function assertNoFilePathInCache(
   fs: FileSystem,
   dir: string,
   projectRoot: string,
+) {
+  // For debugging purposes, log all instances of the projectRoot in the cache.
+  // Otherwise, fail the test if one is found.
+  if (process.env.ATLASPACK_DEBUG_CACHE_FILEPATH != null)
+    return assertNoFilePathInCacheDebug(fs, dir, projectRoot);
```
mattcompiles, line 1337: No brace hey? How fancy of you

mattcompiles (approved):
Nice one mate

packages/core/test-utils/src/utils.ts
marcins: It's possibly Cursor's fault and I didn't notice.. if you don't like it, lint for it 🧌

PR #730, opened 2025-08-01
APVM support via unified package

alshdavid:
## Motivation

Enable APVM support with the existing unified `atlaspack` npm package: https://www.npmjs.com/package/atlaspack

## Changes

- APVM changes
  - Download/extract the `atlaspack` npm package
  - Link the `atlaspack` package
  - Support for resolving npm tags (like `latest` `dev`) 
- Added type re-exports to core packages in the `atlaspack` npm package 

#### Atlaspack unmodified

The `atlaspack` package distributes Atlaspack completely unmodified so there are no consumer side changes required.

#### yarn install

APVM will download and do a `yarn install` on the `atlaspack` package. This works fine but is a bit slow (~45s). 

This can be skipped with: 

```
apvm install --skip-postinstall latest
```

#### Reuse `node_modules` at link target

If the consumer already has a version of Atlaspack installed and you want to reuse their `node_modules`, use the `--exclude-node-modules` option:

```
apvm link --exclude-node-modules 2.0.6
```

#### Optimised install times

If a project already has Atlaspack installed and you want to optimize the install times, run:

```
apvm install --skip-postinstall 2.0.6
apvm link --exclude-node-modules 2.0.6
```

## Checklist

### DIFF
--- a/crates/apvm/src/cmd/install.rs
+++ b/crates/apvm/src/cmd/install.rs
@@ -4,7 +4,7 @@ use std::time::SystemTime;
 use clap::Parser;
 
 use crate::cmd::install_git::install_from_git;
-use crate::cmd::install_npm::install_from_npm;
+use crate::cmd::install_npm::{install_from_npm, resolve_from_npm};
 use crate::cmd::install_release::install_from_release;
 use crate::context::Context;
 use crate::platform::package::ManagedPackage;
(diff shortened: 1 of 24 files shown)


crates/apvm/src/cmd/install.rs
```diff
@@ -33,7 +33,17 @@ pub struct InstallCommand {
 pub fn main(ctx: Context, cmd: InstallCommand) -> anyhow::Result<()> {
   let start_time = SystemTime::now();
 
-  let specifier = ctx.resolver.resolve_specifier(&cmd.version)?;
+  let specifier_raw = ctx.resolver.resolve_specifier(&cmd.version)?;
+
+  let specifier = match &specifier_raw {
+    Specifier::Npm { version } => Specifier::Npm {
+      version: resolve_from_npm(&ctx, version)?,
```
alshdavid, line 40: Enables resolving version specifiers like `latest` or `dev`

crates/apvm/src/cmd/install_npm.rs
```diff
@@ -67,5 +77,37 @@ pub fn install_from_npm(
...
+    return Ok(());
+  }
+
+  // If the git sha is specified, use the yarn.lock from the repo
+  if fs::exists(pkg.contents().join("GIT_SHA"))? {
+    let git_sha = fs::read_to_string(pkg.contents().join("GIT_SHA"))?;
+    let url_lock_file = format!("{}/{}/{}", c::GITHUB_RAW, git_sha, "yarn.lock");
+    let bytes = http::download_bytes(url_lock_file)?;
+    fs::write(pkg.contents().join("yarn.lock"), bytes)?;
+  }
```
alshdavid, line 91: For faster `yarn install` times. It looks like the lock file is only partially used, however

packages/core/conditional-import-types/index.d.ts
```diff
@@ -41,3 +41,5 @@ declare function importCond<CondT, CondF>(
     ? ConditionalImport<CondT, CondF>
     : never
   : never;
+
+export {};
```
alshdavid, line 45: This isn't a module without any exports

yamadapc (approved):

PR #731, opened 2025-08-05
Add support for isolating inline scripts

marcins:
## Motivation

In some use cases we want to bundle multiple HTML entries with inline scripts that depend on shared code, however we don't want Atlaspack to produce a shared bundle - we want to keep all the bundled code within the inline script.

Essentially we want the inline script to also be isolated, something that is not previously supported by Atlaspack.

## Changes

Implement a new attribute to tell Atlaspack that an inline script should be isolated: `<script data-atlaspack-isolated>`. With this attribute, the script will use a new bundle behaviour called `inlineIsolated` which is a combination of both inline and isolated.

The majority of the change is adding conditions for this new `bundleBehavior` everywhere there's a check for `inline` or `isolated` to ensure the new `bundleBehavior` is both.

There is a new feature flag `inlineIsolatedScripts` which enables parsing the new attribute in the HTML transformer (both in JS and Rust) and setting the new `bundleBehavior` on the dependency, however there is no feature flag checks elsewhere in the code - the new `bundleBehavior` will only ever be set by the HTML transformers behind gated code, so I believe this is a safe approach.

## Checklist

- [x] Added documentation for any new features to the `docs/` folder (TODO)

### DIFF
--- a/crates/atlaspack_plugin_transformer_html/src/html_transformer.rs
+++ b/crates/atlaspack_plugin_transformer_html/src/html_transformer.rs
@@ -29,7 +29,10 @@ impl AtlaspackHtmlTransformerPlugin {
   pub fn new(ctx: &PluginContext) -> Self {
     AtlaspackHtmlTransformerPlugin {
       project_root: ctx.options.project_root.clone(),
-      enable_inline_isolated: ctx.options.feature_flags.bool_enabled("inlineIsolated"),
+      enable_inline_isolated: ctx
+        .options
+        .feature_flags
+        .bool_enabled("inlineIsolatedScripts"),
     }
   }
 }
--- a/crates/atlaspack_plugin_transformer_html/src/html_transformer.rs
+++ b/crates/atlaspack_plugin_transformer_html/src/html_transformer.rs
@@ -22,12 +22,14 @@ use crate::html_dependencies_visitor::HtmlDependenciesVisitor;
 #[derive(Debug)]
 pub struct AtlaspackHtmlTransformerPlugin {
   project_root: PathBuf,
+  enabled_inline_isolated: bool,
 }
 
 impl AtlaspackHtmlTransformerPlugin {
   pub fn new(ctx: &PluginContext) -> Self {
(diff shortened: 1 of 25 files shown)


crates/atlaspack_plugin_transformer_html/src/html_transformer.rs
```diff
@@ -22,12 +22,14 @@ use crate::html_dependencies_visitor::HtmlDependenciesVisitor;
 #[derive(Debug)]
 pub struct AtlaspackHtmlTransformerPlugin {
   project_root: PathBuf,
+  enabled_inline_isolated: bool,
```
marcins, line 25: I should rename this to `enable_` for consistency.

packages/bundlers/default/src/bundleMerge.ts
```diff
@@ -12,8 +12,13 @@ function getBundlesForBundleGroup(
 ): number {
   let count = 0;
   bundleGraph.traverse((nodeId) => {
-    // @ts-expect-error TS2339
-    if (bundleGraph.getNode(nodeId)?.bundleBehavior !== 'inline') {
+    const node = bundleGraph.getNode(nodeId);
```
marcins, line 15: Extracted this and added some more checks to fix a TS error.

packages/transformers/html/src/inline.ts
```diff
@@ -120,6 +126,14 @@ export default function extractInlineAssets(
           delete node.attrs.type;
         }
 
+        let bundleBehavior: BundleBehavior = 'inline';
+        if (
+          getFeatureFlag('inlineIsolatedScripts') &&
+          typeof node.attrs['data-atlaspack-isolated'] !== 'undefined'
+        ) {
+          bundleBehavior = 'inlineIsolated';
```
marcins, line 134: Should remove the attribute here and check that in the integration test for consistency with the rust impl.

mattcompiles (approved):
Look good mate. Definitely feels like we should remodel this at some point to have separate states for inline and isolated (e.g. `bundle.isInline`) but this is fine for now.

PR #732, opened 2025-08-06
Add tesseract context

vykimnguyen:
## Motivation

Adding new `tesseract` context target in Atlaspack. This is initially a fork of the `web-worker` context.

## Changes

- Added `tesseract` to `EnvironmentContext`
- Added tests for new `tesseract` context

## Checklist


packages/core/core/src/atlaspack-v3/worker/compat/environment.ts
```diff
@@ -66,13 +66,21 @@ export class Environment implements IEnvironment {
   }
 
   isWorker(): boolean {
-    return this.context === 'web-worker' || this.context === 'service-worker';
+    return (
+      this.context === 'web-worker' ||
+      this.context === 'service-worker' ||
+      this.context === 'tesseract'
```
marcins, line 72: Should we instead change all the places in Atlaspack that do `isWorker()` to do `isWorker() || isTesseract()` instead?

.. and when we switch to using the new env in AFM, you'd need to find all the checks for `isWorker()` (if any) in custom plugins there and update the usages.

packages/core/core/src/public/Environment.ts
```diff
@@ -26,11 +26,12 @@ export const BROWSER_ENVS: Set<string> = new Set<string>([
   'web-worker',
   'service-worker',
   'worklet',
+  'tesseract',
   'electron-renderer',
 ]);
 const ELECTRON_ENVS = new Set(['electron-main', 'electron-renderer']);
 const NODE_ENVS = new Set(['node', ...ELECTRON_ENVS]);
-const WORKER_ENVS = new Set(['web-worker', 'service-worker']);
+const WORKER_ENVS = new Set(['web-worker', 'service-worker', 'tesseract']);
```
marcins, line 34: Similar for this, I'd add `tesseract` to `ISOLATED_ENVS` and update checks to also check for `isTesseract` where there's checks for `isWorker`.

marcins (approved):
Just some minor feedback to make more use of `isTesseract` since you'll be able to couple moving to the new env with any consumer usages of `isWorker`.

PR #739, opened 2025-08-07
Add cacheable NX configuration

yamadapc:
This commits fixes cycles and moves to a model where tasks are cacheable by
`nx`. This moves gulp/babel builds to be ran for each package and fixes
depedency declarations.

On the FS package, some windows compatibility code has been removed to fix a
cycle.

This doesn't change the current _build_. What is changed is that declaration
files are used for type-checking.

So when we type-check, `nx` will build dependencies of packages before building
them. So packages will type-check against the type declarations of the package
rather than their source-code.

[no-changeset] Nothing changes

Test Plan: N/A


matt-koko (approved):
LGTM (assuming we can get integration tests passing)

PR #741, opened 2025-08-07
combine FeatureFlags and DEFAULT_FEATURE_FLAGS into one object

matt-koko:
The `FeatureFlags` and `DEFAULT_FEATURE_FLAGS` objects have been merged into one.

Previously we could not do this due to Flow shenaniganry. 

Now that the Flow->Typescript migration has landed, we can improve our feature flag workflow.

## Checklist

### DIFF
--- a/packages/core/feature-flags/src/index.ts
+++ b/packages/core/feature-flags/src/index.ts
@@ -1,158 +1,5 @@
 // Converted from Flow to TypeScript
 
-export type FeatureFlags = {
-  // This feature flag mostly exists to test the feature flag system, and doesn't have any build/runtime effect
-  readonly exampleFeature: boolean;
-  readonly exampleConsistencyCheckFeature: ConsistencyCheckFeatureFlagValue;
-
-  /**
-   * Rust backed requests
-   */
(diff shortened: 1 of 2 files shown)


packages/core/feature-flags/src/index.ts
```diff
@@ -1,199 +1,169 @@
...
-  unusedComputedPropertyFix: process.env.NODE_ENV === 'test',
-  emptyFileStarRexportFix: process.env.NODE_ENV === 'test',
-  cliProgressReportingImprovements: false,
-  supportWebpackChunkName: process.env.NODE_ENV === 'test',
-  condbDevFallbackDev: false,
-  condbDevFallbackProd: false,
-  incrementalBundlingVersioning: process.env.NODE_ENV === 'test',
   inlineIsolatedScripts: process.env.NODE_ENV === 'test',
-};
+} as const;
```
yamadapc, line 164: You want to remove this ; `as const` will make types equal to the initialisation values

yamadapc, line 164:
```suggestion
};
```

PR #742, opened 2025-08-07
Fix missing dependency on tests

yamadapc:
[no-changeset] Build changes

Test Plan: N/A

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -30,6 +30,12 @@
       ],
       "outputs": ["{projectRoot}/lib/**/*.js", "{projectRoot}/bin/**/*.js"],
       "cache": true
+    },
+    "test:integration": {
+      "dependsOn": ["^check-ts"]
+    },
+    "test:integration:ci": {
+      "dependsOn": ["^build:lib", "^check-ts"]
     }
   }
 }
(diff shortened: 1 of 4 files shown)


nx.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/nrwl/nx/refs/heads/master/packages/nx/schemas/nx-schema.json",
-  "parallel": true,
+  "parallel": "200%",
```
mattcompiles, line 3: Why 200%?

.changeset/angry-bears-pull.md
```diff
@@ -0,0 +1,136 @@
...
+'@atlaspack/cache': patch
+'@atlaspack/graph': patch
+'@atlaspack/types': patch
+'@atlaspack/utils': patch
+'@atlaspack/core': patch
+'@atlaspack/rust': patch
+'@atlaspack/query': patch
+'@atlaspack/cli': patch
+'@atlaspack/fs': patch
+'@workspace/scripts': patch
```
mattcompiles, line 133: Why are we bumping every package manually?

nx.json
yamadapc: I was experimenting with a few values and this seemed ok ; we should tune it

.changeset/angry-bears-pull.md
yamadapc: There were a lot of dependency fixes

PR #760, opened 2025-08-20
Fix some more random `ts-expect-errors`

marcins:
## Motivation

We have many ts-expect-errors, we should have fewer.

## Changes

Have fewer ts-expect-errors.

## Checklist

### DIFF
--- a/packages/core/feature-flags/src/index.ts
+++ b/packages/core/feature-flags/src/index.ts
@@ -207,8 +207,7 @@ export function runWithConsistencyCheck<Result, CustomDiagnostic>(
   ) => void,
 ): Result {
   const value = featureFlagValues[flag];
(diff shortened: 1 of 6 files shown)


packages/core/feature-flags/src/index.ts
```diff
@@ -207,8 +207,7 @@ export function runWithConsistencyCheck<Result, CustomDiagnostic>(
   ) => void,
 ): Result {
   const value = featureFlagValues[flag];
-  // @ts-expect-error - TypeScript doesn't understand the union type comparison
```
marcins, line 210: @alshdavid The type of `value` is `boolean | string` - so checking both `!value` and `value === false` don't make sense.

Oh, unless it was to check for an empty string? I guess I can explicitly add that as (`!'' === true`).

PR #769, opened 2025-08-29
Support custom `env` properties in targets

marcins:
## Motivation

We have a build where there are multiple entries being built, but a transformer needs to behave differently when it's processing an asset that originates from one of the entries but not the others - and the assets overlap. 

Currently this happens by having each entry built in it's own isolated bundler instance (which works because the output is independent) but for efficiency we'd like to build this all in a single instance.

Atlaspack already supports building the same files multiple times with different Environments - these are designed for things like different browser lists (e.g. "modern" vs "IE11" in the old days), or different contexts like node vs browser.

We however want to build all the entries for, effectively, the same Environment, but we want the assets for one entry to be processed differently. 

We can do this by adding the ability to set custom attributes on the Environment that can be used in custom transformers to change the behaviour based on the attribute.

This change is built on #767 as I also need that feature for this, and make use of it in the integration test, but is independent in terms of implementation.

## Changes

A new feature flag has been added `customEnvInTargets` that when enabled will add support for a `customEnv` property in a target. This `env` is then available in a transformer via `asset.env.customEnv`.

e.g. you can specify targets like this which will build two entries with one config, and one with another: 

```ts
targets: {
          one: {
            source: './input.js',
            context: 'browser',
            distDir: './dist',
            env: {
              MY_ENV: 'one',
            },
          },
          two: {
            source: ['./input2.js', './input3.js'],
            context: 'browser',
            distDir: './dist',
            env: {
              MY_ENV: 'two',
            },
          },
        },
```

## Checklist

### DIFF
--- a/.changeset/lucky-ligers-join.md
+++ b/.changeset/lucky-ligers-join.md
@@ -1,0 +1,5 @@
+---
+'@atlaspack/resolver-tesseract': patch
+---
+
+small fixes for tesseract resolver
--- a/.changeset/rotten-beers-look.md
+++ b/.changeset/rotten-beers-look.md
@@ -1,0 +1,8 @@
+---
+'@atlaspack/types-internal': minor
+'@atlaspack/feature-flags': minor
+'@atlaspack/core': minor
+'@atlaspack/rust': minor
+---
+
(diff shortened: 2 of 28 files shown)


mattcompiles (approved):
Looks good mate

PR #774, opened 2025-09-01
Add export classifier to remove helper on imports when not required

JakeLane:
## Motivation

Currently all or most exports are wrapped in getter functions. This is to support ESM live bindings.

Given code that looks like:

```
export const b = 2;
export let a = 1;
export function x() {
  a += 1;
  return a * b;
}
```

We produce code that looks like

```
const b = 2;
let a = 1;
function x() {
  a += 1;
  return a * b;
}

$export('b', () => b);
$export('a', () => a);
$export('x', () => x);
```

## Changes

This change introduces a visiting phase which has two phases

- The first phase is visiting all ESM export nodes and following the tree to find all identifiers that are exported.
- The second phase is used to find all identifiers that are exported, so we can know the type of export (const, let, var, function, class, etc)

After this, we've collected all the info into a hashmap which can be used in the collect code which informs the ScopeHoistingPackager. There we know whether to use the helper function or not.

### Limitations

I skipped implementing a few types and defaulted to the current behaviour to save effort. These include:

- CJS exports (probably valuable, but I haven't researched these yet)
- Non-identifier exports (when you do something like `export default 'some-value'`). This meant tracking non-identifiers as a type in the HashMap.

## Checklist

### DIFF
--- a/packages/core/core/src/atlaspack-v3/worker/compat/asset-symbols.ts
+++ b/packages/core/core/src/atlaspack-v3/worker/compat/asset-symbols.ts
@@ -170,6 +170,7 @@ export class MutableAssetSymbols implements IMutableAssetSymbols {
             }
           : undefined,
         isEsmExport: Boolean(meta?.isEsm),
+        isStaticBindingSafe: Boolean(meta?.isStaticBindingSafe),
         isWeak: false,
         selfReferenced: Boolean(meta?.selfReferenced),
       });


packages/packagers/js/src/ScopeHoistingPackager.ts
```diff
@@ -1594,21 +1554,33 @@ ${code}
...
-                let get = this.buildFunctionExpression([], resolvedSymbol);
-                let set = asset.meta.hasCJSExports
-                  ? ', ' +
-                    this.buildFunctionExpression(['v'], `${resolvedSymbol} = v`)
-                  : '';
-                prepend += `$parcel$export($${assetId}$exports, ${JSON.stringify(
-                  symbol,
-                )}, ${get}${set});\n`;
-                this.usedHelpers.add('$parcel$export');
-                prependLineCount++;
```
benjervis, line 1611: Also was this meant to be removed?

yamadapc (approved):

mgfalzon:
@JakeLane +1 on separating the TS changes


Can we add unit tests for the new functionality? I didn't see any, but I may have missed it

packages/packagers/js/src/ScopeHoistingPackager.ts
```diff
@@ -1599,16 +1599,29 @@ ${code}
...
+
+                if (
+                  getFeatureFlag('exportsRebindingOptimisation') &&
+                  asset.meta?.isStaticBindingSafe
```
mattcompiles, line 1605: Pretty sure we can remove this block as `asset.meta.isStaticBindingSafe` is never set.

mattcompiles (approved):

packages/packagers/js/src/ScopeHoistingPackager.ts
```diff
@@ -1648,29 +1648,37 @@ ${code}
...
+            exp,
+            undefined,
+            replacements,
+          );
+          const meta = asset.symbols.get(exp)?.meta;
+          if (
+            getFeatureFlag('exportsRebindingOptimisation') &&
+            meta?.isStaticBindingSafe
+          ) {
+            append += `$${assetId}$exports[${JSON.stringify(
```
marcins, line 1665: I assume source map line counts get handled properly for this?

marcins (approved):

packages/packagers/js/src/ScopeHoistingPackager.ts
JakeLane: Yeah I think it should work as we follow the existing behaviour. `prependLines` is used in the sourcemap offset.

PR #775, opened 2025-09-03
Fixes for tesseract resolver

vykimnguyen:
## Motivation

Added fixes for tesseract resolver to successfully build

## Changes
- Removed context environment check to fix module css resolution error
- Fix preResolved map initialisation from config
- Fix analytics client library from being pushed multiple times to ignoreModules
- Add back proxy for `isLibrary` to resolve in tesseract environment

## Checklist

### DIFF
--- a/packages/resolvers/tesseract/src/TesseractResolver.ts
+++ b/packages/resolvers/tesseract/src/TesseractResolver.ts
@@ -11,7 +11,7 @@ interface TesseractResolverConfig {
   browserResolvedNodeBuiltins?: Array<string>;
 
   /** Module mappings that bypass normal resolution (e.g., for SSR compatibility). */
-  preResolved?: Map<string, string>;
+  preResolved?: Record<string, string>;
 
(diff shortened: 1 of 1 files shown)


packages/resolvers/tesseract/src/TesseractResolver.ts
```diff
@@ -191,21 +192,37 @@ export default new Resolver({
       (options.env.STATIC_FALLBACK === 'true' &&
         STATIC_FALLBACK_MODULES.includes(specifier));
 
+    const snapvmEnv = new Proxy(dependency.env, {
```
marcins, line 195: Ideally we wouldn't need this proxy as it's likely not great for performance.. is the issue that the environment needs to behave differently for one of the resolvers and not the other?

Any other way around this?

If we do need to keep this (at least for now), given that it only applies when `useBrowser` is true then we should only create / use the proxy when `useBrowser` is true.

marcins (approved):

PR #779, opened 2025-09-08
Preserve unstableSingleFileOutput in CSS asset env

marcins:
## Motivation

In a transformer that touches CSS files after the default CSS transform, we require access to the `asset.env.unstableSingleFileOutput` property, however this property is not preserved when the CSS transformer normalises the asset environment.

## Changes

Add a new feature flag `preserveUnstableSingleFileOutputInCss` which when enabled will copy the `unstableSingleFileOutput` property to the new environment for the CSS asset.

## Checklist


packages/core/integration-tests/test/integration/css/env-test-transformer.js
marcins, line 1: This is a real file because I was testing the changes with V3, and V3 doesn't like reading plugins from OverlayFS.

PR #780, opened 2025-09-09
Fix v3 exports opt

JakeLane:
## Motivation

There's a type issue with symbols which covered an issue with the v3 behaviour.

## Changes

This fixes the types and makes sure to pass the isStaticBindingSafe property

## Checklist

### DIFF
--- a/packages/core/types-internal/src/index.ts
+++ b/packages/core/types-internal/src/index.ts
@@ -491,9 +491,9 @@ export type Symbol = string;
 export interface AssetSymbols
   extends Iterable<
     [
-      symbol,
+      Symbol,
       {
-        local: symbol;
+        local: Symbol;
         loc: SourceLocation | null | undefined;
         meta?: Meta | null | undefined;
       },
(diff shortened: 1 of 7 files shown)


packages/core/types-internal/src/index.ts
```diff
@@ -1691,19 +1691,19 @@ export interface BundleGraph<TBundle extends Bundle> {
...
-   * <code>asset</code> exports <code>symbol</code>, try to find the asset where the \
+   * <code>asset</code> exports <code>Symbol</code>, try to find the asset where the \
    * corresponding variable lives (resolves re-exports). Stop resolving transitively once \
-   * <code>boundary</code> was left (<code>bundle.hasAsset(asset) === false</code>), then <code>result.symbol</code> is undefined.
+   * <code>boundary</code> was left (<code>bundle.hasAsset(asset) === false</code>), then <code>result.Symbol</code> is undefined.
    */
   getSymbolResolution(
     asset: Asset,
-    symbol: symbol,
+    Symbol: Symbol,
```
mattcompiles, line 1706: This should be `symbol: Symbol,`

PR #781, opened 2025-09-09
Add attribution for feature flags

matt-koko:
We currently have 33 flags in Atlaspack. That's a lot. 

In order to prevent this number from ballooning, we're adding an attribution field and lint rule to enforce developers own the their flag and its cleanup.

## Checklist

### DIFF
--- a/packages/dev/eslint-plugin-internal/src/rules/feature-flag-author.js
+++ b/packages/dev/eslint-plugin-internal/src/rules/feature-flag-author.js
@@ -9,10 +9,12 @@
 
 'use strict';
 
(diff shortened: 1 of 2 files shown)


packages/core/feature-flags/src/index.ts
```diff
@@ -17,28 +17,38 @@ export const DEFAULT_FEATURE_FLAGS = {
 
   /**
    * Rust backed requests
+   *
+   * @author Pedro Tacla Yamada <pyamada@atlassian.com>
```
mattcompiles, line 21: It makes more sense to associate this with me.

PR #782, opened 2025-09-09
Add @atlaspack/inspector

yamadapc:
Adds the source code for `@atlaspack/inspector`. A cache debugging and bundle size analysis tool.

This has been used successfully internally to find bundle size improvement opportunities for a short amount of time, although it is still experimental. It has been built to have good analysis UI performance for very large applications and caches.

The tool is coupled with `cachePerformanceImprovements=true` and works by opening-up the atlaspack cache directory and serving data from it over a REST HTTP API to a React front-end.

The React front-end renders tree-maps of the bundle's asset trees, along with tables with interactive "reasons" why a certain sub-tree has been included into the bundle. Reasons can be clicked to navigate to their source-code in BitBucket and GitHub repositories.

There are basic unit-tests written with `jest` for the back-end and front-end and playwright tests that smoke test the application end-to-end.

The back-end API opens the caches and exposes data with `express`.

The front-end application uses `react`, `mobx` and `foamtree` to render data. The `sigma` library is used to render graph visualisations, though they are mostly decorative.

`@atlaskit/*` components are used for UI chrome.

Adding this tool into the repository would be useful because:

- Ideally it'd be integrated into the main CLI or releases
- It is coupled with internals and versioning of the main CLI

bottom-up


.github/workflows/ci.yml
```diff
@@ -293,3 +293,75 @@ jobs:
...
+          # Linux AMD64
+          - os: ubuntu-22.04
+            node: 22
+            project: '@atlaspack/inspector'
+          - os: ubuntu-22.04
+            node: 22
+            project: '@atlaspack/inspector-frontend'
+
+          # MacOS ARM64
+          # - os: macos-15
```
JakeLane, line 313: Why's this commented out?

JakeLane (approved):

packages/dev/atlaspack-inspector/src/backend/index.ts
```diff
@@ -0,0 +1,271 @@
...
+  projectRoot,
+  repositoryRoot,
+  sourceCodeURL,
+}: BuildInspectorAppParams): express.Express {
+  const app = express();
+
+  app.use(loggingMiddleware());
+  app.use(
+    cors({
+      // origin: 'http://localhost:3333',
```
yamadapc, line 131: this is not very safe ; origin should be restricted

PR #785, opened 2025-09-10
Fix/types field publish script

matt-koko:
## Problem
TypeScript compilation is failing in consumer projects (like Confluence) because atlaspack packages are missing their type declarations. The published packages have their `types` field pointing to `./lib/index.d.ts` but the actual declaration files are located at `./lib/types/index.d.ts`.

## Root Cause
The `scripts/rewrite-package-types.mjs` script was incorrectly transforming the types field from `./src/index.ts` to `./lib/index.d.ts` instead of `./lib/types/index.d.ts`, which is where TypeScript actually outputs the declaration files according to the `declarationDir` setting in `tsconfig.json`.

We began seeing this issue once packages were published after #772 was merged.

## Solution
Updated the script to correctly point to `./lib/types/` directory where the declaration files are actually generated.

## Impact
- Fixes TypeScript errors in consumer projects
- Ensures proper type checking for atlaspack packages
- No breaking changes to the API

### DIFF
--- a/scripts/rewrite-package-types.mjs
+++ b/scripts/rewrite-package-types.mjs
@@ -21,7 +21,9 @@ for (const workspace of rootPkg.workspaces) {
     if (!pkg.source) continue;
     if (pkg.source != pkg.types) continue;
 
-    pkg.types = pkg.types.replace('./src/', './lib/types/').replace('.ts', '.d.ts');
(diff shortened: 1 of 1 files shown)


mattcompiles (approved):
How was this working before? 🤷

PR #796, opened 2025-09-15
Key lookups by Asset reference instead of id string in ScopeHoistingPackager

benjervis:
## Motivation

Most of the Maps and Sets that we have in the codebase are keyed on object references rather than id strings, but for some reason there are a bunch in the scope hoisting packager.

In theory this should be more efficient, as a reference takes up less space than a string, but this is really just about consistency.

## Changes

The `assetOutputs`, `seenAssets`, and `wrappedAssets` Map/Sets have been migrated to use an `Asset` as their key.

I was also able to remove a ts-expect-error by typing the queue.

## Checklist

### DIFF
--- a/packages/packagers/js/src/ScopeHoistingPackager.ts
+++ b/packages/packagers/js/src/ScopeHoistingPackager.ts
@@ -111,7 +111,7 @@ export class ScopeHoistingPackager {
   topLevelNames: Map<string, number> = new Map();
   seenAssets: Set<Asset> = new Set();
   wrappedAssets: Set<Asset> = new Set();
-  hoistedRequires: Map<string, Map<string, string>> = new Map();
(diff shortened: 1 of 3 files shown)


packages/packagers/js/src/ScopeHoistingPackager.ts
```diff
@@ -432,7 +433,7 @@ export class ScopeHoistingPackager {
             .getIncomingDependencies(asset)
             .some((dep) => dep.priority === 'lazy')
         ) {
-          this.wrappedAssets.add(asset.id);
+          this.wrappedAssets.add(asset);
           wrapped.push(asset);
```
mattcompiles, line 437: May as well remove the `wrapped` list as well here then yeah?

marcins (approved):

PR #824, opened 2025-09-26
[DRAFT DO NOT MERGE] Initial implementation of native tokens plugin

marcins:
## Motivation

## Changes

## Checklist


Cargo.toml
```diff
@@ -7,6 +7,9 @@ members = [
   "packages/utils/dev-dep-resolver",
 ]
 
+[workspace.package]
```
marcins, line 10: The plugin crate uses `edition = { workspace = true }` - maybe we should update our crates as well.

PR #827, opened 2025-09-30
Fix custom environment properties sometimes being ignored

OscarCookeAbbott:
## Motivation

While investigating avenues for configuring WIP native transformers I noticed that custom environment properties were not being used by the bundler and found that the second place assignment could be done seems to have just been overlooked.

## Changes

* Add missing `customEnv` assignment from descriptor in `resolvePackageTargets`
* Add to existing tests to verify addition

## Checklist


marcins (approved):
Nice one identifying that bug!

PR #830, opened 2025-10-01
Add dist differ script

benjervis:
## Motivation

As we make changes to the native side, there are often changes or refactors to make that should not change the output at all.

This script provides a way to validate that two builds are completely identical.

## Changes

There's a new `dist-differ.ts` script in the scripts directory. It can be run with node's strip types flag, pointing to the instance of the file in your local clone of Atlaspack, and needs to be passed the two directories that you want to compare.

```
node --experimental-strip-types ~/dev/atlaspack/scripts/dist-differ.ts dist-control dist-changes
```

## Checklist

[no-changeset]: Local script only, no release required

### DIFF
--- a/scripts/dist-differ.ts
+++ b/scripts/dist-differ.ts
@@ -1,0 +1,222 @@
+/* eslint-disable no-console */
+
+/**
+ * Diff the results of two different directories
+ *
+ * 1. Check if they have the same number of files
+ * 2. Report the diffreences in files and stop if they differ
+ * 3. Check if the files have the same names
+ * 4. Check that the file contents is the same
+ * 5. Report the diff on the files that differ
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+import {execSync} from 'child_process';
+
+interface FileInfo {
+  relativePath: string;
+  fullPath: string;
+  size: number;
+}
+
+function getAllFiles(dir: string, baseDir: string = dir): FileInfo[] {
+  const files: FileInfo[] = [];
+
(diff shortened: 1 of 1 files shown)


scripts/dist-differ.ts
```diff
@@ -0,0 +1,222 @@
...
+import * as path from 'path';
+import {execSync} from 'child_process';
+
+interface FileInfo {
+  relativePath: string;
+  fullPath: string;
+  size: number;
+}
+
+function getAllFiles(dir: string, baseDir: string = dir): FileInfo[] {
```
marcins, line 23: `fast-glob` would work for this as well and probably be faster.

scripts/dist-differ.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    if (Buffer.compare(content1, content2) === 0) {
+      return null; // Files are identical
+    }
+
+    // Try to get a readable diff using the system diff command
+    try {
+      const diffOutput = execSync(`diff -u "${file1}" "${file2}"`, {
+        encoding: 'utf8',
+        maxBuffer: 1024 * 1024, // 1MB buffer
```
marcins, line 69: Is that big enough for bundles?

marcins (approved):
Looks fine to me, I think I can recognise LLM generated scripts now!

Probably could make a bunch of it async for better performance, but it's fine to get it in for now and can be iterated on later.

Would be good for differing bundles to be able to compare whether it's just asset ids that changed, for example!

PR #832, opened 2025-10-03
Add native memoization cache

mattcompiles:
This PR is the first step to adding a native cache to our rust code via a new crate which is dubbed `atlaspack_memoization_cache` for now. This crate exposes a `CacheHandler` that can be bound to pretty much any persistence layer you'd like. It also has an implementation for our LMDB native client which is what we'll use once this is utilized in live code.

The `run` method allows you to pass pretty much any function as long as it's input implement `Hash` and it's result is serializable/deserializable. This can then be used to check LMDB for an existing result before computing the real result. If there's no cache hit, it will automatically persist the result for you before passing it back. 

I've also added a feature that will run validation of the cache via a configured sample rate. This is intended to help us ensure we aren't getting false positive cache hits. For now it will just log when this occurs but we can attach analytics to it later down the line.

The known next steps to get native cache going are:
- Implement plugin hashes for use in this API (likely using `packages/utils/dev-dep-resolver/src/lib.rs`) to mark plugins/pipelines as cachable
- Add new cache safe Transformer API where all side effects and IO must be run up front
- Find a way to detect and fail when invalid IO is detected in cachable Transformer plugins 
- Add caching to the asset_request when running pipelines

## Checklist

[no-changeset]: Code isn't active yet


benjervis (approved):

crates/atlaspack_memoization_cache/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+authors = ["Matt jones<mattjones701@gmail.com>"]
+name = "atlaspack_memoization_cache"
+version = "0.1.0"
+edition = "2021"
```
marcins, line 5: /nit I think we have 2024 in most other crates

Also I saw in AFM you can actually do `edition = { workspace = true }` so we should probably just do that (in a separate PR).

marcins (approved):
Seems fine to me! I don't have enough depth with async Rust (or I guess advanced Rust in general) to understand what sutble issues there might be here, but I guess we'll work those out as we start to use the implementation!

crates/atlaspack_memoization_cache/Cargo.toml
marcins: https://github.com/atlassian-labs/atlaspack/pull/836

PR #837, opened 2025-10-07
Add native `global_this_aliaser` JS transformer

OscarCookeAbbott:
## Motivation

This branch natively re-implements an in-use Babel transformer to improve build performance in applicable projects to the ultimate end goal of removing Babel transformations completely when all required re-implementations are complete.

There are no build output differences when using this transformer versus the existing Babel transformer.

## Changes

* Implement new `global_this_aliaser` transformer
* Hook up transformer using env to enable/disable

## Checklist

### DIFF
--- a/packages/transformers/js/core/src/global_this_aliaser.rs
+++ b/packages/transformers/js/core/src/global_this_aliaser.rs
@@ -80,6 +80,7 @@ impl GlobalThisAliaser {
   }
 
   pub fn should_transform(filename: &str) -> bool {
+    // Exclude files named "js.cookie.js" to avoid breaking the js-cookie library in static fallback
     !filename.contains("js.cookie.js")
   }
 }


packages/transformers/js/core/src/global_this_aliaser.rs
```diff
@@ -0,0 +1,426 @@
...
+        .any(|exclude| filename.contains(exclude))
+    } else {
+      false
+    }
+  }
+}
+
+impl VisitMut for GlobalThisAliaser {
+  fn visit_mut_expr(&mut self, node: &mut Expr) {
+    if self.should_skip_file() {
```
OscarCookeAbbott, line 108: It's possible this could be more efficient by caching etc but I elected to maintain 1:1 execution parity with the Babel transformer, at least for now.

packages/transformers/js/core/src/global_this_aliaser.rs
OscarCookeAbbott, line 1: I'm all ears for a better name for this file.

packages/transformers/js/src/JSTransformer.ts
```diff
@@ -303,6 +303,10 @@ export default new Transformer({
     let magicComments = false;
     let addReactDisplayName = false;
 
+    let enableGlobalThisAliaser = Boolean(
+      options.env.NATIVE_GLOBAL_THIS_ALIASER,
```
OscarCookeAbbott, line 307: All ears for a nicer `env` flag than this

packages/transformers/js/core/src/global_this_aliaser.rs
mattcompiles, line 108: I would suggest it'd be cleaner (and more performant) to make the whole pass conditional. Otherwise you're running this check per expression rather than per file.

packages/transformers/js/core/src/global_this_aliaser.rs
```diff
@@ -0,0 +1,426 @@
...
+        }
+      "#,
+      |run_test_context: RunTestContext| {
+        GlobalThisAliaser::new(run_test_context.unresolved_mark, None)
+      },
+    );
+
+    assert_eq!(
+      output_code,
+      "{\n    const document = 25;\n    const global = 'global value';\n    const navigator = {\n        key: 'value'\n    };\n    const window = true;\n}"
```
mattcompiles, line 351: Could we make it so all the tests use `indoc`?

packages/transformers/js/src/JSTransformer.ts
marcins: I know env is easier to flag for now, but should this be transformer config? (i.e. configured in `package.json`).

.changeset/brave-frogs-type.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atlaspack/transformer-js': minor
```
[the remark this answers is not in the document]
marcins: I don't think it needs a major bump for a new feature, it's not API breaking, it's additive.

packages/transformers/js/src/JSTransformer.ts
OscarCookeAbbott: It probably makes more sense overall, however yes that would be far more difficult to gate and I believe I recall someone said it causes perceived diffs due to the `package.json` hash changing?

packages/transformers/js/core/src/global_this_aliaser.rs
OscarCookeAbbott: Yeah I commented on the line above about this, this is a 1:1 duplicate of the Babel implementation to ensure consistency in case the re-checking matters for some reason. I can make it more efficient as you say though - is there any better approach than me adding a static condition to its `Optional::new()` condition in the transformer `lib.rs`? That seems the most efficient way, only sacrificing the configurability of the exclusion which isn't even used anyway.

PR #839, opened 2025-10-07
Add workflow_dispatch trigger to CI

benjervis:
## Motivation

Bit annoying having to open a PR just to get CI to run.

## Changes

Added the `workflow_dispatch` trigger to the CI flow, so that it can be run manually without a PR if desired.

## Checklist

[no-changeset]: CI changes only

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: Continuous Integration
 
 on:
+  workflow_dispatch:
   merge_group:
   pull_request:
   push:


matt-koko:
Is the idea that we can manually run CI for a branch without raising a PR from GitHub Actions? 

Does GitHub Actions allow us to select a branch to run this on? Or do we need to add a parameter to the action event?

Something like

```yaml
on:
  workflow_dispatch:
    inputs:
      branch:
        description: 'Branch to run the CI on'
        required: false
        default: 'main'
```

marcins (approved):

benjervis:
@matt-koko 

> Is the idea that we can manually run CI for a branch without raising a PR from GitHub Actions?

Yes

> Does GitHub Actions allow us to select a branch to run this on?

Also yes, that is part of the inputs on every action I believe. The ones you specify are just extras

PR #840, opened 2025-10-08
Cleanup inline string replacement perf flag

vykimnguyen:
## Motivation

Cleaning up `inlineStringReplacementPerf` flag as it has been proven to improve SSR build times

## Changes

## Checklist


packages/core/utils/src/replaceBundleReferences.ts
```diff
@@ -236,17 +235,10 @@ function performReplacement(
 } {
   let finalContents = contents;
 
-  if (getFeatureFlag('inlineStringReplacementPerf')) {
-    let replacementList = Array.from(replacements.values());
+  let replacementList = Array.from(replacements.values());
```
marcins, line 238: Could be a `const`, just a nit though, fine unless the linter complains!

PR #951, opened 2026-01-08
[AFBH-1740] : Normalize public URL handling in dev server

SravanKumar2305:
## Motivation

The dev server currently allows `publicUrl` values without trailing slashes (e.g., `http://localhost:8080` or `/assets`), which can result in malformed asset URLs like `localhost:8080assets.json` instead of `localhost:8080/assets.json`. This leads to broken asset loading in the development environment.

## Changes

- Added URL normalization that affects both dev server routing and bundled asset URLs 
- Added a new feature flag `normalizePublicUrlTrailingSlash` that automatically appends a trailing slash to `publicUrl` values in the dev server when enabled
- The feature flag is enabled by default in test environment but disabled in production to maintain backward compatibility
- Added integration tests to verify the normalization works correctly for both relative paths (e.g., `/assets`) and full URLs (e.g., `http://localhost:8080`)
- Updated both `packages/core/cli/src/normalizeOptions.ts` and `packages/core/feature-flags/src/index.ts`

## Checklist

---
Changeset :
---

This change adds a new feature flag `normalizePublicUrlTrailingSlash` that, when enabled, automatically adds a trailing slash to `publicUrl` values in the dev server. This prevents issues where URLs without trailing slashes (e.g., `http://localhost:8080` or `/assets`) could result in malformed asset URLs like `localhost:8080assets.json` instead of `localhost:8080/assets.json`.

The feature flag is disabled by default to maintain backward compatibility. Enable it with:

```js
featureFlags: {
  normalizePublicUrlTrailingSlash: true;
}
```

### DIFF
--- a/.changeset/normalize-publicurl-trailing-slash.md
+++ b/.changeset/normalize-publicurl-trailing-slash.md
@@ -1,0 +1,17 @@
+---
+'@atlaspack/cli': patch
+'@atlaspack/feature-flags': patch
+'@atlaspack/integration-tests': patch
+---
+
+Add feature flag to normalize publicUrl trailing slash in dev server
+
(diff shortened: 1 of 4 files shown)


marcins (approved):
As mentioned in Slack - my main concern is that there might intentionally be a base URL without a trailing slash somewhere, but this is less likely for our internal use case (FWIW Webpack/rspack for example don't append a `/` - I've fallen for that before!).

PR #1030, opened 2026-02-17
MAGMA-4673 Update sync-dynamic-imports reject unresolved imports with error and wrap require() with Promise.resolve

OliverWessels:
## Motivation

* `actual_require_paths` creates a `require(...)` statement, which is not a promise. This is not correct because `import(...)` returns a promise. This adds a new config named `sync_require_paths` intended to replace that, which wraps with `Promise.resolve`.
* Cleaning up `activate_reject_on_unresolved_imports`, which was added to feature-gate enable.

## Changes

- **Removed `activate_reject_on_unresolved_imports`** config option — unresolved dynamic imports now always generate
rejecting promises
- **Added `sync_require_paths`** config option — behaves the same as `actual_require_paths` (substring matching) but
wraps matched imports in `Promise.resolve(require(...))` instead of bare `require()`. Takes priority over
`actual_require_paths`
- **Added `syncDynamicImportRejectWithError` feature flag** — when enabled, unresolved imports reject with `new
Error(message)` (with `.skipSsr = true`) instead of a plain string
- Updated Rust (`sync_dynamic_import.rs`, `js_transformer.rs`, `lib.rs`), TypeScript (`JSTransformer.ts`), and feature
flags (`index.ts`)

### DIFF
--- a/.changeset/sync-dynamic-import-improvements.md
+++ b/.changeset/sync-dynamic-import-improvements.md
@@ -1,0 +1,7 @@
+---
+'@atlaspack/transformer-js': minor
+'@atlaspack/rust': minor
+'@atlaspack/feature-flags': minor
+---
+
+sync_dynamic_import: Remove `activate_reject_on_unresolved_imports` feature gate (now always enabled). Add `sync_require_paths` config with glob matching that wraps matched imports in `Promise.resolve(require(...))` instead of bare `require()`. Add `syncDynamicImportRejectWithError` feature flag that, when enabled, causes unresolved dynamic imports to reject with `new Error(message)` (with `.skipSsr = true`) instead of a plain string.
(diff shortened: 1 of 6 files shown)


OscarCookeAbbott (commented):
Is this change safe to just push up like this, not feature gated for rollout etc?

OliverWessels:
> Is this change safe to just push up like this, not feature gated for rollout etc?

Hey @OscarCookeAbbott I can't see exactly which code you're referring to. I believe this is feature gate safe because:
* `activate_reject_on_unresolved_imports` is always true in Jira code where it's used - as per https://switcheroo.atlassian.com/ui/gates/1d871225-44fe-4b26-97bd-c14411e11328/key/jfp-magma-error-on-unresolved-dynamic-imports
* The new `sync_require_paths` config option is new unused code.
* The `syncDynamicImportRejectWithError` feature flag controls moving from `reject(<string>)` to `reject(<Error>)`

OscarCookeAbbott:
@OliverWessels Yes I just meant in general are these changes safe without rollout etc.

It seems ok to me then, but given a recent change here needed a rollback @marcins do you have a second to review too please?

marcins (approved):

PR #1104, opened 2026-04-28
Detect resolve_import_binding calls outside Compiled CSS API bodies (AFB-2000)

JakeLane:
## Summary

Add a strict CSS-block guard that panics (when opt-in via `strict_css_block_guard: true`) if `resolve_import_binding` is called outside a Compiled CSS API call body — detecting any code path that silently crosses import boundaries for value inlining outside of `css()`/`styled.X()`/`cssMap()`/`keyframes()`/`<ClassNames>`.

### In-CSS-block tracking
- New `in_css_block: bool` field on `Metadata`, preserved across all `with_*` helpers.
- `Metadata::enter_css_block()` marks the context as inside a CSS API body.
- Every Compiled API entry point (css-prop, css-map, styled, class-names, keyframes, and `build_css` as defence-in-depth) calls `enter_css_block()`.

### Strict guard in resolve_import_binding
- Panics with a descriptive message when `in_css_block=false` and `strict_css_block_guard=true`.
- `strict_css_block_guard` is plumbed through `PluginOptions`, `CompiledCssInJsConfig`, `CompiledCssInJsTransformConfig`, and the NAPI `CompiledCssInJsConfigPlugin` struct.

### Context propagation fixes (found by running against Confluence)

| Location | Fix |
|---|---|
| `generate_cache_for_css_map_with_builder` | Skip import bindings entirely — imported `cssMap` is banned, avoids false-positives for e.g. enum identifiers inside `xcss={cx(local.x, Enum.X !== Y && local.y)}` |
| `build_css_internal` identifier branch | Propagate `in_css_block` from caller meta into `binding.meta` before recursive `build_css_internal` |
| `traverse_identifier` | Propagate `in_css_block` when evaluating a binding's initializer node |
| `resolve_variable_binding` | Propagate `in_css_block` into both the evaluation pass and the diagnostic `expression_references_import` walk |
| `expression_references_import` | Propagate `in_css_block` in recursive binding-boundary traversal |

## Tests
Three unit tests added in `utils/resolve-binding.rs`:
- Guard disabled (default): import resolution succeeds outside CSS block.
- Guard enabled + inside CSS block: import resolution succeeds.
- Guard enabled + outside CSS block: panics with "outside a CSS block".

## Validation
Ran `--build-all confluence` (7093 bundles) against the AFM monorepo with `strict_css_block_guard: true` enabled:
- Before: multiple panics (`ResultsStates`, `FULL_PAGE_EDITOR_TOOLBAR_HEIGHT`, `cursorImg`, `GRADIENT_BACKGROUND_SUBTLEST_START`, ...)
- After: **panicked=0**; 7088 ok, 5 unrelated node-builtin resolution failures.

Relates to AFB-2000.

### DIFF
--- a/crates/atlassian-swc-compiled-css/src/css-prop/index.rs
+++ b/crates/atlassian-swc-compiled-css/src/css-prop/index.rs
@@ -89,6 +89,10 @@ where
     return;
   };
 
+  // Mark `css={...}` JSX prop processing as being inside a CSS block.
+  let meta_owned = meta.enter_css_block();
+  let meta = &meta_owned;
+
   if std::env::var("COMPILED_CLI_TRACE").is_ok() {
     use swc_core::ecma::ast::JSXElementName;
     let name = match &element.opening.name {
--- a/crates/atlassian-swc-compiled-css/src/class-names/index.rs
+++ b/crates/atlassian-swc-compiled-css/src/class-names/index.rs
@@ -324,6 +324,10 @@ where
     return false;
   }
 
+  // Mark `<ClassNames>` body processing as being inside a CSS block.
+  let meta_owned = meta.enter_css_block();
+  let meta = &meta_owned;
+
   let runtime_helper = get_runtime_class_name_library(meta);
   let children_expr = ensure_children_function(element, meta);
   let (css_identifiers, style_identifiers) = collect_css_and_style_aliases(children_expr);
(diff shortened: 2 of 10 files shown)


crates/atlassian-swc-compiled-css/src/css-prop/index.rs
```diff
@@ -89,6 +89,10 @@ where
     return;
   };
 
+  // Mark `css={...}` JSX prop processing as being inside a CSS block.
+  let meta_owned = meta.enter_css_block();
```
mattcompiles, line 93: Is this right? I thought we wanted to panic on css-prop?

JakeLane: This wasn't supposed to be raised 😅

PR #1118, opened 2026-07-15
Small fixes for SyncDynamicImport plugin

OliverWessels:
## Motivation

In `SyncDynamicImport` (internal plugin only used by Jira SSR):
* `activate_reject_on_unresolved_imports` is now always-true at the only consuming point.
* The code in `create_rejecting_promise` rejects with a string, but JS typically expects a rejection to be an Error.

## Changes

* Removed `activate_reject_on_unresolved_imports` since this it is fully rolled out now and hard-coded to true in consuming location (feature gate cleaned up).
* Updated `create_rejecting_promise` to wrap the returned string in an `Error`. This is still guarded at runtime by the `__SSR_TEMP_THROW_ON_UNRESOLVED_DYNAMIC_IMPORT` global, so won't be executed until that is activated via feature gate.

## Checklist


packages/transformers/js/core/src/sync_dynamic_import.rs
```diff
@@ -650,44 +619,11 @@ mod tests {
...
-        const dummy = new Promise(()=>{});
-      "#}
-    );
-  }
-
-  #[test]
-  fn test_rejecting_promise_when_config_enabled() {
-    // When activate_reject_on_unresolved_imports is true, should generate rejecting promise
+  fn test_rejecting_promise_when_config_present() {
+    // When a config is present, unresolved imports generate a rejecting promise
```
nickrobson, line 623: By "When a config is present" does that mean the config could be not present and it would do something else?

nickrobson (approved):
Nice! Rejecting with the Error will be so nice 😄

packages/transformers/js/core/src/sync_dynamic_import.rs
OliverWessels: I'm not sure why but this code has handling for "no config" / passing in `None` rather than `Some(SyncDynamicImportConfig)`. It's always present in practise.

JakeLane (approved):
just confirming `__SSR_TEMP_THROW_ON_UNRESOLVED_DYNAMIC_IMPORT` will gate this to avoid risk rolling out as a bump? I see the Error object is functionality change on the clean up so good to have a backup

OscarCookeAbbott (approved):
Code looks fine. Needs ticket omission from metadata though.

